PR #84, opened 2017-04-07
Feat/bem styles util

rayrutjes:
This is a POC of having a utility method for generating class names that can be overloaded by the user.

The method name is intentionally short to not pollute the DOM. There is a nice improvement in readability IMHO.

Also users can now override any class by doing:
```html
<ais-pagination :classNames="{'ais-pagination': 'my-custom-pagination-class'}" />
```

**Next steps**:
- Validate the idea
- Implement this feature at the Component Mixing level so that every component can leverage it
- Leverage it in every component

### DIFF
--- a/packages/vue-instantsearch-tree-menu/src/TreeMenu.vue
+++ b/packages/vue-instantsearch-tree-menu/src/TreeMenu.vue
@@ -3,7 +3,7 @@
 
     <slot name="header"></slot>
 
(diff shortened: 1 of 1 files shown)


rafal-olszewski:
It would be nice to provide a full example on how to set custom classes to each element, and how to override the default pagination (e.g. I still can't set mine for BS4).

PR #138, opened 2017-06-06
Docs/emily feedback

rayrutjes:


rayrutjes:
Again, thanks a lot for your detailed feedback @ehayman .

Here are a few answers to the questions you raised:

**Will docs stay powered by GitBook?**
No, eventually we will use metalsmith, but I wanted to focus on content for now.
There is some ongoing work regarding docs generators and I was hoping that by waiting I could use something more future proof than yet another independent and custom build pipeline.

**Add yarn snippets**
Initially I had yarn snippets along npm snippets. It felt a bit overkill given if you are a Yarn user, then you probably know that you can use it as a replacement for npm.
So the question is, should we promote the usage of yarn? cc @Haroenv 

My POV is that it adds a tiny bit of confusion to have both in the same snippet.

**Tone**
I tried to adjust the tone here and there but might have forgotten some.

***Getting Started*
This still needs some work, and your suggestions regarding explaining everything makes a lot of sense.
However, I think that we should also keep it thin so that users can create a quick working experience, even if they don't get every single line they write.
This is about balance, but there is definitely some room left for improvement.
Some things are a bit hard to explain, and deserve references to the original docs, like 'destructuring of scoped parameters'.

**Final Word:**
I did a first iteration based on your feedback, and I'd love to iterate from here.

Haroenv:
It’s probably okay, most people know to just copy the package name. I do that even when using npm

docs/README.md
```diff
@@ -97,7 +95,7 @@ Open the `src/App.vue` component, and replace the content between the `<template
 </template>
 ```
```
ehayman: I think just a bit more info here about what's going on in this code would be helpful for newcomers.

docs/best-practices.md
```diff
@@ -0,0 +1,28 @@
+Best Practices
+---------
+
+This page is a placeholder for best practices we recommend you to review while implementing Vue InstantSearch.
```
ehayman: The updates to this content / new page is great! Do we have any app-level best practices we'd like to offer as well?

docs/search-store-instance.md
```diff
@@ -53,9 +56,15 @@ const client = algoliaClient('app_id', 'api_key');
...
+Note that there is no reason to use the Algolia API Client if you were not already using it.
+
 ### Create a search store from an Algolia Helper instance
 
-If we already have an instance of the [Algolia Helper](https://github.com/algolia/algoliasearch-helper-js), here is how we would create a search store instance:
+The [Algolia Helper](https://github.com/algolia/algoliasearch-helper-js) is a JavaScript library that is built on top of the Algolia API Client. It's goal is to expose a simple API to achieve advanced queries and it also exposes some utility methods and behaviours like keeping track of the last result.
+
+As for the Algolia Client, if you do not have an existing Algolia Helper in your application, there should be no reason for you to instantiate the search Store like showcased below.
+
+If you do have an existing Algolia Helper, her is how you can instantiate the search Store:
```
ehayman: minor typo (her => here)

docs/best-practices.md
rayrutjes: Not sure, I will try to inspire from Vue.js official docs I think.
What we could add though is security best practices.

rayrutjes:
Thanks @ehayman , we are definitely on the same page. Will iterate on that new batch of feedback.

docs/best-practices.md
ehayman: Totally agree - I think that security info would be perfect to include in this section

docs/README.md
rayrutjes: I added more information in 70a65b1

docs/best-practices.md
rayrutjes: Bootstrapped in ce44636

docs/search-store-instance.md
rayrutjes: fixed in a22c23c

PR #142, opened 2017-06-12
docs(getting started): review from @vvo

vvo:
Read and went through it as a Vue newcomer. Removed some noise
sentences/added some.

Also added screenshots to ease reading and breath.


docs/README.md
```diff
@@ -5,56 +5,52 @@
 Vue InstantSearch is the ultimate toolbox for creating instant-search
 experiences using [Vue.js](https://vuejs.org/) and [Algolia](https://www.algolia.com/).
 
-## Algolia demo credentials
```
vvo: Since the code snippets already embeds those demo credentials, I moved this block down the getting started. So we reach the "live demo" state as soon as possible.

docs/README.md
```diff
@@ -5,56 +5,52 @@
...
-For the sake of simplicity, here are some demo credentials that will give you access to
-an e-commerce website's index.
-
- - `appId`: `latency`
- - `searchKey`: `3d9875e51fbd20c7754e65422f7ce5e1`
- - `indexName`: `bestbuy`
-
 ## Setup a new Vue project
 
-You can bootstrap a new Vue project with the [official CLI tool](https://vuejs.org/v2/guide/installation.html#CLI).
```
vvo: I removed all "You can" "You should" to go for action/present sentences (this is a getting started tutorial, we must be directive so there's no place for the user to wonder/ask questions).

docs/README.md
```diff
@@ -5,56 +5,52 @@
...
 
 ```sh
+cd vue-instantsearch-getting-started
 npm install
 ```
 
 ## Install `vue-instantsearch`
 
-You should now add Vue InstantSearch as a dependency.
+Add Vue InstantSearch as a dependency, it's published on [npm](https://www.npmjs.com):
```
vvo: I believe this is a good example of a sentence change that looks good to me. Let me know.

vvo:
I was a bit lost at the end of the getting started because I don't know if it's finished or not:

```
 Now that `result` is available, we can customize the html inside the template.
  
  In the example we provided, we display the highlighted version of the name of the result.
```

When I am here I am like "do I need to do something to customize?"

rayrutjes:
I love all the changes, thanks a lot @vvo .

Regarding the abrupt ending, indeed it should be improved.

At every iteration it will get better ;)

rayrutjes:
Please go ahead and merge ;)

vvo:
> Please go ahead and merge ;)

Let's go. I am reviewing the whole documentation because I have been pushing this for long :)

Issue #143, opened 2017-06-12
Add [EDIT THIS PAGE] to docs


vvo:
Especially for new libraries, this will help a lot people to quickly edit mistakes.

rayrutjes:
I think we should create 1 issue that is "create the build system for the docs" ;)

vvo:
Doesn't gitbook provide this by default as an option?

Haroenv:
https://github.com/rtCamp/gitbook-plugin-edit-link

vvo:
and _voilà_

PR #146, opened 2017-06-12
docs: add edit in GitHub link

rayrutjes:
Closes: #143

### DIFF
--- a/book.json
+++ b/book.json
@@ -1,4 +1,12 @@
 {
   "root": "./docs",
-  "plugins": []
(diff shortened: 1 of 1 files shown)


Haroenv:
does that work without adding it to dependencies? Interesting!

rayrutjes:
image
No @Haroenv 

It works only if you run `gitbook install` before which apparently is currently not the case!

vvo:
awesome GItBook

Issue #144, opened 2017-06-12
Add a GitHub issue template


vvo:
Example: https://raw.githubusercontent.com/algolia/instantsearch.js/2bd93adaf9dde42e52407d9a75190045bb45d778/.github/ISSUE_TEMPLATE.md

PR #147, opened 2017-06-12
chore: add github issue template

rayrutjes:
Closes: #144

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,46 @@
+<!--
+  ***************************
+  Thanks for contributing to this project!
+
+  This form is to report issues or new features.
+  As for general questions like "How to hide results when the query is empty?",
+  please search or post a question to our discourse forum: https://discourse.algolia.com/.
+
+  In any case,
+    - make sure you are using the latest version of the library;
+    - do at least one search in current issues or questions, your question might already be answered;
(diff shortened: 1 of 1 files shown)


rayrutjes:
For PR's I feel like it is a bit of an overhead for now.
Maybe we could introduce it the moment we see value in it.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,46 @@
...
+**Do you want to request a *feature* or report a *bug*?**
+
+<!--
+  ***************************
+  If the current behavior is a bug, please provide all the steps to reproduce and a minimal
+  [CodePen](https://codepen.io/) example or a repository on GitHub that we can `npm install`
+  and `npm start`.
+
+  We have a simple online template for you to use as a base to explain your bug or issue:
+  https://codepen.io/rayrutjes/pen/BRgyGV
```
vvo: It's easier to modify it (maybe), what we do on instantsearch.js is that we fork it on personnal account and change the link if needed. Not the best but still ok

vvo:
> For PR's I feel like it is a bit of an overhead for now.
> Maybe we could introduce it the moment we see value in it.

👌

.github/ISSUE_TEMPLATE.md
rayrutjes: Yes that would be better.

PR #155, opened 2017-06-14
chore(eslint): use eslint-config-algolia

vvo:
In our way to make this module part of the InstantSearch family, one
part of it is to use our eslint-config-algolia and enforce it.

Some rules are currently disabled on some parts that will need to be
rewritten.


vvo:
Is there a dedicated eslint plugin for Vue? We are using eslint-plugin-html

vvo:
Also updated all deps

rayrutjes:
Did you check all breaking changes of the dependencies? It looks like there are some major version bumps.

rayrutjes:
For the linting there is https://github.com/vuejs/eslint-plugin-vue

vvo:
> Did you check all breaking changes of the dependencies?

Yes I checked the CHANGELOGS and could not find anything obvious or that should break for us.

vvo:
Did you try eslint-plugin-vue?

rayrutjes:
No I tried none of them ;) because @Haroenv told me about the algolia setup and I didn't know how much it was going to conflict 😂

src/components/Highlight.vue
```diff
@@ -1,5 +1,4 @@
-<script>
-import { HIGHLIGHT_PRE_TAG, HIGHLIGHT_POST_TAG } from '../store';
+<script>import { HIGHLIGHT_PRE_TAG, HIGHLIGHT_POST_TAG } from '../store';
```
Haroenv: I wouldn’t put this on the same line

src/components/PriceRange.vue
```diff
@@ -69,7 +68,7 @@ export default {
     },
     from: {
       get() {
-        for (let refinement in this.searchStore.activeRefinements) {
+        for (const refinement in this.searchStore.activeRefinements) {
```
Haroenv: Are you sure that for / in works with const? Didn’t see that pattern before!

vvo: I think this was automatically rewritten by eslint, but yes that's lol

src/components/Highlight.vue
vvo: weird that it did that. (prettier)

vvo: ok so prettier does that automatically, no way I can fix it for now (maybe a vue plugin would work better than the html  one)

src/components/PriceRange.vue
vvo: Also we should not use for const in on arrays but prefer for of or array methods

vvo: just checked, works perfectly, we might just be two newbes not getting the spec here

src/components/Highlight.vue
rayrutjes: Prettier in atom formats this properly. Not sure the html plugin is needed.

vvo: The html plugin was already there previously. Tomorrow I will try without it or with the vue one.

vvo:
Updated PR to only include eslint changes and no other package upgrade.

vvo:
This should be ready to merge now. Only drawback being that `<script>first line` being automatically formatted.

This is being sovled by prettier/eslint/vue right now but ultimately you don't have to think too much about it, it's automatically reformated.

src/components/__tests__/menu-tree.js
```diff
@@ -1,38 +1,36 @@
...
-          },
-        ],
-        isRefined: true,
-        name: 'Cooking',
-        path: 'Cooking',
-      },
-    ],
-  };
-});
+const getFacetValues = jest.fn(attributeName => ({
```
rayrutjes: I find this less readable than the explicit return.
I this a common good practice?

Haroenv: Yes, either use `function` with return or an arrow function with implicit return, that’s a rule in our eslint config, and it’s  usually more readable. It also gives less indentation and less brackets, so that’s better IMO

rayrutjes: Sounds good, not very used to implicit returns, mostly used in JSX these days :trollface:

vvo:
I am adding the official vue plugin since it works so we can merge this :)

vvo:
Update: still not quiet here so we should move forward with the working html plugin and deal with it for now.

vvo:
Mege?

rayrutjes:
Gogogo

vvo:
image

PR #163, opened 2017-07-06
fix(RefinementList): repeated item class

olance:
The refinement list item class is repeated when the facet value is refined

### DIFF
--- a/src/components/RefinementList.vue
+++ b/src/components/RefinementList.vue
@@ -3,7 +3,7 @@
 
     <slot name="header"></slot>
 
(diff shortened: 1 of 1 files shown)


rayrutjes:
This PR does unfortunately not address the issue properly.
The `item` class is indeed duplicated, something we need to fix.
But `active` needs to stay a "modifier" of the "item" element.

Can we revert this?

olance:
Oh sorry, I had missed this :|

rayrutjes:
No problem @olance , you still address a valid existing issue that had no open ticket. Now we are sure to tackle it. Should be done this week.

PR #166, opened 2017-07-08
Pass index along with result in Results component

julienbourdeau:
While displaying your results you may want to use a different template depending on its position.

* Use a different component for_even_ and _odd_ items
* Highlight the first 3 results using another component

### Example

```html
<ais-results>

    <template scope="{ result, index }">

        <market-sm v-if="index % 2" :market="result"></market-sm>
        <market-xl v-else :market="result"></market-xl>

    </template>
    
</ais-results>
```

### DIFF
--- a/src/components/Results.vue
+++ b/src/components/Results.vue
@@ -3,8 +3,8 @@
 
     <slot name="header"></slot>
 
-    <slot v-for="result in results" :result="result">
-      Result 'objectID': {{ result.objectID }}
+    <slot v-for="(result, index) in results" :result="result" :index="index">
+      Result {{ index }}: objectID={{ result.objectID }}
     </slot>
 
     <slot name="footer"></slot>


src/components/Results.vue
```diff
@@ -3,8 +3,8 @@
 
     <slot name="header"></slot>
 
-    <slot v-for="result in results" :result="result">
-      Result 'objectID': {{ result.objectID }}
+    <slot v-for="(result, index) in results" :result="result" :index="index">
+      Result key ( {{ index }} ) 'objectID': {{ result.objectID }}
```
Haroenv: I’m not sure if the index should be visible in the default rendering

Haroenv: Although that makes it clear you can use it. What about 

```
Results[{{ index }}]: 'objectID': {{ result.objectID }}

julienbourdeau: Yeah. it might be confusing to see "key" as well (I used it at first but it's a reserved word ^^)

rayrutjes:
Looks pretty good, also this is considered a best practice and we will need to pass the index to every scoped slot.

Will check and merge this on monday!

vvo:
Can we use indexName as for the indexName? index can be vague and is used in a lot of situations for representing iteration numbers.

Ideally every index name should be named indexName (or index-name)

Haroenv:
This isn't the indexName, but the index of the loop @vvo

julienbourdeau:
Another reason to call it `key` :D

rayrutjes:
No I think it should be "index" by convention, I will double check.

rayrutjes:
Actually it should be `key`: https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/require-v-for-key.md

julienbourdeau:
I used key at first but it didnt work. I'll try again

(https://github.com/algolia/vue-instantsearch/pull/166#discussion_r126282833)

rayrutjes:
Duplicate #220

Issue #173, opened 2017-07-12
Add possibility to use result._highlightResult.name.value


maxiloc:
It would take the default value.

For the security part of this, it should escape at retrieval and not at rendering

rayrutjes:
Done in #191

Issue #172, opened 2017-07-12
ais-highlight only replace the first occurence of the match


maxiloc:
![image](https://user-images.githubusercontent.com/1689007/28120253-c40d849c-6717-11e7-81a3-7d328c041a52.png)

rayrutjes:
Done in #191

PR #187, opened 2017-07-18
refactor(highlighting): remove highlighting special treatment

rayrutjes:
Closes: #173
Closes: #172

This PR aims to:
- Remove Highlight component
- Remove Snippet component
- Remove the post & pre highlight tag dance

We came to the conclusion that those 2 components do not provide any value and led to confusion when you intended to use the `result._highlightResult.<attrName>.value`.

Alternative solutions would be to escape all values at result retrieval at the store level. This could lead to significant impact on performances.

For now, the best we can do is remove the complexity.


docs/README.md
```diff
@@ -83,7 +83,7 @@ Open the `src/App.vue` component. Then replace the whole beginning of the file,
     <ais-search-box />
     <ais-results>
       <template scope="{ result }">
-        <h2><ais-highlight :result="result" attribute-name="name" /></h2>
+        <h2 v-html="result._snippetResult.name.value"></h2>
```
Haroenv: add a note here that you should sanitise your database

rayrutjes: For now this is an engine limitation.
Sanitizing this is a pain, and adding a comment will probably do more harm than be useful.
I think we can ship a "results sanitizer" as a decoupled piece that would just do the heavy lifting, but this needs some work & time that I don't currently have :/

WDYT?

Haroenv: Well it should just say that this is a limitation we have at the moment, and that sanitizing is completely in your own hands, because you should do it at indexing time.

rayrutjes: I changed my mind, I'm going to sanitize every highlighted and snippetted value at results retrieval.
I will try to test the perf impact of doing so and it is, as far as I can see, the easiest way to go about it.

vvo:
Some context about providing ways to get highlighting data more easily (be it components or functions):
- we did not wanted users to have to dig inside _snippetResult.name.value because 1. _field means this is "private" field (defacto standard in JS) 2. It's error prone because you need to type a lot and don't forget to use .value while most of the time you just want to show the "title" field highlighted
- we wanted to have a dedicated API to enforce users to use it (this can be solved also by having a guide). But this way we are explicit instead of asking to dig into the data

No matter what you do (outside of the XSS challenge), we need to keep this in mind.

rayrutjes:
Agreed @vvo ,

The current implementation I will try out is separate concerns.
Escaping is one thing, accessing the actual value is another.

There is also a third concern I have, is Algolia users that know about the actual _highlightResult. Those users would want to use that instead of custom, limiting, components.

I will have something by tomorrow, happy to sync on it!

rayrutjes:
Done in #191

PR #189, opened 2017-07-19
Ensure page never drops below 1 in pagination component

### LINKED ISSUE #182: Pagination disabled but still triggering page change

rayrutjes:
Makes sure pagination never goes below 1.
Also hides the pagination component if there are not search results in the current context. Thanks @olance for spotting that weird behaviour 👍 

fixes #182

### DIFF
--- a/src/components/Pagination.vue
+++ b/src/components/Pagination.vue
@@ -90,10 +90,7 @@ export default {
   },
   methods: {
     goToPage(page) {
-      let p = page;
-      if (page <= 1) {
-        p = 1;
-      }
+      const p = Math.max(1, page);
       this.searchStore.page = Math.min(this.totalPages, p);
     },
     goToFirstPage() {
--- a/src/components/Pagination.vue
+++ b/src/components/Pagination.vue
@@ -87,7 +87,11 @@ export default {
   },
   methods: {
     goToPage(page) {
-      this.searchStore.page = Math.min(this.totalPages, page);
+      let p = page;
+      if (page <= 1) {
+        p = 1;
+      }
+      this.searchStore.page = Math.min(this.totalPages, p);
     },
     goToFirstPage() {
       this.goToPage(1);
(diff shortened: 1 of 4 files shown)


src/components/Pagination.vue
```diff
@@ -84,10 +84,17 @@ export default {
...
       return pages;
     },
+    totalResults() {
+      return this.searchStore.totalResults;
+    },
   },
   methods: {
     goToPage(page) {
-      this.searchStore.page = Math.min(this.totalPages, page);
+      let p = page;
```
olance: Wouldn't `const p = Math.max(page, 1)` be better? :)

olance: Moreover, pagination with Algolia is zero-based (see https://www.algolia.com/doc/guides/searching/pagination/#response-information) so shouldn't you check for `0` instead of `1`?

rayrutjes: I implemented your style suggestion.
The pagination starts at 1 in instantsearch libraries so I made this library do the same for consistency purposes.

olance:
@Haroenv what would you show when there are absolutely no results?
Disabled previous/next buttons without any page number between them?

Haroenv:
I'm not sure, I have the feeling that should be left up to the user, some implementations hide pagination on no results (Google, Yarn), I can imagine that some might have disabled buttons with 1...8 still in it, but I'm not very sure if that's a good  after all

rayrutjes:
I have no strong opinion here.
I would say that hiding the pagination when there are no pages at all makes sense.
I would also say that somehow showing it as disabled also makes sense in terms of DX because you always see the component which might help.

What is the current behaviour on other implementations? @Haroenv 
@olance in general, for clients, what do we do? do you have valid use cases where pagination is always displayed?

olance:
I have nothing in mind, no.
Pagination happens to be a separate widget but IMO it only makes sense coupled to a list that gets paginated and browsed thanks to that.

To me, no results is a different state of the UI where you display something completely different to inform the users their search was not successful. There's no list to be paginated, so I don't see any valid reason why anyone would display disabled pagination links then.

PR #190, opened 2017-07-19
Add autofocus capabilities to SearchBox component

rayrutjes:
I also explicitly passed down the search store to the other components to ease testing.

### DIFF
--- a/docs/components/search-box.md
+++ b/docs/components/search-box.md
@@ -11,13 +11,20 @@ Basic usage:
 <ais-search-box placeholder="Find products..."></ais-search-box>
 ```
 
+With autofocus:
+
+```html
+<ais-search-box :autofocus="true"></ais-search-box>
+```
+
 ## Props
 
-| Name         | Type   | Default    | Description            |
-|:-------------|:-------|:-----------|:-----------------------|
-| placeholder  | String | `''`       | The input placeholder  |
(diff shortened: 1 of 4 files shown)


docs/components/search-box.md
```diff
@@ -11,13 +11,20 @@ Basic usage:
 <ais-search-box placeholder="Find products..."></ais-search-box>
 ```
 
+With autofocus:
+
+```html
+<ais-search-box :autofocus="true"></ais-search-box>
```
Haroenv:
```html
<ais-search-box autofocus="true"></ais-search-box>
```

This should be sufficient, no?

rayrutjes: No, because this is an actual Vue prop.

The valid HTML is `autofocus` alone, or, `autofocus="autofocus"`.

In this case, by binding it, we actually pass down a boolean to the searchbox component which would in turn be bound to the real input.

Vue will then display the real HTML attribute if it is true (as a boolean).

I hope that makes sense.

From Vue 2.4, we can now forward all unknown attributes to a given html element easily.
We will be able to leverage that so that we can write: `<ais-search-box autofocus></ais-search-box>`, but for now, that release does not work with our library :/

Haroenv: makes sense!

PR #191, opened 2017-07-21
Feat/results sanitizing

rayrutjes:
WIP

Todo:
- [x] Make sure results only gets sanitized once (not the case currently)
- [x] Allow to customize the highlighting tag
- [ ] Test perfs
- [x] Adjust highlight and snippet behaviours or remove them until we have them bring more value

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,5 +2,6 @@ module.exports = {
   extends: 'algolia/vue',
   "rules": {
     "no-warning-comments": "warn", // we have many Todo:, this will remind us to deal with them
+    "no-use-before-define": "off", 
   }
 };
--- a/src/__tests__/sanitize-results.js
+++ b/src/__tests__/sanitize-results.js
@@ -11,23 +11,20 @@ test('it ensures results are provided in the correct format', () => {
 });
 
 test('it ensures tags are provided in the correct format', () => {
+  const expectedException = new TypeError(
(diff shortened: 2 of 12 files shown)


rayrutjes:
For the perf, I won't have the time to do extended tests because we need to move forward quickly on this.

Docs will be updated once we merged @LukyVj work.

Let me know what you think.

rayrutjes:
> I'm still not convinced about any of the possible tags. What I'd like is the behaviour of before, where there's no unsafe html inserts, and with the non-html highlight tag, but with an opt-out completely where the tag isn't touched.

Not sure I get it.

What happens here is that every snippet/highlight value is escaped when we retrieve the results form the helper. This is only done once per results.
We could add a way to opt-out from the escaping, but html does not make any sense in highlighting given the HTML could be broken because of the truncating happening in Algolia.

My conclusion is, there is no reason to use unsafe values so we just escape it.

Regarding the HL components, their only job is to go get the value in the _highlightResult...path
It could then also offer fallbacks to default value and stuff.

Drawbacks of current implementation:
- Not possible to change the hl tag per component
- Not possible to access unescaped values

Pros
- allows to use standard _highlightResult.name.value syntax and have it be safe
- easier to reason about overall (IMHO)

I hope that makes sense @Haroenv , do you have scenarios where you would opt-out?

Issue #137, opened 2017-05-31
Reject store sync when error occured


rayrutjes:
When [this](https://github.com/algolia/algoliasearch-helper-js/pull/504) is resolved, we can reject the Promise [here](https://github.com/algolia/vue-instantsearch/blob/master/src/store.js#L376).

PR #193, opened 2017-07-21
fix(store): reject wait sync if error occurs

rayrutjes:
Closes: #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "algoliasearch": "^3.18.1",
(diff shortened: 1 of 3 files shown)


rayrutjes:
With a special thanks to @bobylito for making it possible in the Helper.

rayrutjes:
The only use case where we really needed this so far is server side rendering.
1. set query parameters
2. send query to Algolia
3. wait for the query to come back successfully (this is where we use it)
4. if it does not, we are able to output the error which would otherwise by hidden many layers away

PR #220, opened 2017-07-30
feat: add keys to all for loops

### LINKED ISSUE #168: Use vuejs/eslint-plugin-vue

rayrutjes:
By using [Vetur](https://github.com/vuejs/vetur), this was very easy to fix.

I think it's safe to close #168 for now.

### DIFF
--- a/src/components/Rating.vue
+++ b/src/components/Rating.vue
@@ -2,20 +2,20 @@
   <div :class="bem()" v-if="show">
     <slot name="header"></slot>
 
-    <a href="#" @click.default="clear" :class="bem('clear')" v-if="currentValue">
+    <a href="#" @click.prevent="clear" :class="bem('clear')" v-if="currentValue">
       <slot name="clear">Clear</slot>
     </a>
 
-    <div v-for="facet in facetValues" :class="[bem('item'), facet.isRefined ? bem('item', 'active') : '']" >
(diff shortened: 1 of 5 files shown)


rayrutjes:
This needs to be tested a bit before merging.

Haroenv:
Weird that key didn't work in #166

rayrutjes:
Yes, that's why I wanted to try it out first and I'm now confident it works.
Must have been a typo or something, or something related to a specific version of Vue.

src/components/Rating.vue
```diff
@@ -2,20 +2,20 @@
   <div :class="bem()" v-if="show">
     <slot name="header"></slot>
 
-    <a href="#" @click.default="clear" :class="bem('clear')" v-if="currentValue">
+    <a href="#" @click.prevent="clear" :class="bem('clear')" v-if="currentValue">
```
Haroenv: This seems like a different change, right?

rayrutjes: yes, but was lazy enough to not separate the commit :/
default is not a valid "modifier", "default" is.

rayrutjes: Actually, I will double check this

rayrutjes: Seems correct.

julienbourdeau:
🎉 🎈

Issue #196, opened 2017-07-22
[Vue warn]: Injection "_searchStore" not found


rayrutjes:
In Vue 2.4.2, the index components throws the following error:

```
vue.esm.js?65d7:479 [Vue warn]: Injection "_searchStore" not found

found in

---> <AisIndex>
       <App> at /Users/raymond/Code/vue-instantsearch-examples/examples/e-commerce/src/App.vue
         <Root>
warn @ vue.esm.js?65d7:479
resolveInject @ vue.esm.js?65d7:3429
initInjections @ vue.esm.js?65d7:3387
Vue._init @ vue.esm.js?65d7:4223
VueComponent @ vue.esm.js?65d7:4396
createComponentInstanceForVnode @ vue.esm.js?65d7:3678
init @ vue.esm.js?65d7:3495
createComponent @ vue.esm.js?65d7:5147
createElm @ vue.esm.js?65d7:5090
createChildren @ vue.esm.js?65d7:5218
createElm @ vue.esm.js?65d7:5123
patch @ vue.esm.js?65d7:5570
Vue._update @ vue.esm.js?65d7:2414
updateComponent @ vue.esm.js?65d7:2538
get @ vue.esm.js?65d7:2881
Watcher @ vue.esm.js?65d7:2870
mountComponent @ vue.esm.js?65d7:2542
Vue$3.$mount @ vue.esm.js?65d7:7863
Vue$3.$mount @ vue.esm.js?65d7:10066
init @ vue.esm.js?65d7:3501
createComponent @ vue.esm.js?65d7:5147
createElm @ vue.esm.js?65d7:5090
patch @ vue.esm.js?65d7:5606
Vue._update @ vue.esm.js?65d7:2414
updateComponent @ vue.esm.js?65d7:2538
get @ vue.esm.js?65d7:2881
Watcher @ vue.esm.js?65d7:2870
mountComponent @ vue.esm.js?65d7:2542
Vue$3.$mount @ vue.esm.js?65d7:7863
Vue$3.$mount @ vue.esm.js?65d7:10066
Vue._init @ vue.esm.js?65d7:4236
Vue$3 @ vue.esm.js?65d7:4321
(anonymous) @ main.js?1c90:10
(anonymous) @ app.js:1666
__webpack_require__ @ app.js:658
fn @ app.js:86
(anonymous) @ app.js:2933
__webpack_require__ @ app.js:658
(anonymous) @ app.js:707
(anonymous) @ app.js:710
```

[...]

PR #223, opened 2017-07-31
fix(component): avoid raising injection failure warning

rayrutjes:
Closes: #196

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -4,16 +4,15 @@ export default {
     searchStore: {
       type: Object,
       default() {
-        return this._searchStore;
-      },
-      validator(value) {
-        if (typeof value !== 'object') {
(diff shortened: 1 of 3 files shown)


rayrutjes:
Yes, actually we should separate concerns between bem utility and search store context retrieval.
The powered by does not require the search store.

rayrutjes:
Last commit fixes the powered by issue @Haroenv and makes the search store optional there.
I the future, I will refactor a bit the Component mixing and extract the bem classes utility functions.

Haroenv:
Since poweredBy is the only component not _needing_ the searchStore, this seems like a reasonable tradeoff

rayrutjes:
Yes, for now, but I keep in mind that we can do better 👍

PR #226, opened 2017-08-01
fix(spelling): run most code through a spell checker

Haroenv:
I did change `unserialize` to `deserialize`, so it's not only comments and text that changed, but it's not a public method, so it should be okay.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -3,7 +3,7 @@ import algoliaHelper from 'algoliasearch-helper';
 import { version } from '../package.json';
 import {
   serialize as serializeHelper,
-  unserialize as unserializeHelper,
+  deserialize as deserializeHelper,
 } from './helper-serializer';
 
 import sanitizeResults from './sanitize-results';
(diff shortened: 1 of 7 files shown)


src/store.js
```diff
@@ -199,6 +199,7 @@ export class Store {
     return this._helper.lastResults.processingTimeMS;
   }
 
+  // todo: change to goToFirstPage()
```
Haroenv: I'm not sure how to change this one without breaking change

rayrutjes: I was thinking of removing those.
We don't use them.

rayrutjes: BTW that is what Beta's are for, I doubt there are custom components relying on that un-documented method.

rayrutjes: Please do a separate PR though if you are to merge this one.

Haroenv: Will do a second PR to remove this

rayrutjes:
You push it, you merge it @Haroenv  😁

PR #235, opened 2017-08-02
chore: flag src dir as Vue

rayrutjes:


Haroenv:
Were the `.vue` files not picked up before?

rayrutjes:
Yes they were but not enough in the balance apparently.

Haroenv:
probably because of the test files?

rayrutjes:
yes, and some components are pure JS as well.

PR #240, opened 2017-08-03
Add BEM classes to some RefinementList tags

julienbourdeau:
I added to be closer to the original InstantSearch.


julienbourdeau:
🎉

Issue #241, opened 2017-08-04
Refresh results in single page application


a4anes:
After adding new record, instant search <ais-results> does not update. I need a way to show new items. Is there any solution?

```vue
<ais-results>
	<template scope="{ result }">
		<li class="list-group-item">
			<a href="#"><strong><ais-highlight :result="result" attribute-name="name"/></strong></a>
		</li>
	</template>
</ais-results>
```

Haroenv:
Just for clarity, you mean that when you add new items to the index, they don't show up immediately in the `results`?

a4anes:
@Haroenv Yes, exactly.

Haroenv:
It's not as simple as it seems, but it's possible. The first thing you need to do is use a custom search store, so that you can call methods on the API client. That's described [here](https://community.algolia.com/vue-instantsearch/getting-started/search-store.html#how-to-manually-create-a-search-store). Then in the place you're indexing, you'll need to [waitTask](https://www.algolia.com/doc/api-client/php/indexing/#wait-for-operations) so that you know _when_ a certain record is indexed. 

Once you have that information, you can send an event to your frontend (there are lots of ways to do that, for example with sockets) that signifies that the data needs to be updated. Once you receive that event, you can do `client.clearCache`.

Is that clear for you? Feel free to reopen if I didn't answer completely

a4anes:
@Haroenv Is there no other way to do it? Like just reloading search component or appending results?

Haroenv:
Not currently, but we could consider it. What kind of API would work well for that use case?

a4anes:
Api for frontend or backend?

a4anes:
What if i use search helper to get all results and render through vue js? Is there be any performance issue? Then i can push new item to results.

rayrutjes:
Hey @a4anes ,
I think adding a method `refresh` to the search store makes sense.
This method would take care of clearing the underlying Algolia client's cache, and fetch the results again.

Would that work for you?

a4anes:
@rayrutjes I'm using laravel package of algolia for backend and vue-instantsearch for frontend. I don't want to create custom search store but instead use vue-instantsearch components. Can you let me know what will work best in this case? It need to be index before refresh.

rayrutjes:
@a4anes in your use case, I'm not sure how you could make your frontend aware that you indexed new things from the backend. I feel like this is out of the scope of this integration.

[...]

PR #247, opened 2017-08-05
feat(store): add methods to interact with cache

rayrutjes:
Closes: #241

### DIFF
--- a/src/components/Index.vue
+++ b/src/components/Index.vue
@@ -58,8 +58,8 @@ export default {
     },
     cache: {
       type: Boolean,
-      default: true
-    }
+      default: true,
+    },
   },
   data() {
     return {
--- a/src/components/Index.vue
+++ b/src/components/Index.vue
@@ -56,6 +56,10 @@ export default {
     queryParameters: {
       type: Object,
     },
+    cache: {
+      type: Boolean,
+      default: true
+    }
   },
   data() {
     return {
@@ -83,6 +87,12 @@ export default {
 
     if (this.queryParameters) {
       this._localSearchStore.queryParameters = this.queryParameters;
+    }
+
+    if (this.cache) {
+      this._localSearchStore.enableCache();
+    } else {
+      this._localSearchStore.disableCache();
     }
 
     return {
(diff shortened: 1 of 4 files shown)


src/components/Index.vue
```diff
@@ -85,6 +89,12 @@ export default {
       this._localSearchStore.queryParameters = this.queryParameters;
     }
 
+    if (this.cache) {
+      this._localSearchStore.enableCache();
```
Haroenv: Is this needed?

src/store.js
```diff
@@ -355,11 +357,25 @@ export class Store {
     };
   }
 
-  // Todo: find a better name for this function.
   refresh() {
+    if (this._cacheEnabled === false) {
```
Haroenv: shouldn't it be true?

src/components/Index.vue
rayrutjes: This allows users to disable the cache without requiring to instantiate a search store manually.

At the beginning I didn't even think of having an `ais-index` component, but I guess it makes implementation really easy.

src/store.js
rayrutjes: No,

if the cache is disabled, then we need to clear it.
otherwise we do nothing because it is enabled by default in the client.

anyway, it needs tests

rayrutjes:
Agreed @Haroenv , the store is doing way too much. This would need to be addressed at some point, but for now, I feel like it is still OKish and at least everything is in a single place (again, not saying it's a good thing).

In my mind, we should rework it the moment we have the v4 of the client.

Regarding this specific cache feature, it is more or less a balance between what is planned for v4, and what we currently have.

Vue is often used for SPAs, and having cache in that case can be a real issue.

I even thought of turning it off by default, but I think an opt-out for now is probably safer. (gut feeling).

rayrutjes:
Otherwise we could disable the caching all-together which would result in a single line change instead of this PR :/

src/store.js
Haroenv: Oh, I understand what's going on now!

PR #248, opened 2017-08-05
fix: use unique keys in all v-for loops

rayrutjes:
cf https://vuejs.org/v2/api/#key

### DIFF
--- a/src/components/Pagination.vue
+++ b/src/components/Pagination.vue
@@ -10,7 +10,7 @@
         <slot name="previous">&lt;</slot>
       </a>
     </li>
-    <li v-for="(item, key) in pages" :key="key" :class="[bem('item'), page === item ? bem('item', 'active') : '']" >
+    <li v-for="item in pages" :key="item" :class="[bem('item'), page === item ? bem('item', 'active') : '']" >
       <a href="#" @click.prevent="goToPage(item)">
         <slot :value="item" :active="item === page">
           {{ item }}
--- a/src/components/Rating.vue
+++ b/src/components/Rating.vue
@@ -13,9 +13,9 @@
               :max="max"
               :count="facet.count"
         >
-          <template v-for="(n, key) in max">
-            <span v-if="n <= facet.value" :class="bem('star')" :key="key">&#9733;</span>
-            <span v-else :class="bem('star', 'empty')" :key="key">&#9734;</span>
+          <template v-for="n in max">
+            <span v-if="n <= facet.value" :class="bem('star')" :key="n">&#9733;</span>
+            <span v-else :class="bem('star', 'empty')" :key="n">&#9734;</span>
           </template>
           &nbsp;&amp; up
           <span :class="bem('count')">{{facet.count}}</span>
--- a/src/components/RefinementList.vue
+++ b/src/components/RefinementList.vue
@@ -3,7 +3,7 @@
 
     <slot name="header"></slot>
 
-    <div v-for="(facet, key) in facetValues" :key="key" :class="[bem('item'), facet.isRefined ? bem('item', 'active') : '']">
+    <div v-for="facet in facetValues" :key="facet.name" :class="[bem('item'), facet.isRefined ? bem('item', 'active') : '']">
       <label :class="bem('label')">
         <input type="checkbox"
                :class="bem('checkbox')"
--- a/src/components/Results.vue
+++ b/src/components/Results.vue
@@ -3,7 +3,7 @@
 
     <slot name="header"></slot>
 
-    <slot v-for="(result, key) in results" :result="result" :key="key">
+    <slot v-for="result in results" :result="result" :key="result.objectID">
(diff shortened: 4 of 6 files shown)


Haroenv:
Can you access the Key when you're using the template?

rayrutjes:
The key will be accessible from a template if it is assigned to a scoped slot (like in the Results component). Was that the question?

I'm wondering how this would handle changes that have been done on results between 2 renderings. I need to test that.

src/components/Pagination.vue
```diff
@@ -10,7 +10,7 @@
         <slot name="previous">&lt;</slot>
       </a>
     </li>
-    <li v-for="(item, key) in pages" :key="key" :class="[bem('item'), page === item ? bem('item', 'active') : '']" >
+    <li v-for="item in pages" :key="item" :class="[bem('item'), page === item ? bem('item', 'active') : '']" >
       <a href="#" @click.prevent="goToPage(item)">
         <slot :value="item" :active="item === page">
```
Haroenv: call this `currentPage`

src/components/Pagination.vue
```diff
@@ -10,7 +10,7 @@
         <slot name="previous">&lt;</slot>
       </a>
     </li>
-    <li v-for="(item, key) in pages" :key="key" :class="[bem('item'), page === item ? bem('item', 'active') : '']" >
+    <li v-for="item in pages" :key="item" :class="[bem('item'), page === item ? bem('item', 'active') : '']" >
```
Haroenv: call item `page`

src/components/Rating.vue
```diff
@@ -13,9 +13,9 @@
               :max="max"
               :count="facet.count"
         >
-          <template v-for="(n, key) in max">
-            <span v-if="n <= facet.value" :class="bem('star')" :key="key">&#9733;</span>
-            <span v-else :class="bem('star', 'empty')" :key="key">&#9734;</span>
+          <template v-for="n in max">
+            <span v-if="n <= facet.value" :class="bem('star')" :key="n">&#9733;</span>
+            <span v-else :class="bem('star', 'empty')" :key="n">&#9734;</span>
```
Haroenv: I'm not sure if these keys are unique

rayrutjes: I don't know, I need to better understand how `key` is being used internally in Vue.
Because in the case of stars for example, the counts will evolve for a given key.

In any case, I feel like Vue is still able to see if the content of a key has changed an replaces it.

rayrutjes:
From reading the docs, we can learn that keys are used:
- for removing/destroying part of the node (which works with this PR)
- for re-ordering elements instead of diffing more stuff (I'm just wondering what happens if content for a key changes)

We will also need to convert non string values to strings (like in rating).

What all started this was the eslint vue plugin which encourages the use of keys. I think the improvement in terms of perfs must be huge.
That being said we now need to figure out how Vue handles the change of content inside of an already existant key. I think that the content will only be refreshed if it is no longer part of the list. Which I think is OK for results for example. For other elements like facet values it would be a problem if it doesn't update the counts. We would need to generate a key that contains both the name + the count, and I'm not even sure that would be sufficient.

Haroenv:
for facets, the name and the count should be sufficient.

rayrutjes:
I just tried to use this branch in the example, and everything works fine :/
Content in each iteration is properly re-rendered even when it keeps same key and content changes.

So I think we should use keys in the exact flavor as this PR.
Every key should be a unique ID of the item we are looping over.

Trying to dig a bit more right now.

rayrutjes:
Here is a quick pen that illustrates the fact that Vue is able to detect changes: https://codepen.io/team/algolia/pen/mMRVrW?editors=1011

My conclusion is that the key is mainly used for re-ordering and destroying nodes, but Vue is still able to catch changes inside of each node.

So this PR is good if we fix the int->string + add tests

src/components/Pagination.vue
rayrutjes: page is already used, but it should probably be renamed to currentPage.
this would be a breaking change but do-able.

src/components/Rating.vue
rayrutjes: yes given the conclusion in the comments of this PR, the keys here are unique enough.
I need to check though if we need to cast the integer to string.

rayrutjes:
So after digging into the vdom implementation, I see that:
- vnodes are always patched
- it always checks if the entire node is different or not: https://github.com/vuejs/vue/blob/dev/src/core/vdom/patch.js#L441

so keys are really just about perf but it doesn't go as far as to avoid diffing the content of each v-node.

Regarding ratings, it is ok to keep the numbers: https://github.com/vuejs/vue/blob/5981529b497f5ef064b123aeb58eb9655172d6eb/src/core/instance/render-helpers/render-list.js#L12

rayrutjes:
I'd love a last review @Haroenv

PR #252, opened 2017-08-06
test(results): bootstrap Results component tests

rayrutjes:


src/components/__tests__/results.js
```diff
@@ -0,0 +1,110 @@
...
+});
+
+test('should allow to override result rendering', () => {
+  const searchStore = {
+    results: [{ objectID: 1 }, { objectID: 2 }],
+  };
+
+  const Component = Vue.extend({
+    render(h) {
+      return h('ais-results', {
```
Haroenv: I think Jest supports JSX by default because it's using babel, you could try it

rayrutjes: I don't think it will make this more readable.
Also I think that would require `babel-jsx`.
And currently Vue's JSX format is kinda different from the Vue one: https://github.com/vuejs/babel-plugin-transform-vue-jsx#difference-from-react-jsx

I'll probably try it out a bit when switching to babel.

PR #255, opened 2017-08-06
Chore/better testing setup

rayrutjes:
image

🎉

### DIFF
--- a/src/components/__tests__/__snapshots__/results.js.snap
+++ b/src/components/__tests__/__snapshots__/results.js.snap
@@ -3,7 +3,9 @@
 exports[`can stack results 1`] = `
 
 <div class="ais-results">
-  Result 'objectID': 1Result 'objectID': 2
+  Result 'objectID': 1
+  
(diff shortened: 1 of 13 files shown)


rayrutjes:
I was expecting to fight hard to get this working.
In the end, there are just some weird line endings in snapshots, not sure where this comes from.

src/components/__tests__/__snapshots__/results.js.snap
```diff
@@ -3,23 +3,37 @@
 exports[`can stack results 1`] = `
 
 <div class="ais-results">
-  Result 'objectID': 1Result 'objectID': 2
+  Result 'objectID': 1
+  
+    Result 'objectID': 2
```
Haroenv: This spacing seems odd. What's the cause of it?

rayrutjes: I think it is because of the Vue compiler which is used behind the scenes.

PR #256, opened 2017-08-06
Refactor/results

rayrutjes:
~Missing docs.~

### DIFF
--- a/docs/docgen/src/components/results.md
+++ b/docs/docgen/src/components/results.md
@@ -28,18 +28,40 @@ Basic usage:
 </ais-results>
 ```
 
+Display results in a table:
+
+```html
+<table>
+  <thead>
+    <tr>
+      <th>Name</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <ais-results tag-name="tbody">
+    <template scope="{ result }">
+      <tr>
+        <td>{{ result.name }}</td>
+        <td>{{ result.description }}</td>
+      </tr>
+    </template>
+  </ais-results>
+</table>
+```
+
 ## Props
 
 | Name             | Type    | Default | Description                                                                    |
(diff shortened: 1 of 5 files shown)


rayrutjes:
So the problem with `is="ais-results"` is that `thead` gets replaced with the component div:
image

That being said, you make me realize that my solution is probably not good either for in-dom templates, because we now have <ais-results> under <table> which would get removed by browsers.

I'm not sure there is a solution here actually. Because even if we use `is` in conjunction to specifying a `tbody` tag, we would then need to pass a <template> instead of a <tr> :/

Haroenv:
Is it possible to do something like:

```html
<table>
  <tbody is="ais-results”>
    <tr is=“template" scope="{ result }">
      <td>{{ result.name }}</td>
      <td>{{ result.description }}</td>
    </tr>
  </tbody>
</table>
```

rayrutjes:
Nope, just tried, but nice attempt.

rayrutjes:
The issue is the `<template>` tag.

Haroenv:
Yeah, that’s what I found too now. Maybe ask on the forum?

rayrutjes:
Only clean solution would be to have let the users use the `searchStore.results` directly.

rayrutjes:
I found the solution:

```
<ais-results inline-template>
        <table>
          <tbody>
            <tr v-for="result in results" :key="result.objectID">
              <td>{{ result.name }}</td>
              <td>{{ result.description }}</td>
            </tr>
          </tbody>
        </table>
      </ais-results>
```

This overrides the default template but keeps the actual scope of the component.
Makes it very hard to reason about the scope but I guess this exists for this kind of use cases.

rayrutjes:
I guess that it makes this PR useless.

Issue #229, opened 2017-08-01
[feature/doc] Render results in a table


blomdahldaniel:
I know you can make awesome stuff when you are free to format the results from super awesome instant search. But how about traditional tables..?

Would be nice to have the ability to present the result in a plain table. (or at least some documentation for it)

Been playing around with it and even though I set the <ais-result> component to render as an tbody it just keeps on returning itself as a div and breaks the table formatting.

```html
<tbody is="ais-results"> ... </tbody>
```
Still returns a div.

If we could just render the ais-results component as a tbody we would be all good.

rayrutjes:
Interesting use case, I'll need to try that out 👍

rayrutjes:
So indeed this is currently not do-able.
We would need to code our Results component as a render function and accepting the "tag name" as a prop. cc https://stackoverflow.com/questions/41748934/how-to-create-dynamic-tag-based-on-props-with-vue-2

Given the results component is simple, we might want to do the move.
I first want to make sure there is no other way.

rayrutjes:
Hey @blomdahldaniel , what do you think of #256?

blomdahldaniel:
Wow!
Haven't had the time to look it up but yeah,as far as I understood there was no other way but to solve it like you say.

Would be a great feature that would allow for even more flexibility and (for how I understand it) for a low cost.

rayrutjes:
Hey @blomdahldaniel , here is a working solution that already works:

```html
<ais-results inline-template>
  <table>
    <tbody>
      <tr v-for="result in results" :key="result.objectID">
        <td>{{ result.name }}</td>
        <td>{{ result.description }}</td>
      </tr>
    </tbody>
  </table>
</ais-results>
```

Let me know how that works out for you!

rayrutjes:
Here is a quick codepen that illustrates it: https://codepen.io/team/algolia/pen/qXRpzR?editors=1000

I'm also adding this to the docs in #257

blomdahldaniel:
waooo!! Bingo! Thanks for pointing that out! Ofc `inline-template` super rad man! Big thanks!
Glad that you'll add this to the docs. Like I said, this will offer even more flexibility!

Good job and thank you! :)

blomdahldaniel:
And thank you for spitting out the code example for me here in this issue!

techytuppers:
[...]

PR #257, opened 2017-08-06
docs(results): add snippet for displaying results in a table

rayrutjes:
Closes: #229

### DIFF
--- a/docs/docgen/src/components/results.md
+++ b/docs/docgen/src/components/results.md
@@ -28,10 +28,27 @@ Basic usage:
 </ais-results>
 ```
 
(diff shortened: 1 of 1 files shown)


rayrutjes:
Here is a codepen to prove it https://codepen.io/team/algolia/pen/qXRpzR?editors=1000

baphorn:
Hi @rayrutjes,
I was wondering if we can do the same without using vue.js ?

Haroenv:
What do you mean @baphorn? This is Vue InstantSearch, which library are you using? Please open an issue in the relevant repository. Thanks!

Issue #167, opened 2017-07-08
'item' class is duplicated in refinementList


Haroenv:
**Do you want to request a *feature* or report a *bug*?**
bug

**Bug: What is the current behavior?**
item is duplicated
**Bug: What is the expected behavior?**
classes being deduplicated by `bem`
**Bug: What is the proposed solution?**
find a way to dedupe in bem itself

cc @olance 

related #163 #165

PR #258, opened 2017-08-06
fix(refinement-list): remove duplicated class

rayrutjes:
Closes: #167


Haroenv:
This seems like the best change indeed

PR #260, opened 2017-08-06
Make all components SSR ready

rayrutjes:
This PR removes all usage of the `$mounted` lifecycle hook.
Only `created` and `beforeCreated` are used on the server side.

Remaining `destroyed` and `mounted` still makes sense (unless I forgot some).

### DIFF
--- a/docs/docgen/src/components/index.md
+++ b/docs/docgen/src/components/index.md
@@ -51,8 +51,8 @@ Provide search query parameters:
 | index-name       | String  | ``      | The index name                                                                                                                                     |
 | query            | String  | ``      | The search query                                                                                                                                   |
 | query-parameters | Object  | ``      | The search query parameters. Available options are [documented here](https://www.algolia.com/doc/api-client/javascript/search/#search-parameters). |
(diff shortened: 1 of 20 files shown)


rayrutjes:
Given #263 , created hook will be triggered during rendering process which is supposed to happen after the data retrieval.
Not sure this is still relevant.

rayrutjes:
Seems like the React world has the same issue: https://github.com/facebook/react/issues/1739

rayrutjes:
Would appreciate a quick review @Haroenv

docs/docgen/src/components/index.md
```diff
@@ -52,6 +52,7 @@ Provide search query parameters:
 | query            | String  | ``      | The search query                                                                                                                                   |
 | query-parameters | Object  | ``      | The search query parameters. Available options are [documented here](https://www.algolia.com/doc/api-client/javascript/search/#search-parameters). |
 | cache            | Boolean | `true`  | Wether to cache results or not. See: https://www.algolia.com/doc/tutorials/getting-started/quick-start-with-the-api-client/javascript/#cache       |
+| auto-search      | Boolean | `true`  | Wether to initiate a query to Algolia when this component is mounted                                                                               |
```
Haroenv: -> Whether

same above

Issue #262, opened 2017-08-07
Docs build system class name issue


rayrutjes:
Current docs build system transforms `ais-refinement-list__item--active` into `ais-refinement-list__item–active`.

Haroenv:
Does that also happen in code tags?

rayrutjes:
Don't know, need to try it

PR #265, opened 2017-08-08
docs: disable special typography treatment

rayrutjes:
Closes: #262


rayrutjes:
It's a bit scary to have MD do such operations no?

Haroenv:
Well all it does is fix apostrophes like `we've` => `we’ve` etc. As long as things that shouldn't be transformed are explicitly in a `code` block, it should be okay, because those are strict improvements

PR #269, opened 2017-08-09
Feat/better query parameters

rayrutjes:


src/__tests__/store.js
```diff
@@ -76,11 +78,11 @@ describe('Store', () => {
...
   });
 
-  test('should use "em" as default highlighting tag', () => {
+  test('should default highlighting tags should be undefined', () => {
     const store = createStore();
 
-    expect(store.highlightPreTag).toEqual('<em>');
-    expect(store.highlightPostTag).toEqual('</em>');
+    expect(store.highlightPreTag).toEqual(undefined);
+    expect(store.highlightPostTag).toEqual(undefined);
```
Haroenv: Why do you still need to test this?

src/__tests__/store.js
```diff
@@ -271,7 +264,9 @@ describe('Store', () => {
     };
 
     // Make sure distinct parameter is gone when overridden with undefined.
-    expect(store.queryParameters).not.toHaveProperty('distinct');
+    expect(store.queryParameters).toEqual(
+      expect.objectContaining({ distinct: undefined })
```
Haroenv: Does `undefined` get stripped out, I thought only `null` did (at engine-level)

src/__tests__/store.js
rayrutjes: Because we get the state from the helper, and we push special HL tags for XSS.
This test ensures we properly return user provided tags if any, or undefined otherwise.

Haroenv: Why don't we return the custom highlight tags if that's the case?

src/__tests__/store.js
rayrutjes: Re-using the helper logic on this one, and kept support for `null`.
`null` gets transformed into `undefined` by the store, and `undefined` will make the helper remove it from the call.

src/__tests__/store.js
rayrutjes: Tricky decision :/
Truth is it made stuff simpler for the URL sync.
But in the end we have the same issue with the page.

If think making this return the actual tags makes more sense.

I will probably revert this tomorrow.

rayrutjes:
Should be good now @Haroenv 🍸

PR #276, opened 2017-08-15
Feat/add error on missing configuration

rayrutjes:
This gets nicely caught by Vue:
image

This is disabled in production and will get removed in minified build.

Will try to add couple of tests.

### DIFF
--- a/src/components/__tests__/highlight.js
+++ b/src/components/__tests__/highlight.js
@@ -1,6 +1,12 @@
 import Vue from 'vue';
 import Highlight from '../Highlight';
 
+function restoreTestProcessEnv() {
+  process.env.NODE_ENV = 'test';
+}
+
+afterEach(restoreTestProcessEnv);
+
 test('renders proper HTML', () => {
   const result = {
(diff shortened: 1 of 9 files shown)


src/components/__tests__/highlight.js
```diff
@@ -50,3 +57,28 @@ test('should render an empty string if attribute is not highlighted', () => {
...
+          result,
+        },
+      });
+    },
+    components: {
+      Highlight,
+    },
+  }).$mount();
+
+  expect(global.console.error).toHaveBeenCalled();
```
Haroenv: why not wrap it in a function and do `expect(() => {/*everything before*/}).toThrow()` or the specific error

rayrutjes: This is because Vue catches errors happening during rendering. See: https://vuejs.org/v2/api/#errorHandler

This PR is not finished yet though, there is some cleanup coming.

rayrutjes:
Review appreciated here @Haroenv ;)

PR #279, opened 2017-08-16
feat(toggle): add toggle component

rayrutjes:
Closes: #278

Some questions:
- consider adding footer/header template
- hide if no results
- limitations due to underlying helper


tomvo:
-consider adding footer/header template
I don't think this would be necessary as I've seen you already placed the label inside the component.

- hide if no results

I would not hide it because having it visible informs the user about any potential filter options he can choose. Maybe a disabled state could work though.

- limitations due to underlying helper

Haroenv:
Will do all of the consistency work in v2

PR #281, opened 2017-08-22
docs(using-components): proofreading

codeharmonics:


docs/src/getting-started/using-components.md
```diff
@@ -107,11 +107,11 @@ All search components must be wrapped in an [`Index`](components/index.html) com
 </template>
 ```
 
-To read why a search component needs to be wrapped into an Index component, have a look at the [custom components](getting-started/custom-components.html) page.
+To read why a search component needs to be wrapped in an Index component, have a look at the [custom components](getting-started/custom-components.html) page.
 
 ### Avoiding the use of an `Index` component
 
-You might not want to use a wrapping `Index` component or maybe you want to do some [server-side rendering](advanced/server-side-rendering.html).
+You might not want to use a wrapping `Index` component, or maybe you'll want to do some [server-side rendering](advanced/server-side-rendering.html).
```
Haroenv: `-some`

rayrutjes: I made the change, thx!

PR #300, opened 2017-08-30
Only show pagination when there are multiple pages

juukie:
Makes sence to me 😉

### DIFF
--- a/src/components/Pagination.vue
+++ b/src/components/Pagination.vue
@@ -1,5 +1,5 @@
 <template>
-  <ul :class="bem()" v-show="totalResults > 0">
+  <ul :class="bem()" v-show="totalPages > 1">
     <li :class="[bem('item', 'first'), page === 1 ? bem('item', 'disabled') : '']">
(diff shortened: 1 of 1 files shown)


rayrutjes:
Hoi @juukie ,

The current condition is a design choice.
We wanted to let users know that if they were multiple pages, they would see them.

It think hiding pagination could otherwise be misleading.

Does that make sense to you?

juukie:
I can understand that @rayrutjes. Then leave it as is although I personnaly prefer to hide it.

skyrpex:
What about adding the condition to a 'show' computed property? This would allow easy overriding using the extends keyword

rayrutjes:
Yes @skyrpex , I'm all in for a way to override this.
Exposing a computed property is one way.

Other solution would be to accept a function as a callback?

Feel free to do a PR in the direction you feel most suitable and we'll discuss it.

Eventually i'd like to apply the logic on all components so that every event handler and other methods can be overridden with ease.

Issue #303, opened 2017-09-06
Automatically update UMD version number shown in the website


vvo:
It's very good to provide a fixed ve.rs.ion here: https://community.algolia.com/vue-instantsearch/examples/umd.html. Because it fixes the version and allow us to push updates without crashing websites when they dot not require an update.

But it's not up-to-date with v1.0.0. Is it feasible to update this token when we publish the doc? Using the latest package.json version

Haroenv:
We did this in instantsearch.js: https://github.com/algolia/instantsearch.js/pull/2259

Be careful when having `{{}}` in the same file though (see https://github.com/algolia/instantsearch.js/pull/2273)

PR #305, opened 2017-09-09
docs: dynamically inject package version

rayrutjes:
Closes: #303


rayrutjes:
Handlebar will only run on files having the `.hbs` extension.
Those files do not contain other brackets.

PR #321, opened 2017-10-10
Add pagination classes

rayrutjes:


rayrutjes:
For whatever reason Travis does not seem to be able to do its job.

PR #326, opened 2017-10-13
Feat/menu widget

### LINKED ISSUE #270: Add Menu component

iam4x:
Closes #270

### DIFF
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -69,10 +69,7 @@
           h(
             'div',
             {
-              class: [
-                this.bem('item'),
-                isRefined ? this.bem('item', 'active') : '',
-              ],
+              class: isRefined ? this.bem('item', 'active') : this.bem('item'),
             },
             [
               h(
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -70,7 +70,7 @@
     },
 
     created() {
-      this.searchStore.setMaxValuesPerFacet(this.limit);
+      this.searchStore.maxValuesPerFacet = this.limit;
       this.searchStore.addFacet(
         {
           name: this.attribute,
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -1,3 +1,26 @@
+<template>
+  <div :class="bem()" v-if="show">
+    <slot name="header"></slot>
+
+    <div
+      v-for="(facet, key) in facetValues"
+      :key="key"
+      :class="facet.isRefined ? bem('item', 'active') : bem('item')"
+    >
+      <a
+        href="#"
+        :class="bem('link')"
+        @click.prevent="handleClick(facet.path)"
+      >
+        {{facet.name}}
+        <span :class="bem('count')">{{facet.count}}</span>
+      </a>
+    </div>
+
+    <slot name="footer"></slot>
+  </div>
+</template>
+
 <script>import algoliaComponent from '../component';
   import { FACET_TREE } from '../store';
 
@@ -29,11 +52,13 @@
 
         return data;
       },
+      show() {
+        return this.facetValues.length > 0;
+      },
     },
 
     methods: {
-      handleClick(event, path) {
-        event.preventDefault();
+      handleClick(path) {
         this.searchStore.toggleFacetRefinement(this.attribute, path);
       },
     },
@@ -57,39 +82,6 @@
 
     destroyed() {
       this.searchStore.removeFacet(this.attribute);
-    },
-
-    render(h) {
-      if (this.show === false) return undefined;
-
-      const children = [];
-      if (this.$slots.header) children.push(this.$slots.header);
-
-      children.push(
-        this.facetValues.map(({ name, path, count, isRefined }) =>
-          h(
-            'div',
-            {
-              class: isRefined ? this.bem('item', 'active') : this.bem('item'),
-            },
-            [
-              h(
-                'a',
-                {
-                  class: this.bem('link'),
-                  domProps: { href: '#' },
-                  on: { click: event => this.handleClick(event, path) },
-                },
-                [name, h('span', { class: this.bem('count') }, count)]
-              ),
-            ]
-          )
-        )
-      );
-
-      if (this.$slots.footer) children.push(this.$slots.footer);
-
-      return h('div', { class: this.bem() }, children);
     },
   };
 </script>
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -45,6 +45,7 @@
     },
 
     created() {
+      this.searchStore.setMaxValuesPerFacet(this.limit);
       this.searchStore.addFacet(
         {
           name: this.attribute,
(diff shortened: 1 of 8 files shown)


src/components/Menu.vue
```diff
@@ -0,0 +1,96 @@
...
+      if (this.$slots.header) children.push(this.$slots.header);
+
+      children.push(
+        this.facetValues.map(({ name, path, count, isRefined }) =>
+          h(
+            'div',
+            {
+              class: [
+                this.bem('item'),
+                isRefined ? this.bem('item', 'active') : '',
```
rayrutjes: `isRefined ? this.bem('item', 'active') : this.bem('item')`

By default, `this.bem('item', 'active')` will output `ais-menu__item ais-menu__item--active`. To avoid duplicate you can do the above suggestion, or pass `false` as the last parameter to avoid outputting element class: `this.bem('item', 'active', false)`

src/components/Menu.vue
```diff
@@ -0,0 +1,96 @@
+<script>import algoliaComponent from '../component';
+  import { FACET_TREE } from '../store';
+
+  export default {
+    mixins: [algoliaComponent],
+
+    props: {
+      attribute: {
+        type: String,
```
rayrutjes: Let's make this a required property.

src/components/__tests__/menu.js
```diff
@@ -0,0 +1,101 @@
...
+      count: 319,
+      data: null,
+      isRefined: false,
+      name: 'Bathroom',
+      path: 'Bathroom',
+    },
+    {
+      count: 100,
+      data: null,
+      isRefined: false,
```
rayrutjes: How about we make one `isRefined = true`? This will allow us to catch some issues regarding class names for example.

iam4x:
Should be good now @rayrutjes thanks for the review 👍

rayrutjes:
@iam4x I changed to use disjunctive facets which seems to be what is used elsewhere for the Menu component. The helper doesn't like having to much hierarchical facets (>1 ?) so I guess it works around that as well.

Is there a specific reason you didn't use the Vue syntax for the templating? I think it would make the code more readable given there is no recursivity. I'll leave that up to you if you want to give it a try 👍

iam4x:
I've used the same strategy as we do in InstantSearch.js for the hierarchical facet (https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/menu/connectMenu.js#L156-L173) and I think we should have the same business logic in every instantsearch libraries. I would love to have @bobylito thoughts on this to understand the pros and cons of theses two strategies.

For the syntax I just did not knew and I was complaining about this ah! Thanks will give it a try 👍

src/components/Menu.vue
```diff
@@ -0,0 +1,105 @@
...
+        let alreadyRefined = false;
+        for (const refinement of this.searchStore.activeRefinements) {
+          if (
+            refinement.attributeName === this.attribute &&
+            refinement.name === value
+          ) {
+            alreadyRefined = true;
+            break;
+          }
+        }
```
iam4x:
```js
const alreadyRefined = this.searchStore.activeRefinements
  .some(({ attributeName, name }) => attributeName === this.attribute && name === value);
```

rayrutjes: That looks like Scala ;) go for it!

rayrutjes:
> I've used the same strategy as we do in InstantSearch.js for the hierarchical facet 

Arf I looked at react-instantsearch: https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/src/connectors/connectMenu.js#L200

rayrutjes:
I'll let you decide with @bobylito if you want to go with hierarchical or disjunctive.
Feel free to revert my commits if needed. I do know that I experienced issues when using multiple hierarchical trees though, so might be worst testing with some complete example.

src/components/Menu.vue
Haroenv: JavaScript nowadays is just Scala but the the tiny bit more verbose we need to allow mortals to understand it

bobylito:
The hierarchical faceting is closer to the behavior we want for the menu. It's really an implementation detail as it's hidden to the user, and therefore it should probably be used.

rayrutjes:
@iam4x I just removed the commit changing the faceting back to hierarchical.
For whatever reason GH does not display the change.
I just had the same issue on another branch.
Let's wait to see if it catches up.
I should have reverted instead of force pushing I guess :/

src/components/Menu.vue
```diff
@@ -0,0 +1,94 @@
...
+
+  export default {
+    mixins: [algoliaComponent],
+
+    props: {
+      attribute: {
+        type: String,
+        required: true,
+      },
+      limit: {
```
rayrutjes: Looks like this is not used.

rayrutjes:
Hey @iam4x , it seems that the `limit` property isn't used.
Also, it looks like there isn't any recursivity in the display which means we can use the Vue templating syntax. WDYT?

iam4x:
Since you pushed force @rayrutjes I had too, because it was a merge without changes from the origin branch.

Everything should be good now 👍

src/components/Menu.vue
```diff
@@ -0,0 +1,87 @@
...
+    },
+
+    data() {
+      return {
+        blockClassName: 'ais-menu',
+      };
+    },
+
+    created() {
+      this.searchStore.setMaxValuesPerFacet(this.limit);
```
rayrutjes: Is this how we do it in other IS? Because this will impact all other components as well.
It think we should remove the limit from this component maybe?

src/store.js
```diff
@@ -195,6 +195,14 @@ export class Store {
     return this._helper.lastResults.processingTimeMS;
   }
 
+  setMaxValuesPerFacet(limit) {
```
rayrutjes: All other options are using an ES6 setter syntax.
Can we do the same here for consistency?

src/components/Menu.vue
iam4x: Yes this is done like that, we will fix the `maxValuesPerFacet` to the maximum known limit from widgets.

src/store.js
iam4x: I don't think so, by definition a setter is used to modify a property of a class, this is not the case because we are updating the inner state of the helper to query with the correct `maxValuesPerFacet` parameter.

iam4x: Like you did for:

```js  
addFacetRefinement(attribute, value) {
    if (this._helper.state.isConjunctiveFacet(attribute)) {
      this._helper.addFacetRefinement(attribute, value);
    } else if (this._helper.state.isDisjunctiveFacet(attribute)) {
      this._helper.addDisjunctiveFacetRefinement(attribute, value);
    } else if (this._helper.state.isHierarchicalFacet(attribute)) {
      this._helper.addHierarchicalFacetRefinement(attribute, value);
    }
  }
```

rayrutjes: This is done on purpose everywhere in order to have Vue being able to observe the properties.
The Store acts as a proxy to the helper to circumvent some shortcomings of the helper.

Users can then `bind` values to that property.

src/components/Menu.vue
rayrutjes: Ok, then it needs proper documentation because this can be somewhat misleading.

iam4x:
🚢

PR #327, opened 2017-10-13
Update stats component to use toLocaleString()

ehayman:
Thoughts on updating stats component to format numbers using toLocaleString() as opposed to leaving numbers entirely unformatted?

### DIFF
--- a/src/components/Stats.vue
+++ b/src/components/Stats.vue
@@ -20,7 +20,7 @@ export default {
       return this.searchStore.query;
     },
     totalResults() {
-      return this.searchStore.totalResults;
+      return this.searchStore.totalResults.toLocaleString();
     },
     processingTime() {
       return this.searchStore.processingTimeMS;


rayrutjes:
Glad you bring it up.
I did so in the beginning and rolled back because of IE compatibility.
`toLocaleString ` is only in IE11+.

Not sure how much of an issue that is, and maybe we can do a check to test if the function exists.
WDYT?

ehayman:
Ah, good point - I would argue though with < IE11 usage being at less than 1% of the total browser usage and steadily declining, we can safely ignore it: https://www.netmarketshare.com/browser-market-share.aspx?qprid=2&qpcustomd=0 End of support was announced by Microsoft 1.5+ years ago at this point.

That said, if you feel it's still important we can definitely just do a check - this is more of a high-level question about browser support than necessarily related to this one component. :)

rayrutjes:
There you go @ehayman glad to hear we can move forward with this ;)

I have no idea why travis failed though. Will relaunch manually.

src/components/Stats.vue
```diff
@@ -20,7 +20,7 @@ export default {
       return this.searchStore.query;
     },
     totalResults() {
-      return this.searchStore.totalResults;
+      return this.searchStore.totalResults.toLocaleString();
```
Haroenv: I think we can have an argument to this function and forwarding the locale into the `toLocaleString()` call 🤔

rayrutjes:
I just changed the commit message so that it can be picked up automatically by changelog while generating the new release.

Did you see @Haroenv 's comment @ehayman ? Do you want to add that or you think the current iteration is good as a starter?

rayrutjes:
Lets do another iteration on @Haroenv suggestion in another PR if we want ;) or wait until the real need arises.

ehayman:
Sorry just catching up on GH notifs :)

Thanks for merging / accepting!

PR #331, opened 2017-10-17
fix(Stats): use computed property for conditional rendering

### LINKED ISSUE #330: Stats component is not always display
Bug

The `Stats` component is not always visible even if the `totalResults` is greater than `0`.

- on initial load the `Stats` component is not visible (`'10,000' > 0` => `false`)
- search for `He`
- it will be now visible (`'444' > 0` => `true`)

https://codepen.io/anon/pen/BwGGwE?editors=1010#0

The `Stats` component should always be visible when `totalResults` is greater than `0`.

All.

**Bug: What is the proposed solution?**

The `totalResults` computed property pass to the view is converted to a string with `toLocaleString()`. So the condition for display the content is not always predictable. 

We should pass two variables to the view, one for the display the count and the other for the condition.

`1.2.0`

samouss:
**Summary**

Fix #330

### DIFF
--- a/src/components/Stats.vue
+++ b/src/components/Stats.vue
@@ -1,5 +1,5 @@
 <template>
-  <div :class="bem()" v-if="totalResults > 0">
+  <div :class="bem()" v-if="isVisible">
     <slot :totalResults="totalResults" :processingTime="processingTime" :query="query">
       {{ totalResults }} results found in {{ processingTime }}ms
     </slot>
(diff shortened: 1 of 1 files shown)


rayrutjes:
Thanks for the PR.
I did one as well adding tests to ensure this does not break again: #332 
I think it is better to only do the toLocaleString at display time and not introduce more to the component.

WDYT?

samouss:
Close in favor of #332.

PR #336, opened 2017-10-17
Fix/vue 2.5 compat

rayrutjes:
cc @samouss 

It turns out that having a `key` binding on a slot now raises an error. Probably since 2.5.


rayrutjes:
Second commit should disappear soon, I remove it but for some reason GH still displays it.

rayrutjes:
Will close and re-open.

PR #339, opened 2017-10-17
🕹 Feat/add stories

rayrutjes:
https://59e661aca6188f358de1cf25--vue-instantsearch.netlify.com/stories/

cc @mthuret @iam4x @samouss

### DIFF
--- a/docs/src/stylesheets/components/_buttons.sass
+++ b/docs/src/stylesheets/components/_buttons.sass
@@ -80,7 +80,7 @@
 
 .btn-static-theme
   border: none
-  color: #fff
+  color: #fff !important
   background-image: linear-gradient(to right, #4DBA87, #2F9088)
   &:hover, &:focus
     color: #fff
--- a/src/components/Results.vue
+++ b/src/components/Results.vue
@@ -3,7 +3,7 @@
 
     <slot name="header"></slot>
 
-    <slot v-for="result in results" :result="result" :key="result.objectID">
+    <slot v-for="result in results" :result="result">
       Result 'objectID': {{ result.objectID }}
     </slot>
 
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,14 @@
+import { configure } from "@storybook/vue";
+
+import Vue from "vue";
+import InstantSearch, { createStoreFromAlgoliaCredentials } from "../src/instantsearch";
+
+Vue.use(InstantSearch);
+
+const req = require.context("../stories", true, /\.stories\.js$/);
+
+function loadStories() {
+  req.keys().forEach(filename => req(filename));
+}
+
+configure(loadStories, module);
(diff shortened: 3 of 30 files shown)


docs/src/stylesheets/components/_buttons.sass
```diff
@@ -80,7 +80,7 @@
 
 .btn-static-theme
   border: none
-  color: #fff
+  color: #fff !important
```
Haroenv: I'm not sure if we should add this

src/components/Results.vue
```diff
@@ -3,7 +3,7 @@
 
     <slot name="header"></slot>
 
-    <slot v-for="result in results" :result="result" :key="result.objectID">
+    <slot v-for="result in results" :result="result">
```
Haroenv: Why was this removed? Now images will be replaced weirdly, right?

samouss:
Nice! I had planned to do it too :)

src/components/Results.vue
samouss: The `slot` can't have the `:key` attribute (an error is raised) but the `template` can (see [this issue](https://github.com/vuejs/vue/issues/5575)).

docs/src/stylesheets/components/_buttons.sass
rayrutjes: Dirty fix 🙈

src/components/Results.vue
rayrutjes: > Why was this removed? Now images will be replaced weirdly, right?

The key wasn't used so far. We should try to use a key on every hit to see if that fixes the images issue.

rayrutjes:
> is it possible to pass some arguments to it? To be able to simulate a search without results or some others use cases

I guess everything is possible ;) I kept it simple for now but feel free to implement whatever you think makes sense!

rayrutjes:
I'll merge this so we can iterate, some good ideas here  🚀  feel free to implement stuff!

mthuret:
Awesome @rayrutjes o/

rayrutjes:
We will need your expertise @mthuret on StoryBook to move it forward!
Super cool stuff, it was seamless to integrate.

mthuret:
For sure, some things I can see right away: 

- we could share the same look and feel -> ie the frame around every widgets
- I usually add a `playground` story if there are some props the user can play with. 
- about each stories to create, I mostly tried to add variation of the widget state that are important for a user. 

@marielaures is currently working on adding some addons to the react storybook (usage & accessibility), she might be helpful working on this one too.

rayrutjes:
That sounds great @mthuret .

> we could share the same look and feel -> ie the frame around every widgets

Yes yes yes 👏 , I didn't do any styling and eventually was hoping you would say we can share the styles 🚀 

> I usually add a playground story if there are some props the user can play with

I did see the playground part in react-instantsearch stories, but didn't quite understand everything. Sounds super kool though.

> about each stories to create, I mostly tried to add variation of the widget state that are important for a user.

I added most of the default implementations, and now we can iterate on adding more. I'd love having someone from the team working on it!

PR #341, opened 2017-10-18
chore(stories): add Menu component stories

rayrutjes:
This PR shows that the "limit" is not respected.
https://deploy-preview-341--vue-instantsearch.netlify.com/stories/?selectedKind=Menu&selectedStory=limit%20the%20facet%20values&full=0&down=1&left=1&panelRight=0

cc @iam4x

### DIFF
--- a/docs/src/components/menu.md
+++ b/docs/src/components/menu.md
@@ -11,6 +11,8 @@ githubSource: docs/src/components/menu.md
 
 Create a menu based on a facet. A menu displays facet values and let the user selects only one value at a time. It also displays an empty value which lets the user “unselect” any previous selection.
 
+<a class="btn btn-static-theme" href="stories/?selectedKind=Menu">🕹 try out live</a>
+
 ## Usage
 
 ```html
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -70,6 +70,7 @@
     },
 
     created() {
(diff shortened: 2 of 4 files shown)


Haroenv:
I don't really understand why `limitValuesPerFacet` is something set over the whole store and not for a single facet

rayrutjes:
I raised the point in the PR @Haroenv, but it seems we do the same in IS V2.
I personally think the current implementation is misleading on that part.
The only way to solve it would be to a disjunctive facet and use the limit option there.
@bobylito thinks however that a hierarchical facet is better for this component.

iam4x:
Hey @rayrutjes can you explain me what is not working? I see 3 items when the limit=3

rayrutjes:
@iam4x when you refine on a value, you get more values than the limit.

rayrutjes:
@iam4x any news on this? Does it behave the same in other implementations?

iam4x:
This is definitely a bug for sure, not the expected behaviour.

rayrutjes:
The issue comes from the fact that hierarchical facet in the helper will trigger multiple requests.
Currently we handle the limit with maxValuesPerFacet, so if there are 2 requests, then it will go above the limit.

I'm wondering how this is handled in IS V2.

I think using maxValuesPerFacet is not correct here given it affects other components.

PR #354, opened 2017-11-03
feat: provide a loading indicator

bobylito:
This PR provides a way for the searchbox to display a spinner when the search is stalled.

The approach taken here is [similar to the one from react-instantsearch](https://github.com/algolia/react-instantsearch/pull/544): the mechanism that detects the stalled search is actually implemented at the top level. As in RIS this makes the implementation of this state update easier in multiple widgets, which seems like a good strategy.

On the UI implementation, because vue-IS doesn't implement any style, the magnifying glass svg is toggle with a spinner in the template. The effect is good as far as I can tell.

Missing elements:
 - [x] actual API to customize the parameters
 - [x] tests

Question: 
 - not sure about the implementation in store.js/waitUntilSync @rayrutjes 

Related to https://github.com/algolia/instantsearch.js/pull/2509

### DIFF
--- a/src/components/Index.vue
+++ b/src/components/Index.vue
@@ -61,6 +61,10 @@ export default {
       type: Boolean,
       default: true,
     },
+    stalledSearchTimeout: {
+      type: Number,
+      default: 200,
+    },
   },
   data() {
     return {
@@ -71,7 +75,8 @@ export default {
     if (!this.searchStore) {
       this._localSearchStore = createFromAlgoliaCredentials(
         this.appId,
-        this.apiKey
+        this.apiKey,
+        this.stalledSearchTimeout,
       );
     } else {
       this._localSearchStore = this.searchStore;
--- a/src/components/Index.vue
+++ b/src/components/Index.vue
@@ -62,7 +62,7 @@ export default {
       type: Boolean,
       default: true,
     },
-    stalledSearchTimeout: {
+    stalledSearchDelay: {
       type: Number,
       default: 200,
     },
@@ -77,7 +77,7 @@ export default {
       this._localSearchStore = createFromAlgoliaCredentials(
         this.appId,
         this.apiKey,
-        this.stalledSearchTimeout
+        { stalledSearchDelay: this.stalledSearchDelay }
       );
     } else {
       this._localSearchStore = this.searchStore;
--- a/src/components/Index.vue
+++ b/src/components/Index.vue
@@ -76,7 +76,7 @@ export default {
       this._localSearchStore = createFromAlgoliaCredentials(
         this.appId,
         this.apiKey,
-        this.stalledSearchTimeout,
+        this.stalledSearchTimeout
       );
     } else {
       this._localSearchStore = this.searchStore;
--- a/src/components/SearchBox.vue
+++ b/src/components/SearchBox.vue
@@ -3,14 +3,14 @@
     <slot>
       <ais-input :search-store="searchStore" :placeholder="placeholder" :autofocus="autofocus"></ais-input>
       <button type="submit" :class="bem('submit')">
-        <svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 40 40" v-if="!searchStore.isSearchStalled">
+        <svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 40 40" v-if="!showLoadingIndicator || !searchStore.isSearchStalled">
           <title>{{ submitTitle }}</title>
           <path
             d="M26.804 29.01c-2.832 2.34-6.465 3.746-10.426 3.746C7.333 32.756 0 25.424 0 16.378 0 7.333 7.333 0 16.378 0c9.046 0 16.378 7.333 16.378 16.378 0 3.96-1.406 7.594-3.746 10.426l10.534 10.534c.607.607.61 1.59-.004 2.202-.61.61-1.597.61-2.202.004L26.804 29.01zm-10.426.627c7.323 0 13.26-5.936 13.26-13.26 0-7.32-5.937-13.257-13.26-13.257C9.056 3.12 3.12 9.056 3.12 16.378c0 7.323 5.936 13.26 13.258 13.26z"
             fillRule="evenodd"
           />
         </svg>
-        <svg width="1em" height="1em" viewBox="0 0 38 38" xmlns="http://www.w3.org/2000/svg" stroke="#000" v-if="searchStore.isSearchStalled">
+        <svg width="1em" height="1em" viewBox="0 0 38 38" xmlns="http://www.w3.org/2000/svg" stroke="#000" v-if="showLoadingIndicator && searchStore.isSearchStalled">
           <g fill="none" fill-rule="evenodd">
               <g transform="translate(1 1)" stroke-width="2">
                   <circle stroke-opacity=".5" cx="18" cy="18" r="18"/>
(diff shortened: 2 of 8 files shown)


src/components/SearchBox.vue
```diff
@@ -3,13 +3,29 @@
     <slot>
       <ais-input :search-store="searchStore" :placeholder="placeholder" :autofocus="autofocus"></ais-input>
       <button type="submit" :class="bem('submit')">
-        <svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 40 40">
+        <svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 40 40" v-if="!searchStore.isSearchStalled">
```
Haroenv: Works really nice 😄 

Maybe have this as an option though

src/components/SearchBox.vue
```diff
@@ -3,13 +3,29 @@
...
           <title>{{ submitTitle }}</title>
           <path
             d="M26.804 29.01c-2.832 2.34-6.465 3.746-10.426 3.746C7.333 32.756 0 25.424 0 16.378 0 7.333 7.333 0 16.378 0c9.046 0 16.378 7.333 16.378 16.378 0 3.96-1.406 7.594-3.746 10.426l10.534 10.534c.607.607.61 1.59-.004 2.202-.61.61-1.597.61-2.202.004L26.804 29.01zm-10.426.627c7.323 0 13.26-5.936 13.26-13.26 0-7.32-5.937-13.257-13.26-13.257C9.056 3.12 3.12 9.056 3.12 16.378c0 7.323 5.936 13.26 13.258 13.26z"
             fillRule="evenodd"
           />
         </svg>
+        <svg width="1em" height="1em" viewBox="0 0 38 38" xmlns="http://www.w3.org/2000/svg" stroke="#000" v-if="searchStore.isSearchStalled">
```
Haroenv: this should be a slot I think

src/components/SearchBox.vue
```diff
@@ -3,13 +3,29 @@
...
       <button type="submit" :class="bem('submit')">
-        <svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 40 40">
+        <svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 40 40" v-if="!showLoadingIndicator || !searchStore.isSearchStalled">
           <title>{{ submitTitle }}</title>
           <path
             d="M26.804 29.01c-2.832 2.34-6.465 3.746-10.426 3.746C7.333 32.756 0 25.424 0 16.378 0 7.333 7.333 0 16.378 0c9.046 0 16.378 7.333 16.378 16.378 0 3.96-1.406 7.594-3.746 10.426l10.534 10.534c.607.607.61 1.59-.004 2.202-.61.61-1.597.61-2.202.004L26.804 29.01zm-10.426.627c7.323 0 13.26-5.936 13.26-13.26 0-7.32-5.937-13.257-13.26-13.257C9.056 3.12 3.12 9.056 3.12 16.378c0 7.323 5.936 13.26 13.258 13.26z"
             fillRule="evenodd"
           />
         </svg>
+        <svg width="1em" height="1em" viewBox="0 0 38 38" xmlns="http://www.w3.org/2000/svg" stroke="#000" v-if="showLoadingIndicator && searchStore.isSearchStalled">
```
Haroenv: Would still be nice to have this as a slot :)

src/store.js
```diff
@@ -450,4 +475,18 @@ const onHelperResult = function(response) {
...
+
+  if (!this._helper.hasPendingRequests()) {
+    clearTimeout(this._stalledSearchTimer);
+    this._stalledSearchTimer = null;
+    this.isSearchStalled = false;
+  }
+};
+
+const onHelperSearch = function() {
+  if (!this._stalledSearchTimer) {
```
Haroenv: not sure if this is a big deal, but maybe it could not be done only if the stalled search feature is enabled? Might be hard to find out about it here...

src/components/SearchBox.vue
bobylito: Indeed. Will make this happen :)

src/store.js
bobylito: The feature is always there, for simplicity sake.

bobylito:
I've updated this PR to be consistent with the other InstantSearch implementations.

src/components/SearchBox.vue
```diff
@@ -24,8 +44,7 @@
   </form>
 </template>
 
-<script>
-import algoliaComponent from '../component';
+<script>import algoliaComponent from '../component';
```
rayrutjes: Looks like the weird syntax is coming back here @Haroenv . Expected?

src/components/Index.vue
```diff
@@ -72,7 +76,8 @@ export default {
     if (!this.searchStore) {
       this._localSearchStore = createFromAlgoliaCredentials(
         this.appId,
-        this.apiKey
+        this.apiKey,
+        { stalledSearchDelay: this.stalledSearchDelay }
```
rayrutjes: Can we consider not adding another "options" parameter, and introduce a getter instead?
So far, the Store is just proxying as much as possible.
Also, if tomorrow the `stalledSearchDelay` is added to the helper, we won't require the option anymore and it would be an issue to be able to have it in the helper + as an option, because we won't know which one to keep.

src/components/SearchBox.vue
```diff
@@ -2,7 +2,27 @@
   <form role="search" action="" @submit.prevent="onFormSubmit">
     <slot>
       <ais-input :search-store="searchStore" :placeholder="placeholder" :autofocus="autofocus"></ais-input>
-      <button type="submit" :class="bem('submit')">
+      <div v-if="showLoadingIndicator" :style="searchStore.isSearchStalled ? 'display:block;' : 'display:none;'" :class="bem('loading-indicator')" >
```
rayrutjes: I would probably go for `v-show="showLoadingIndicator && searchStore.isSearchStalled"` and remove the `style` attribute.

src/components/SearchBox.vue
Haroenv: not expected, maybe @bobylito's vim setup, should be put on the new line IMO

src/components/SearchBox.vue
bobylito: Agreed for that it would make more sense given the vue-specifics. However I chose that in order to have consistency accross all implementations. This is the exact same markup and changes on it depending on the value of `isSearchStalled` in all implementations.

src/components/SearchBox.vue
bobylito: I'll double check. Maybe I didn't update the packages that do the formating 🤔

src/components/Index.vue
bobylito: I see what you mean. However my thinking behind choosing this as an option is that this is a feature in the scope of InstantSearch, not the helper. This might evolve but the way it is implemented right now is the way it is in the other libraries.

src/components/SearchBox.vue
bobylito: So it's not an error from any linter / formatter. I might have just removed it by mistake :/

rayrutjes: I'm pretty sure it's because of the linter or something.
Happens to me all the time but can't remember what the real issue was.

src/components/SearchBox.vue
Haroenv: You should use the `hidden` attribute, which implies `display: none;` but is better for a11y

Haroenv: `:hidden="!searchStore.isSearchStalled"`

bobylito:
Updated per your  suggestion @Haroenv thanks :)

PR #356, opened 2017-11-10
chore(components): put import on newline

Haroenv:
newer version of Vetur indents it correctly, so I saved all files :fire:


rayrutjes:
What about `yarn lint` though? Does that work as well? Or will it rollback your changes?

Haroenv:
Not sure. I assumed it would work. Does CI not run lint?

PR #358, opened 2017-11-11
Update scoped slot docs

salparadise:
The "scope" attribute for scoped slots have been deprecated and replaced by "slot-scope" since Vue 2.5.


rayrutjes:
Thanks @salparadise !

The question is, are Vue users always upgrading to the latest version of Vue?

I'd be inclined to move forward with the new syntax but for that we also need to specify the min required version somewhere.

Regarding your changes, we could get rid of most of the `template` tags in the new version. Any reason you didn't remove those?

salparadise:
You're right, that's a breaking change and it won't work with previous releases. In my opinion documentation should target current release where possible. It provides a better user experience for those who want to start building from scratch. 

The only reason I left them is that personally I find it more practical to take more verbose path and treat scoped slot as a container as it is easier to structure things. If you wish I can change it. As a matter of fact documentation should also be concise :)

rayrutjes:
Thanks a lot @salparadise !

Issue #357, opened 2017-11-11
Refinement list / tree menu with vue-router


salparadise:
**Do you want to request a *feature* or report a *bug*?**

Report a bug.

**Bug: What is the current behavior?**

I'm using vue-instantsearch along with vue-router to make a simple SPA. There are some refinement list and tree menu components on the search page. After clicking on any router link, even outside of index component, it is behaving as if I'm clicked on refinement list / tree menu item. The numer of requests depends on the number of components. Let's say I have 3 tree menus and 2 refinement lists, then there will be 5 (identical) requests.

**Bug: What is the expected behavior?**

There should be no requests.

**Bug: What browsers are impacted? Which versions?**

All.

**What is the version you are using? Always use the latest one before opening a bug issue.**

1.3.1

Haroenv:
Can you make a small replication in a repo, it will be a lot more straightforward to see which situation you are hitting

salparadise:
Sure, here you go: https://github.com/salparadise/vue-instantsearch-spa

If you navigate to search page and then navigate back to home page, you will see that on "Home" click two requests are being made.

Haroenv:
perfect, will take a look at that soon

rayrutjes:
Just tried it, and I consider it is a bug.

What happens is that when search components gets destroyed, we remove facets from the search store. This triggers a search operation because the query parameters changed.

Here we probably need to either:
1. not remove anything when components get destroyed
2. stop the store before removing the facet

Should be a quick fix.
I'll go with 1. for now has it will introduce less changes.

rayrutjes:
I tried #359 with your example @salparadise and it works ;)

PR #359, opened 2017-11-11
fix: avoid query when components are destroyed

rayrutjes:
Closes: #357

### DIFF
--- a/src/components/__tests__/refinement-list.js
+++ b/src/components/__tests__/refinement-list.js
@@ -17,6 +17,8 @@ const addFacet = jest.fn();
 const searchStore = {
   getFacetValues,
   addFacet,
+  stop: () => {},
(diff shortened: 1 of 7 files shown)


src/components/__tests__/refinement-list.js
```diff
@@ -17,6 +17,8 @@ const addFacet = jest.fn();
 const searchStore = {
   getFacetValues,
   addFacet,
+  stop: () => {},
```
Haroenv: can you jest.fn these so you test they get called correctly?

rayrutjes: just testing they got called add nothing we would need to test the got called in order.

Haroenv:
can you fix the conflict @rayrutjes ?

PR #361, opened 2017-11-15
fix dom

JonasBa:
Duplicate of [this](https://github.com/algolia/react-instantsearch/pull/610) :)
cc @maxiloc


Haroenv:
This seems pretty weird, I assumed pug would do it automatically. IS there some output comparison?

JonasBa:
Haven't tested, but as far as the pug docs go, I can't find anywhere a documentation explaining that pug does this by default.

JonasBa:
Thanks a lot \o/ I might make build an automated testing script to check dom validity, visual regressions testing and possibly contrast issues, so we avoid this in the future :/

rayrutjes:
Arf, this is bad usage of pug, see #362

PR #368, opened 2017-12-04
fix: Powered-by component is not SSR compatible #367

### LINKED ISSUE #367: Powered-by component is not SSR compatible

bug

Steps:

- Create an empty `nuxt` starter project and install `vue-instantsearch`

- add an `<ais-powered-by>` component into one of the templates

- run `npm run dev`

- visit the page with the component and notice the error screen saying `location is undefined`

`<ais-powered-by>` is not SSR compatible, 

The component should be compatible with ssr

This is an SSR issue, so node.js?

**Bug: What is the proposed solution?**

Add a null check for the `location` object

1.3.2

Jexordexan:
Add a null check to the `location` reference in `algoliaUrl`.

fixes #367

### DIFF
--- a/src/components/PoweredBy.vue
+++ b/src/components/PoweredBy.vue
@@ -43,7 +43,7 @@ export default {
         'https://www.algolia.com/?' +
         'utm_source=vue-instantsearch&' +
         'utm_medium=website&' +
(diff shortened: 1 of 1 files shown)


rayrutjes:
Thanks @Jexordexan !

PR #373, opened 2017-12-18
docs(index): correct documentation links

Haroenv:
1. cache: make into a shortened link
2. api parameters: use the correct link, cc @kfirba


kfirba:
Thanks for the quick fix!

Issue #381, opened 2018-01-24
Whitespace between checkbox and default slot in RefinementList


NicTorgersen:
Insanely tiny bug incoming!

**What is the current behavior?**

There is an awkward whitespace between checkbox and text content. This causes a visual bug with the UI when the label has `text-decoration: underline;`.

**What is the expected behavior?**

That there's no hard coded whitespace between the checkbox and the text content.

As seen here, Filter 1 has the issue. Filter 2 has the whitespace removed, which is the expected visual style.
<img width="110" alt="screen shot 2018-01-24 at 11 19 33" src="https://user-images.githubusercontent.com/1327106/35326971-8f518282-00f8-11e8-9b99-7782ab87bff0.png">

**What is the proposed solution?**

I cloned the repo, and had a look at the component source. It looks a lot like there's a newline character which is being converted to a whitespace between the checkbox and default slot.

**What is the version you are using? Always use the latest one before opening a bug issue.**

Version 1.3.3 of vue-instantsearch

For the moment I'm using some CSS tricks to rectify the issue.

```css
.ais-refinement-list__item label { font-size: 0; } /* this does the trick */
.ais-refinement-list__item input { font-size: 16px; }
.ais-refinement-list__item span { font-size: 16px; } /* added inline template with span around content */
```

Haroenv:
Do you want to do a PR for this?

NicTorgersen:
Sure :)

PR #384, opened 2018-01-29
fix(refinementList): avoid unwanted spacing

NicTorgersen:
this should fix https://github.com/algolia/vue-instantsearch/issues/381


Haroenv:
Are you sure it's not the other newline that's causing the space?

NicTorgersen:
@Haroenv yep, just built and tested and it works now :) 👍

Haroenv:
I took another look now, and found the correct newline to remove. Annoying that the indentation looks a bit funky now, but at least there's no stray newline anymore. Waiting for the tests to be ran to merge this

PR #394, opened 2018-02-14
#393 - Feature: pass in the page when emitting the page change event

### LINKED ISSUE #393: Feature: pass in the page when emitting the page change event
**Feature: What is your use case for such a feature?**

Right now when the page change event is emitted it's empty. it'll be nice to get the page it was changed to passed in.

This might not be an actual feature or a bug but rather a minor change. I can submit a PR if needed. ~~I was also wondering if there is a way to access the current page in any other way?~~

Edit: I guess the page can be accessed with the following code:

`this.$refs.search._localSearchStore.page`

However the page being passed as part of the event is helpful.

Thank You.

VinceG:
fixes #393

### DIFF
--- a/src/components/__tests__/pagination.js
+++ b/src/components/__tests__/pagination.js
@@ -90,7 +90,7 @@ test('it should be hidden if there are no results in the current context', () =>
   expect(vm.$el.outerHTML).toMatchSnapshot();
 });
 
-test('it should emit a "page-change" event when page changes', () => {
+test('it should emit a "page-change" event when page changes and pass in the page variable', () => {
   const searchStore = {
     page: 1,
     totalPages: 20,
@@ -108,7 +108,7 @@ test('it should emit a "page-change" event when page changes', () => {
 
   vm.$mount();
 
-  expect(onPageChange).not.toHaveBeenCalled();
+  expect(onPageChange).not.toHaveBeenCalledWith('page');
 
   vm.$el
     .getElementsByTagName('li')[3]
--- a/src/components/__tests__/pagination.js
+++ b/src/components/__tests__/pagination.js
@@ -108,11 +108,15 @@ test('it should emit a "page-change" event when page changes and pass in the pag
 
   vm.$mount();
 
-  expect(onPageChange).not.toHaveBeenCalledWith('page');
+  expect(onPageChange).not.toHaveBeenCalled();
 
   vm.$el
     .getElementsByTagName('li')[3]
     .getElementsByTagName('a')[0]
     .click();
   expect(onPageChange).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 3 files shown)


Haroenv (approved):
Thank you! Since people didn't expect anything in that callback, we can release this as a patch I think

Haroenv:
Can you edit the test too? 

https://github.com/algolia/vue-instantsearch/blob/40c9406c5f0d477324a2e9061c75241f6debad6e/src/components/__tests__/pagination.js#L93-L118

VinceG:
@Haroenv Done.

src/components/__tests__/pagination.js
```diff
@@ -108,7 +108,7 @@ test('it should emit a "page-change" event when page changes', () => {
 
   vm.$mount();
 
-  expect(onPageChange).not.toHaveBeenCalled();
+  expect(onPageChange).not.toHaveBeenCalledWith('page');
```
Haroenv: This should have stayed the same, it's testing that it wasn't called yet

src/components/__tests__/pagination.js
```diff
@@ -108,7 +108,7 @@ test('it should emit a "page-change" event when page changes', () => {
...
 
   vm.$el
     .getElementsByTagName('li')[3]
```
Haroenv: after these lines add: 


```js
expect(onPageChange).toHaveBeenCalledWith(page);
```

Haroenv (changes requested):
Thanks for adding the tests, I think you changed the wrong test though.

src/components/__tests__/pagination.js
VinceG: @Haroenv  How about this:

```
test('it should emit a "page-change" event when page changes and pass in the page variable', () => {
  const searchStore = {
    page: 1,
    totalPages: 20,
    totalResults: 4000,
  };
  const Component = Vue.extend(Pagination);
  const vm = new Component({
    propsData: {
      searchStore,
    },
  });

  const onPageChange = jest.fn();
  vm.$on('page-change', onPageChange);

  vm.$mount();

  expect(onPageChange).not.toHaveBeenCalled();

  vm.$el
    .getElementsByTagName('li')[3]
    .getElementsByTagName('a')[0]
    .click();
  expect(onPageChange).toHaveBeenCalledTimes(1);

  const page = searchStore.page;
  expect(onPageChange).toHaveBeenCalledWith(page);
  expect(page).toBe(2);
});
```

Haroenv: Exactly what I meant!

Haroenv (approved):
Great! Thanks for your contribution. I’ll release this soon

PR #399, opened 2018-03-02
fix(menu): avoid error when no facet values are retrieved

rayrutjes:
Works around a bug in the Menu component implementation:
image

cc @Shipow

### DIFF
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -46,7 +46,7 @@ export default {
 
   computed: {
     facetValues() {
-      const { data = [] } = this.searchStore.getFacetValues(
+      const { data } = this.searchStore.getFacetValues(
         this.attribute,
         this.sortBy
       );
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -51,7 +51,11 @@ export default {
         this.sortBy
       );
 
-      return data;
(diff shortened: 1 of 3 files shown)


src/components/Menu.vue
```diff
@@ -51,7 +51,11 @@ export default {
         this.sortBy
       );
 
-      return data;
+      if (Array.isArray(data)) {
```
Haroenv: in which case can this happen, since the default value is `[]`

Haroenv:
What about adding `|| []` to https://github.com/algolia/vue-instantsearch/blob/db113a41c3f7796f7473634a55477a1269b03b1b/src/store.js#L311 instead?

src/components/Menu.vue
rayrutjes: This can happen when `getFacetValues` returns `{data: null}` like in the tests.

src/components/Menu.vue
```diff
@@ -46,12 +46,16 @@ export default {
 
   computed: {
     facetValues() {
-      const { data = [] } = this.searchStore.getFacetValues(
```
rayrutjes: @iam4x FYI: default value assignment when destructuring only kicks on `undefined` values.
The helper often returns `null` values, which is the case here so we can't solely rely on this.

PR #403, opened 2018-03-09
feat(store): expose userData through a new getter

rayrutjes:
cc @ehayman 

I'm wondering if we should always return an empty array in case no data is available or if we should leave this check up to the user.

### DIFF
--- a/src/__tests__/store.js
+++ b/src/__tests__/store.js
@@ -116,6 +116,27 @@ describe('Store', () => {
     expect(store.page).toEqual(2);
   });
 
+  test('can retrieve userData if it exists', () => {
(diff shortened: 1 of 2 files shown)


ehayman:
Good point... I think in this case, returning the empty array would be safest / simplest

rayrutjes:
@ehayman, I'll let @Haroenv handle the release whenever he sees fit.

> Undefined is the good behaviour imo

I agree with that, let's stay as close as possible to the helper behaviour.

You will need to check if the returned value is undefined or not.

Haroenv:
I’ll release Monday!

PR #406, opened 2018-03-15
fix(RangeInput): make the header slot valid

### LINKED ISSUE #405: Populating header slot in Range Input component causes inputs to disappear
Report a bug

When populating the `header` slot in the `range input` component, the inputs, separator, and button do not appear.

The range inputs, separator, and button should appear.

Latest versions of Chrome, Firefox, and Edge.

1.5.1

See code beginning at line 53 in App.vue in this sandbox:
[Edit example-e-commerce](https://codesandbox.io/s/54z38ykoxx)

Haroenv:
fixes #405

### DIFF
--- a/src/components/RangeInput.vue
+++ b/src/components/RangeInput.vue
@@ -1,6 +1,6 @@
 <template>
   <div :class="bem()">
-    <slot name="header" />
+    <slot name="header"></slot>
 
     <form @submit.prevent="onSubmit(refinement)">
       <input


Haroenv:
Not sure why it broke, but I saw that this was the only difference between RangeInput and PriceRanges w.r.t. slots and this one didn't work, while the other one did

rayrutjes:
For me this only changes something if one tries to directly use the component in the browser instead of using UMD or compiling with the vue compiler.

I might be wrong though :)

Haroenv:
Thought to too, but I don't understand why else that codeSandbox was behaving like it was

PR #420, opened 2018-04-12
feat(widgets): update menu markup && support showMore

iam4x:


Haroenv:
Maybe let's update the test too in the same PR so there are some snapshots to look at?

Haroenv:
According to InstantSearch.css, there should be a `showMore=false`, so you don't have to put `showMoreLimit` to 0 to disable show more

PR #424, opened 2018-04-13
ci(yarn): select and enforce yarn version

vvo:


Haroenv:
shouldn't it also be put in `package.json#engines`?

vvo:
> shouldn't it also be put in package.json#engines?

Yes, but I do remember people complaining about that (contributors). Ideally yes it must be strict

PR #425, opened 2018-04-13
fix(Index/component): avoid doing multiple requests at launch

vvo:
**This PR contains:**
- refactor on how we are making requests to Algolia/using InstantSearch.js internally. Now there's only one request being done and we are more in control about what's going on given the lifecycle of Vue
- cleanup of some files that were breaking tests
- alignment of Hits, HitsPerPage, Stats (the only really finished widgets) on some new stuff (v-if usage)

The interesting commit is this one: https://github.com/algolia/vue-instantsearch/pull/425/commits/d12e1405c7f808bd87072db0f1243bce25b74896. Other commits are cleaning chores.

**Questions:**
- Right now every widget uses v-if="state" to avoid rendering anything the first time it mounts. I wonder if it could be done in a nicer way but not even sure. There's no big issue of doing what we are doing it seems, it just works.

---
**Before**

**After**


src/components/Index.vue
```diff
@@ -1,147 +1,56 @@
 <template>
-  <div :class="suit()">
```
Haroenv: I think that it still needs the class name

src/components/Index.vue
```diff
@@ -1,147 +1,56 @@
...
     stalledSearchDelay: {
-      type: Number,
-      default: 200,
-    },
+      type: Number
+    }
   },
   data() {
     return {
-      widgetName: 'ais-index',
```
Haroenv: with `widgetName: 'Index'`

src/components/Index.vue
```diff
@@ -1,147 +1,56 @@
...
 import instantsearch from 'instantsearch.js/es/';
 
 export default {
-  mixins: [algoliaComponent],
+  provide() {
+    this.instantSearchInstance = instantsearch({
+      appId: this.appId,
+      apiKey: this.apiKey,
+      indexName: this.indexName,
+    });
```
Haroenv: the other props also need to be given here, and it needs to "restart" when they change, like in `algoliaComponent` I think

PR #426, opened 2018-04-18
feat(Hits): allow for easy slot and complete slot

Haroenv:
This allows us not to have a replacement for NoResults


docs/src/components/hits.md
```diff
@@ -0,0 +1,68 @@
...
+      <a :href="result.url">
+        {{ result.title }}
+      </a>
+    </h2>
+    <p>{{ result.description }}</p>
+  </template>
+</ais-hits>
+```
+
+When you want to override the complete hits:
```
vvo: missing example

PR #432, opened 2018-04-25
feat(search-client): Add support for Custom Search Clients

francoischalifour:
_This PR adds supports for the `search-client` prop. This allows any InstantSearch user to plug his own search client (including a custom Algolia backend)._

## Usage

```html
<template>
  <ais-index
    index-name="instant_search"
    :searchClient="searchClient"
  >
  ...
  </ais-index>
</template>

<script>
  import algoliasearch from "algoliasearch";

  export default {
    data() {
      return {
        searchClient: algoliasearch("latency", "6be0576ff61c053d5f9a3225e2a90f76")
      };
    }
  };
</script>
```

## Demo

[View Demo](https://codesandbox.io/s/vv7wr6p4q7)

(see file `src/App.vue`)

## Related feature in InstantSearch.js

> algolia/instantsearch.js#2894

### DIFF
--- a/src/components/Index.vue
+++ b/src/components/Index.vue
@@ -5,7 +5,10 @@
 </template>
 
 <script>
-import { createFromAlgoliaCredentials } from '../store';
+import {
+  createFromAlgoliaCredentials,
+  createFromAlgoliaClient,
+} from '../store';
 import algoliaComponent from '../component';
 
 export default {
@@ -16,6 +19,9 @@ export default {
       default() {
         return this._searchStore;
       },
+    },
+    searchClient: {
+      type: Object,
     },
     apiKey: {
       type: String,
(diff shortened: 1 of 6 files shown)


src/components/Index.vue
```diff
@@ -73,14 +79,46 @@ export default {
...
+        );
+      }
+
+      if (typeof this.searchClient.search !== 'function') {
+        throw new Error(
+          'vue-instantsearch: `searchClient` must implement a method `search(requests)`'
+        );
+      }
+
+      this._localSearchStore = createFromAlgoliaClient(Object.create(this.searchClient));
```
francoischalifour: `Object.create()` is not necessary here. I will remove it.

src/components/Index.vue
```diff
@@ -73,14 +79,46 @@ export default {
...
+      if (typeof this.searchClient.search !== 'function') {
+        throw new Error(
+          'vue-instantsearch: `searchClient` must implement a method `search(requests)`'
+        );
+      }
+
+      this._localSearchStore = createFromAlgoliaClient(this.searchClient);
     } else {
-      this._localSearchStore = this.searchStore;
+      if (!this.searchStore) {
```
Haroenv: can be changed to `else if` I think :)

francoischalifour: Yep, ESLint is never happy with that ahah.

src/store.js
```diff
@@ -86,7 +86,11 @@ export class Store {
     this._helper.on('result', onHelperResult.bind(this));
     this._helper.on('search', onHelperSearch.bind(this));
 
-    this._helper.getClient().addAlgoliaAgent(`vue-instantsearch ${version}`);
+    const client = this._helper.getClient();
+
+    if (typeof client.addAlgoliaAgent === 'function') {
+      client.addAlgoliaAgent(`vue-instantsearch ${version}`);
```
Haroenv: I wonder if we could add this conditionally in the helper so we can call it without checking here

Haroenv:
I wonder if this deserves a mention in the search store guide

francoischalifour:
@Haroenv I've never really used Vue InstantSearch so you decide what's better for the users! How would you see that?

We could add another section below "[Create a search store from an Algolia client instance](https://community.algolia.com/vue-instantsearch/getting-started/search-store.html#create-a-search-store-from-an-algolia-client-instance)" like "Create a search store from an custom client instance".

We'll have to document both `search()` and `searchForFacetValues()` methods.

src/store.js
francoischalifour: I'm OK with this change. It's yet another helper upgrade though haha. WDYT @bobylito?

Haroenv:
A section there seems good!

PR #437, opened 2018-04-27
feat(InfiniteHits): add widget

samouss:


src/components/InfiniteHits.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+  <div :class="suit()" v-if="state">
+    <slot :items="items" :isLastPage="isLastPage" :refine="refine">
+      <ul :class="suit('list')">
```
Haroenv: not related to this change, but I'm having second thoughts on putting results in ul/li. Gov.uk for example uses div > article, and I think that makes more sense. cc @ronanlevesque

Haroenv: I've also seen ol > li, which also makes more sense since results are ordered

src/components/__tests__/InfiniteHits.js
```diff
@@ -0,0 +1,103 @@
...
+  __setState({
+    ...defaultState,
+    showMore,
+  });
+
+  const wrapper = mount(InfiniteHits);
+
+  expect(showMore).not.toHaveBeenCalled();
+
+  wrapper.find('button').trigger('click');
```
Haroenv: would change that for the classname, so it doesn't go failing randomly if there's more than one button, or a button in the items

docs/src/components/InfiniteHits.md
```diff
@@ -0,0 +1,37 @@
...
+Displays an infinite list of hits along with a load more button.
+
+<a class="btn btn-static-theme" href="stories/?selectedKind=InfiniteHits">🕹 try out live</a>
+
+## Usage
+
+```html
+<ais-infinite-hits></ais-infinite-hits>
+```
+
```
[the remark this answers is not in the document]
samouss: Yes but we can add it later as a global guide cross InstantSearch, WDYT? Because most of the users will use a framework specific component (like [`vue-infinite-scroll`](https://github.com/ElemeFE/vue-infinite-scroll) for example).

src/components/InfiniteHits.vue
samouss: Yep why not, `ol > li` makes probably more sense because like you said results are ordered.

src/components/__tests__/InfiniteHits.js
samouss: Agree, that's the all purpose of the `testID` actually.

PR #444, opened 2018-05-08
Stop false positives in No Results Comp!

realtruckjoshdeltener:
This comp misfires a lot when loading legit results (flashes on/off), just watch searchStore.results.length to find out.  This tweak at least makes it so it will only show when you actually launched a query.


Haroenv:
Thanks for contributing! I’ll take a look what the implications of this are tomorrow

Haroenv:
Thanks again!

PR #448, opened 2018-05-17
feat(SortBy): new widget

Haroenv:
thanks for initial pairing @spinach


src/components/SortBy.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <div
+    :class="suit('')"
+    v-if="state"
+  >
+    <slot
+      :items="state.options"
+      :has-no-results="state.hasNoResults"
+      :refine="state.refine"
+      :current-refinement="state.currentRefinement"
```
samouss: I would prefer to use computed properties rather than using `state` directly from the template. I found it more explicit when you look at the code. At the end it's the same but we should agree on one or the other and implement all the components in the same way.

src/components/__tests__/SortBy.js
```diff
@@ -0,0 +1,92 @@
...
+it('renders correctly', () => {
+  __setState({ ...defaultState });
+
+  const wrapper = mount(SortBy, {
+    propsData: {
+      items: [
+        { value: 'some_index', label: 'Relevance' },
+        { value: 'some_index_cool', label: 'Coolness ascending' },
+        { value: 'some_index_quality', label: 'Quality ascending' },
+      ],
```
samouss: You can use the `defaultProps` here.

src/components/__tests__/SortBy.js
```diff
@@ -0,0 +1,92 @@
...
+    },
+    scopedSlots: {
+      default: defaultSlot,
+    },
+  });
+
+  expect(wrapper.html()).toMatchSnapshot();
+});
+
+it('behaves correctly', () => {
```
samouss: You can use a more descriptive name "behaves" is a bit vague.

src/components/__tests__/SortBy.js
```diff
@@ -0,0 +1,92 @@
...
+  });
+  // This is bad 👇🏽 but the only way for now to trigger changes
+  // on a select: https://github.com/vuejs/vue-test-utils/issues/260
+  const select = wrapper.find('select');
+  select.element.value = 'some_index_quality';
+  select.trigger('change');
+
+  expect(refine).toHaveBeenCalledTimes(1);
+  expect(refine).toHaveBeenLastCalledWith('some_index_quality');
+});
```
samouss: You can also add a test for the `selected` attribute, because for some reason it's not printed in the snapshots.

stories/SortBy.stories.js
```diff
@@ -0,0 +1,31 @@
...
+    template: `<ais-sort-by :items="[
+      { name: 'instant_search', label: 'Featured' },
+      { name: 'instant_search_price_asc', label: 'Price asc.' },
+      { name: 'instant_search_price_desc', label: 'Price desc.' },
+    ]">
+    </ais-sort-by>
+    `,
+  }))
+  .add('custom display', () => ({
+    template: `<ais-sort-by :items="[
```
samouss: We clearly should find a solution to not use template string in the stories.

docs/src/components/sort-by.md
```diff
@@ -0,0 +1,77 @@
...
+    </option>
+  </select>
+</ais-sort-by>
+```
+
+## Props
+
+Name | Type | Default | Description | Required
+---|---|---|---|---
+items | Array |  | An array of objects, each one containing a `value` and a `label` key.n | yes
```
samouss: `key.n` => `key.`

src/components/SortBy.vue
Haroenv: not really sure what the computed properties bring of value here :)

src/components/__tests__/SortBy.js
Haroenv: I don't think I use `selected`

src/components/SortBy.vue
samouss: Indeed, to me it's more explicit but it's only my opinion. At the end no matter the decision we choose but we should stick with one code style to avoid it will be easier to follow.

src/components/__tests__/SortBy.js
samouss: Yep you do on the `option`.

https://github.com/algolia/vue-instantsearch/blob/3b4b806e0a3a903f9c64a8f3a73624ff49f57967/src/components/SortBy.vue#L16-L22

PR #451, opened 2018-05-18
 feat(Stats): add more info to slot

Haroenv:
see #449 and #450 for detailed info


Haroenv:
Basically this PR is fine to merge, in a later version we can replace `Stats` completely with a component giving access to rendering arbitrary content from the raw results so it's able to handle a lot more use cases.

PR #457, opened 2018-05-30
chore: add JSX support

samouss:
We already talk about adding support of JSX to increase the readability of the stories. But there is some limitations with the JSX in Vue:

- directive doesn't work
- the slots API is not very friendly

You can check out the README of [the Babel plugin](https://github.com/vuejs/babel-plugin-transform-vue-jsx) to lean more about the differences between the "React" JSX and the "Vue" one. I only update two stories to illustrate the limitations. The PR **is not mergeable** as it is, it's only a proposal. If we choose to go for it we need to:

- update our `esling-config-algolia` to support the `eslint-plugin-vue` (even without merging this PR we should consider it)
- migrate all the stories to JSX


Haroenv:
Thanks! We definitely need to change eslint

samouss:
Indeed but we can use a plugin that shows a tab with the Vue code maybe?

samouss:
Agree to merge this PR?

Haroenv:
yes let's merge

samouss:
My bad, we need to release `eslint-config-algolia` first since we made the changes to the Vue configuration. Then we can drop the changes to the `eslintrc` in this PR, then merge.

PR #467, opened 2018-06-21
Feat(refinementList): add widget

Haroenv:
TODO:

- [x] tests
- [x] docs
- [x] rendering


src/components/SearchBox.vue
```diff
@@ -5,111 +5,30 @@
...
+      <ais-search-input
+        :refine="state.refine"
+        :placeholder="placeholder"
+        :autofocus="autofocus"
+        :show-loading-indicator="showLoadingIndicator"
+        :should-show-loading-indicator="state.isSearchStalled"
+        :submit-title="submitTitle"
+        :clear-title="clearTitle"
       >
-        <input
```
Haroenv: todo: add slots here to allow them to be overridden from parent components (same in RefinementList)

src/components/__tests__/RefinementList.js
```diff
@@ -0,0 +1,22 @@
+import { mount } from '@vue/test-utils';
+import RefinementList from '../RefinementList.vue';
+import { __setState } from '../../component';
+jest.mock('../../component');
+
+it.skip('renders correctly', () => {
```
Haroenv: todo: add tests

docs/src/components/RefinementList.md
```diff
@@ -0,0 +1,43 @@
...
+## CSS classes
+
+Here's a list of CSS classes exposed by this widget. To better understand the underlying
+DOM structure, have a look at the generated DOM in your browser.
+
+Class name | Description
+---|---
+`ais-RefinementList` | Container class
+`ais-RefinementList-item` | An item
+`ais-RefinementList-item--selected` | Selected item
```
samouss: There are some classes that not listed.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
+<template>
+  <div :class="suit('')" v-if="state">
```
samouss: The class `.ais-RefinementList--noRefinement` is missing when the component cannot be refined.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+  computed: {
+    items() {
+      if (this.searchable) {
+        return this.state.items.map(item => {
+          const highlighted = item.highlighted
+            .replace(
+              new RegExp('<em>', 'g'),
+              `<mark class="ais-Highlight-highlighted">`
+            )
+            .replace(new RegExp('</em>', 'g'), `</mark>`);
```
samouss: We can use the `Highlight` component to avoid this logic. We can fake the hit structure.

docs/src/components/RefinementList.md
```diff
@@ -0,0 +1,43 @@
...
+Name | Type | Default | Description | Required
+---|---|---|---|---
+attribute | string | | The attribute to refine on click | yes
+searchable | boolean | `false` | You can also search within the options of this | no
+operator | "or" / "and" | "or" | How to apply refinements | no
+limit | number | 10 | Number of items to show | no
+showMoreLimit | number | 20 | Number of items to show when the user clicked on “show more items” | no
+showMore | boolean | false | Whether or not to have the option to load more values | no
+sortBy | array / sort function | `['isRefined:desc', 'count:desc', 'name:asc']` | array or function to sort the results by | no
+
```
samouss: `transformItems` it not documented.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+      type: Number,
+      default: 20,
+      required: false,
+    },
+    showMore: {
+      type: Boolean,
+      default: false,
+      required: false,
+    },
+    sortBy: {
```
samouss: You can provide the type `[Array, Function]`.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
+<template>
+  <div :class="suit('')" v-if="state">
+    <slot
+      :items="items"
+      :refine="state.refine"
+      :searchable="searchable"
+      :search-for-items="state.searchForItems"
+      :show-more="showMore"
```
samouss: Don't think we need to expose it since it's only for our internal view.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+      :toggle-show-more="state.toggleShowMore"
+      :is-showing-more="state.isShowingMore"
+    >
```
samouss: There is a couple of more values provided to the render function. Why we don't forward them? Like `isFromSearch`, `canRefine`, ...

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+            </label>
+          </slot>
+        </li>
+      </ul>
+      <button
+        :class="suit('showMore')"
+        @click="state.toggleShowMore"
+        v-if="showMore"
+      >
+        Show {{state.isShowingMore ? 'less' : 'more'}}
```
samouss: Should be "Show less" / "Show more"

samouss: We can add a slot to allow the label to be customise.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+                :class="suit('checkbox')"
+                type="checkbox"
+                :value="item.value"
+                :checked="item.isRefined"
+                @change="state.refine(item.value)"
+              />
+              <template v-if="searchable">
+                <span :class="suit('labelText')" v-html="item.highlighted"></span>
+              </template>
+              <template v-else>
```
samouss: Can we inline the `v-else` on the element?

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+              </template>
+              <span :class="suit('count')">{{item.count}}</span>
+            </label>
+          </slot>
+        </li>
+      </ul>
+      <button
+        :class="suit('showMore')"
+        @click="state.toggleShowMore"
+        v-if="showMore"
```
samouss: Should be disabled when `canRefine` is `false`.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+              <template v-else>
+                <span :class="suit('labelText')">{{item.label}}</span>
+              </template>
+              <span :class="suit('count')">{{item.count}}</span>
+            </label>
+          </slot>
+        </li>
+      </ul>
+      <button
+        :class="suit('showMore')"
```
samouss: Should have the class `.ais-RefinementList-showMore--disabled` when it's disabled.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,132 @@
...
+            <label :class="suit('label')">
+              <input
+                :class="suit('checkbox')"
+                type="checkbox"
+                :value="item.value"
+                :checked="item.isRefined"
+                @change="state.refine(item.value)"
+              />
+              <template v-if="searchable">
+                <span :class="suit('labelText')" v-html="item.highlighted"></span>
```
samouss: We do not handle the no results case.

src/components/RefinementList.vue
Haroenv: it depends a bit, if you're making a custom widget that wraps it, you want to do a ternary like we do as well no?

src/components/RefinementList.vue
Haroenv: what's the difference?

src/components/RefinementList.vue
samouss: Maybe I miss something but I don't see how the implementer cannot have the value, but they will have to provide it to the slot though. You can leave as it is, doesn't matter. It's just that in the other widget when a props is only required for the view we don't pass it back to the slot.

src/components/RefinementList.vue
samouss: My bad I didn't see the "Show" before the ternary.

src/components/RefinementList.vue
Haroenv: Yes, they could pass the prop themselves too indeed. I’ll leave it out

src/components/RefinementList.vue
Haroenv: Yep, it should indeed work for this. I guess I forgot it

Haroenv:
we should be doing a `canRefine` pass on all widgets later

docs/src/components/RefinementList.md
```diff
@@ -0,0 +1,61 @@
...
+`ais-RefinementList-list` | The list of possible refinements
+`ais-RefinementList-item` | A refinement item
+`ais-RefinementList-label` | A refinement item's label
+`ais-RefinementList-labelText` | A refinement item's label text
+`ais-RefinementList-checkbox` | A refinement item's checkbox
+`ais-RefinementList-count` | A refinement item's number of matches
+`ais-RefinementList-item--selected` | Selected item
+`ais-RefinementList-searchBox` | The search box when `searchable` is true
+`ais-RefinementList-showMore` | The show more button when `show-more` is true
+`ais-RefinementList-showMore--disabled` | The show more button when `show-more` is true and it's not possible to be clicked
```
samouss: Two class names missing: 

- no refinement
- no results

src/components/RefinementList.vue
```diff
@@ -0,0 +1,150 @@
+<template>
+  <div :class="suit('')" v-if="state">
+    <slot
+      :items="state.items"
+      :refine="state.refine"
+      :searchable="searchable"
```
samouss: Don't think we need to provide the value back since the user already provide it.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,150 @@
...
+      default: 20,
+      required: false,
+    },
+    showMore: {
+      type: Boolean,
+      default: false,
+      required: false,
+    },
+    sortBy: {
+      type: [Array | Function],
```
samouss: Should be `[Array, Function]`

src/components/RefinementList.vue
```diff
@@ -0,0 +1,150 @@
...
+      get() {
+        return this.searchForFacetValuesQuery;
+      },
+      set(value) {
+        this.state.searchForItems(value);
+        this.searchForFacetValuesQuery = value;
+      },
+    },
+    toggleShowMore() {
+      return this.state.toggleShowMore || noop;
```
samouss: We need this because of the IS issue? To avoid the value to be undefined?

src/components/RefinementList.vue
```diff
@@ -0,0 +1,150 @@
...
+      :items="state.items"
+      :refine="state.refine"
+      :searchable="searchable"
+      :search-for-items="state.searchForItems"
+      :toggle-show-more="toggleShowMore"
+      :is-showing-more="state.isShowingMore"
+      :createURL="state.createURL"
+      :is-from-search="state.isFromSearch"
+      :can-refine="state.canRefine"
+      :transform-Item-To-Hit="transformItemToHit"
```
samouss: Maybe use a more meaningful name like to inform the user that is for highlighting?

src/components/RefinementList.vue
```diff
@@ -0,0 +1,150 @@
...
+    >
+      <div :class="suit('searchBox')" v-if="searchable">
+        <ais-search-input v-model="searchForFacetValues"></ais-search-input>
+      </div>
+      <ul :class="suit('list')">
+        <li
+          :class="[suit('item'), {[suit('item', 'selected')]: item.isRefined}]"
```
samouss: You can also use an expression I found it a bit more readable:

```
:class="[suit('item'), item.isRefined && suit('item', 'selected')]"
```

src/components/RefinementList.vue
```diff
@@ -0,0 +1,150 @@
...
+          v-for="item in state.items"
+          :key="item.value"
+        >
+          <slot name="item" :item="item" :refine="state.refine">
```
samouss: We need to provide `searchable`, `transformItemToHit` to the slot to allow the user to do what we do in the default render.

src/components/RefinementList.vue
```diff
@@ -0,0 +1,150 @@
...
+          </slot>
+        </li>
+      </ul>
+      <button
+        :class="[suit('showMore'), { [suit('showMore', 'disabled')]: !state.canToggleShowMore }]"
+        @click="toggleShowMore"
+        v-if="showMore"
+        :disabled="!state.canToggleShowMore"
+      >
+        <slot name="showMoreTitle" :is-showing-more="state.isShowingMore">
```
samouss: We use `label` rather than `title` on the other widgets. We can review this differences in one pass in another PR though.

src/components/RefinementList.vue
Haroenv: yes, if the value is undefined (even when v-if is false) it gives a warning to the console, so I need something which is always callable

src/components/RefinementList.vue
Haroenv: you have access to searchable through the parent (like showmore), so I removed it. transformItemToHit is now part of the items itself

Haroenv:
All comments addressed

Haroenv:
Also needs to be in a panel :)

PR #471, opened 2018-07-19
Emit Event When Refining

hecktarzuli:
This would REALLY be handy if you need to update the url or log data when someone facets.

### DIFF
--- a/src/components/RefinementList.vue
+++ b/src/components/RefinementList.vue
@@ -81,8 +81,8 @@ export default {
     toggleRefinement(value) {
       this.$emit('refined', { 
         attributeName: this.attributeName,
-        value: value.name
-      })
+        value: value.name,
+      });
       return this.searchStore.toggleFacetRefinement(
         this.attributeName,
         value.name
--- a/src/components/RefinementList.vue
+++ b/src/components/RefinementList.vue
@@ -79,7 +79,7 @@ export default {
   },
   methods: {
     toggleRefinement(value) {
-      this.$emit('refined', { 
+      this.$emit('refined', {
         attributeName: this.attributeName,
         value: value.name,
       });
--- a/src/components/RefinementList.vue
+++ b/src/components/RefinementList.vue
@@ -79,6 +79,10 @@ export default {
   },
   methods: {
     toggleRefinement(value) {
+      this.$emit('refined', { 
+        attributeName: this.attributeName,
+        value: value.name
+      })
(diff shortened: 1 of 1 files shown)


Haroenv:
In the next version we will have a proper built-in routing system, can you think of other use cases than routing for which this would be useful?

hecktarzuli:
Built-in routing system eh?  Honestly that really worries me.  I hope it's an optional feature as I would be surprised if you guys could think of all the use-cases. I assume you'd use vue-router?  I assume pager comps would force scroll to top, but facets wouldn't.  Would your comps automatically respect route params?  How would it interact with any type of store serialization or SSR? .. the questions go on and on.

Another use case (I just made up) what if someone wanted custom events to happen when you facet by certain things.  For example, if you faceted by Shoes, you could update the promo section of the page to show a shoe promo etc..   Right now the Refine List is a bit too closed box IMHO.

hecktarzuli:
Just ran into another use case.  Marketing wants to log what people faceted by in Google Analytics but trying to listen for history changes and mutations (because there is no telling when the facets will show up) just to wire a listener for the click is a giant PITA.  I'm going to have to re-create this comp just for logging to GA which is crazy.

Haroenv:
That makes sense. I would like to implement it on every widget rather than just on refinementList though. For now your fastest option will
Indeed be to copy the widget here and use it in your app.

hecktarzuli:
For those just catching up, in the meantime you can just extend the RefinementList comp and override the toggleRefinement method to emit your own event.  Something like this:
```<script>
import { RefinementList } from 'vue-instantsearch'

export default {
    extends: RefinementList,
    methods: {
        toggleRefinement(value) {          
            this.$emit('clicked', {
                attributeName: this.attributeName,
                value: value.name,
            })
            return this.searchStore.toggleFacetRefinement(this.attributeName, value.name)
        },
    },
}
</script>

Haroenv:
Yes, sorry for not getting back to this issue. The reason I didn't merge it as-is is because for the current version on Vue InstantSearch this requires a bunch of code in each widget. In the next version we can think of a more efficient way, and it will be a feature

Haroenv:
While I still think this is a useful feature at some point, we won't make it part of Vue InstantSearch v2 at this point, because most of the use cases have different paradigms that don't need to rely on events like this, since it's overall nicer IMHO to write declarative code in different ways. See also #552 for a place where you can give more feedback and use cases and we can see how to solve them properly.

PR #478, opened 2018-08-09
feat(Menu): add widget

samouss:
## Summary

This PR implements the `Menu` widget.

screen shot 2018-08-09 at 11 00 17

You can try the widget on [Storybook](https://deploy-preview-478--vue-instantsearch.netlify.com/stories/?selectedKind=Menu&full=0&addons=1&stories=1&panelRight=0).

## Questions

One thing to note is that I scoped the top level `div` of the widget under the `slot`. It avoids to still have the top level className when you completely rewrite the widget. But it introduce a top level `div` on top of the widget (see picture below). It can be removed when Vue will support a ["Fragment"](https://github.com/vuejs/vue/issues/7088) like component. Do we want to implement this behaviour on all the widgets?

**With default render**

screen shot 2018-08-09 at 10 56 49 1

**With custom render**

screen shot 2018-08-09 at 10 57 35

### DIFF
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -53,29 +53,29 @@ export default {
       type: Number,
       default: 10,
     },
+    showMoreLimit: {
+      type: Number,
+      default: 20,
+    },
     showMore: {
       type: Boolean,
       default: false,
     },
-    showMoreLimit: {
-      type: Number,
-    },
     sortBy: {
+      // type: Array || Function
       default() {
-        return ['isRefined:desc', 'count:desc', 'name:asc'];
+        return ['name:asc', 'count:desc'];
       },
     },
+
+    // How to implement those values?
     showMoreLabel: {
       type: String,
-      default() {
-        return 'Show more';
-      },
+      default: 'Show more',
     },
     showLessLabel: {
       type: String,
-      default() {
-        return 'Show less';
-      },
+      default: 'Show less',
     },
   },
   computed: {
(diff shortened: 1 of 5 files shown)


src/components/Menu.vue
```diff
@@ -1,46 +1,51 @@
 <template>
-  <div :class="suit()" v-if="state">
```
Haroenv: Does this fit is.css?

src/components/__tests__/__snapshots__/menu.js.snap
```diff
@@ -1,82 +1,557 @@
...
-      </a>
-    </li>
-  </ul>
-  <button class="ais-Menu-showMore">
-    Show more
-  </button>
+exports[`custom default render renders correctly 1`] = `
+
+<div>
+  <div class>
```
Haroenv: Empty class, but not an issue I guess

src/components/Menu.vue
samouss: Yes it fits since the full DOM widget is under the `slot`. It's just that we have now an extra (useless) `div` on top of the widget.

src/components/__tests__/__snapshots__/menu.js.snap
samouss: Yep but we do not care that much since it's the result of the custom render. I add a class only when `canRefine` is false.

samouss:
We didn't pick the solution with the top level `div` for the moment, since the extra `div` could be a bit problematic for users. We can review this proposal once the Fragment are available.

PR #483, opened 2018-08-10
fix(Menu): update the default sort

samouss:


samouss:
I took the same behaviour than React, it doesn't sortBy refined for the Menu.

PR #484, opened 2018-08-13
feat(RangeInput): implement new widget

bobylito:


.github/V2_PROGRESS.md
```diff
@@ -40,8 +40,8 @@ x => done
 * [~] `ais-panel`
   * Just header & footer slot, doesn't get any child info
 * [x] `ais-powered-by`
-* [~] `ais-range-input` Alex
-* [~] `ais-range-slider` Alex
+* [x] `ais-range-input`
+* [-] `ais-range-slider` Alex
```
Haroenv: this is by accident I guess?

docs/src/components/RangeInput.md
```diff
@@ -0,0 +1,46 @@
...
+---
+
+The range input widget allows a user to select a numeric range using a minimum and maximum input.
+
+<a class="btn btn-static-theme" href="stories/?selectedKind=RangeInput">🕹 try out live</a>
+
+## Usage
+
+```html
+<ais-range-input :attributeName="attribute"></ais-range-input>
```
Haroenv: should be `attribute`

.github/V2_PROGRESS.md
bobylito: I thought that when this PR is merged it's gonna be done. So I went ahead and marked as done.

docs/src/components/RangeInput.md
bobylito: Erf oki doc :)

bobylito:
Done @Haroenv

src/components/RangeInput.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+<div v-if="state" :class="suit()">
+  <form :class="suit('form')" @submit.prevent="refine()">
```
samouss: We should have a slot to support customization

src/components/RangeInput.vue
```diff
@@ -0,0 +1,66 @@
...
+    <label :class="suit('label')">
+      <input :class="[suit('input'), suit('input', 'min')]" type="number" ref="minInput" :min="this.min" :placeholder="this.min"/>
```
samouss: We should put the `min` & `max` restrictions on both input

src/components/RangeInput.vue
```diff
@@ -0,0 +1,66 @@
...
+        min: this.min,
+        max: this.max,
+        precision: this.precision,
+      };
+    },
+  },
+  methods: {
+    refine() {
+      const minValue = this.$refs.minInput.value && parseInt(this.$refs.minInput.value, 10);
+      const maxValue = this.$refs.maxInput.value && parseInt(this.$refs.maxInput.value, 10);
```
samouss: We should avoid to rely on `$refs` because in case the state is already defined (e.g. routing) the value won't be filled. We should rely on computed properties for the min & max.

src/components/RangeInput.vue
bobylito: Does it mean that we should have dynamic setting based on the value of the other input?

src/components/RangeInput.vue
bobylito: A global one?

src/components/RangeInput.vue
bobylito: I will need your help on that 😅 Thursday?

src/components/RangeInput.vue
samouss: Yes a default one rather than global depends the definition of global in your question. You can take a look at other [widget implementation](https://github.com/algolia/vue-instantsearch/blob/bb011b70599b3fc1b72b225489d49e1c8e761dbe/src/components/SearchBox.vue#L3-L6).

src/components/RangeInput.vue
samouss: It's not mandatory but we can at least provide the range to min and max. Both inputs will be restricted on the range rather than only have the restriction on min or max.

src/components/RangeInput.vue
samouss: Yep! In the meantime you can also check the implementation of other widgets. The SearchBox is a good example since it also deals with [an input](https://github.com/algolia/vue-instantsearch/blob/feat/connectors/src/components/SearchBox.vue#L152-L161).

PR #486, opened 2018-08-13
POC: widget params update

samouss:
This PR aims to propose a solution for avoid to trigger useless update of the widget (ie remove / add). By default the mixin component watch the `widgetParams` object. On each change of one of its dependencies [the callback](https://github.com/algolia/vue-instantsearch/blob/9ac978cb0766889aa5e1b2818ea91abddf1dcb47/src/component.js#L30-L35) is triggered. This issue with that is that every reference that change between renders will trigger the watcher (e.g an inline array provided as prop).

A possible solution is to let the widget have the "control" on when it needs to be updated. To do so we can perform extra check to verify that the attribute **have** really changed. This naive implementation use an object `widgetParams`  that defines predicate to run once the watcher is called. The huge downside is that we duplicate the map for the `widgetParams` at two different place (on the component description + computed properties).

```js
export default {
  props: {
    attribute: {
      type: String,
      required: true,
    },
    items: {
      type: Array,
      required: true,
    },
    transformItems: {
      type: Function,
      required: false,
      default: items => items,
    },
  },
  widgetParams: {
    attributeName: (previous, next) => previous !== next,
    options: (previous, next) => !isEqual(previous, next),
    transformItems: (previous, next) => previous !== next,
  },
  computed: {
    widgetParams() {
      return {
        attributeName: this.attribute,
        transformItems: this.transformItems,
        options: this.items,
      };
    },
  },
};
```

There is [an other proposal](https://github.com/algolia/vue-instantsearch/pull/487) a bit more powerful.


Haroenv:
We will look back at these PRs when it proves to be a problem

PR #487, opened 2018-08-13
POC: widget params update [alt]

samouss:
This PR aims to propose an other solution to the update `widgetParams` issue ([see more](https://github.com/algolia/vue-instantsearch/pull/486)). This proposal rewrite a bit the `component` mixin to derived the `widgetParams` from the component definition. The benefit of this solution is that we "encapsulate" the name / value / validation of the `widgetParams` at the same place. The mixin that is used is the not the component one but the one called [widget](https://github.com/algolia/vue-instantsearch/blob/557ee64effa4e9f5a9bb85ab2e0c26b5378a0b63/src/mixins/widget.js).

```js
export default {
  props: {
    attribute: {
      type: String,
      required: true,
    },
    items: {
      type: Array,
      required: true,
    },
    transformItems: {
      type: Function,
      required: false,
      default: items => items,
    },
  },
  widgetParams: {
    attributeName: {
      getValue: vm => vm.attribute,
    },
    options: {
      getValue: vm => vm.items,
      shouldUpdate: (previous, next) => !isEqual(previous, next),
    },
    transformItems: {
      getValue: vm => vm.transformItems,
    },
  },
};
```


Haroenv:
I think I like this version more, since `shouldUpdate` can be added as needed. Another option that I see which could be done is the following:

```js
computed.widgetParams = () => {
  return [
    {
      name: 'someAttr',
      value: this.props.whateverItMapsTo,
      shouldUpdate: (prev, next) => !isEqual(preview, next),
    },
    {
      name: 'otherAttr',
      value: this.props.whateverElseItMapsTo,
    },
  ];
};
```

Then to derive the actual parameters from that, it's simply a reduce, rather then to have the annoying Object.entries (which needs to be polyfilled)

samouss:
Which part requires `Object.entires`? We can probably stick with what we have now (not this PR) and iterate on a solution like this once we have an issue. Make sense?

src/mixins/widget.js
```diff
@@ -0,0 +1,64 @@
...
+        }),
+        {}
+      );
+    },
+  },
+  watch: {
+    widgetParams(next, previous) {
+      const { widgetParams } = this.$options;
+
+      const shouldUpdateWidget = Object.keys(widgetParams).some(name => {
```
Haroenv: this is the Object.keys I meant (could have used entries too). My point was that if the thing was already and array, it would be slightly easier to iterate over it, but not a big issue at all

bobylito:
> Which part requires Object.entires? We can probably stick with what we have now (not this PR) and iterate on a solution like this once we have an issue. Make sense?

I don't have much insights here, but just wanted to say that it makes the API for creating widgets quite clear. Also as far as I understand the `shouldUpdate` makes it the update of the widget explicit. I'm wondering if this concept is clear since it comes from an internal limitation...

src/component.js
```diff
@@ -27,11 +27,25 @@ export default {
     this.instantSearchInstance.removeWidget(this.widget);
   },
   watch: {
-    widgetParams(newVal) {
-      const oldWidget = this.widget;
-      this.widget = this.widgetFactory(newVal);
-      this.instantSearchInstance.addWidget(this.widget);
-      this.instantSearchInstance.removeWidget(oldWidget);
+    widgetParams(next, previous) {
```
Haroenv: should this whole file go away then?

Haroenv:
We will look back at these PRs when this proves to be an issue

PR #488, opened 2018-08-14
feat(Panel): implement widget & mixin

samouss:
**Summary**

This PR implements the Panel widget and the two mixins (provider / consumer) to bubble up the `canRefine` informations from the widget to the Panel. We only support this information at the moment we can imaginee to provide more informations in the future. 

The "magic" is implemented inside [the panel](https://github.com/algolia/vue-instantsearch/compare/feat/connectors...feat/panel?expand=1#diff-761c19a861be254b1425ba59c4454db2) mixin. Happy to discuss if you have a better idea of implementation! To keep the PR small enough I've only implemented the Panel for the Menu. The rest of the widgets will come in another PR.

**Usage (for implementers)**

```js
import { createPanelConsumerMixin } from "../panel";
import algoliaComponent from "../component";

export default {
  mixins: [
    algoliaComponent,
    createPanelConsumerMixin({
      mapStateToCanRefine: state => state.canRefine
    })
  ]
  //  ...
};
```

**Usage (for users)**

```html
<ais-panel>
  <template slot="header">Menu</template>
  <ais-menu attribute="brand" />
  <template slot="footer">Footer</template>
</ais-panel>
```


src/components/Panel.vue
```diff
@@ -0,0 +1,31 @@
...
+export default {
+  mixins: [createPanelProviderMixin()],
+  data() {
+    return {
+      widgetName: 'Panel',
+    };
+  },
+  methods: {
+    suit(...args) {
+      return suit(this.widgetName, ...args);
```
Haroenv: when this PR and #485 are merged, I'll make a mixin for the class names

src/panel.js
```diff
@@ -0,0 +1,76 @@
+import Vue from 'vue';
+
+export const PANEL_EMITTER_NAMESPACE = 'instantSearchPanelEmitter';
+export const PANEL_CHANGE_EVENT = 'PANEL_CHANGE_EVENT';
+
+export const createPanelProviderMixin = () => ({
```
Haroenv: would be a bit less consistent, but this can just be an object right? Mixins are copied over so it's safe to use the same reference

src/panel.js
```diff
@@ -0,0 +1,76 @@
...
+    this.emitter.$destroy();
+  },
+  methods: {
+    updateCanRefine(value) {
+      this.canRefine = value;
+    },
+  },
+});
+
+export const createPanelConsumerMixin = ({ mapStateToCanRefine }) => ({
```
Haroenv: maybe this can have a default like `({ mapStateToCanRefine = state => state.canRefine } = {})`?

Or is there a reason you didn't?

stories/Panel.stories.js
```diff
@@ -0,0 +1,37 @@
+import { storiesOf } from '@storybook/vue';
+import { previewWrapper } from './utils';
+
+storiesOf('Panel', module)
+  .addDecorator(previewWrapper())
+  .add('default', () => ({
+    template: `
+      <ais-panel>
+        This is the body of the Panel.
```
Haroenv: there should be some examples with InstantSearch widgets as example I think

src/components/__tests__/Menu.js
```diff
@@ -293,6 +294,27 @@ describe('default render', () => {
 
     expect(toggleShowMore).toHaveBeenCalledTimes(1);
   });
+
+  it('calls the Panel mixin with `canRefine`', () => {
```
Haroenv: looks like we're not actually testing the mixin, but that indeed seems hard to attain

Haroenv: actually, we are testing it enough in panel so it's ok

src/components/Panel.vue
samouss: Was on my list too!

src/panel.js
samouss: Yep indeed it can, it was only for consistency with the other.

src/panel.js
samouss: The param is required so it cannot have a false positive (e.g. always return falsy value). I would prefer to keep it explicit for now until the connector always provide the `canRefine` value.

src/components/__tests__/Menu.js
samouss: Yep indeed I'm only testing the function mapping function here.

stories/Panel.stories.js
samouss: Each widgets will have an example with the Panel, but it's in the next PR. You can take a look at [Menu](https://github.com/algolia/vue-instantsearch/blob/56e8ddef6c05e9fc7efadc1fa6b0df38ef0e8d24/stories/Menu.stories.js#L50-L57) for example.

PR #490, opened 2018-08-16
feat(ToggleRefinement): implement widget

samouss:
**Summary**

This PR implements the `ToggleRefinement` widget.

screen shot 2018-08-16 at 12 57 51

You can try the widget on [Storybook]().

### DIFF
--- a/docs/src/components/ToggleRefinement.md
+++ b/docs/src/components/ToggleRefinement.md
@@ -12,7 +12,7 @@ githubSource: docs/src/components/ToggleRefinement.md
 The toggle widget lets the user either:
 
 * switch between two values for a single facetted attribute (free_shipping / not_free_shipping)
-* toggle a faceted value on and off (only ‘canon’ for brands)
+* toggle a faceted value on and off (only "Canon" for brands)
 
 This widget is particularly useful if you have a boolean value in the records.
 
--- a/docs/src/components/ToggleRefinement.md
+++ b/docs/src/components/ToggleRefinement.md
@@ -1,0 +1,53 @@
+---
+title: ToggleRefinement
+mainTitle: Components
+layout: main.pug
+category: Components
+withHeadings: true
+navWeight: 6
+editable: true
+githubSource: docs/src/components/ToggleRefinement.md
+---
+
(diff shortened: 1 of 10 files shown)


docs/src/components/ToggleRefinement.md
```diff
@@ -0,0 +1,53 @@
...
+withHeadings: true
+navWeight: 6
+editable: true
+githubSource: docs/src/components/ToggleRefinement.md
+---
+
+The toggle widget lets the user either:
+
+* switch between two values for a single facetted attribute (free_shipping / not_free_shipping)
+* toggle a faceted value on and off (only ‘canon’ for brands)
```
Haroenv:
```diff
- only ‘canon’ for brands
+ only "Canon" for brands 
```

Clearer that it's a brand with the capital letter

src/components/ToggleRefinement.vue
```diff
@@ -0,0 +1,86 @@
...
+      type: String,
+      required: true,
+    },
+    label: {
+      type: String,
+      required: true,
+    },
+    on: {
+      type: [String, Number, Boolean],
+      required: false,
```
Haroenv: Did you make it so neither are required so that you can choose to only have off? 

I'd add a `validator` that checks that at least one of both is given, does that make sense?

src/components/ToggleRefinement.vue
```diff
@@ -0,0 +1,86 @@
...
+      default: true,
+    },
+    off: {
+      type: [String, Number, Boolean],
+      required: false,
+      // explicit otherwise Vue coerce the default value
```
Haroenv: Vue `coerces`

src/components/ToggleRefinement.vue
samouss: With the current implementation of the connector none is required. By default the `on` value is `true` and the `off` value is nothing. It means that you can provide none of them, `on`, `off` or both at the same time. That's why I put the `on` as non required. Is it more clear?

Haroenv: ah yes, with a default value it makes sense

PR #492, opened 2018-08-16
fix(RangeInput): provide default slot for customization

bobylito:
This is a refactor of the previous implementation. It relies on the intermediate value in the component as the search experience of this widget is asynchronous (you have to validate the form to do a new search). One of the particularities is that I had to implement two update functions to ease the implementation on the slot side. Also the flow is a little bit special:
 - currentRefinements / state.start should always be used as the source of truth for re-rendering
 - the updateMin / updateMax will update a internal currentRefinements
 - the refine takes the internal currentRefinements and use it to update the state. At this point the stat.start will be in-sync with the UI

Let me know if you have other questions.


stories/RangeInput.stories.js
```diff
@@ -14,4 +14,22 @@ storiesOf('RangeInput', module)
   }))
   .add('with min and max', () => ({
     template: `<ais-range-input attribute="price" :min=10 :max=50></ais-range-input>`,
+  }))
+  .add('using the default slot', () => ({
+    template: `
+      <ais-range-input attribute="price">
+        <template slot-scope="{refine, updateMin, updateMax, currentRefinements}">
+          <form  @submit.prevent="refine()" >
```
Haroenv: hmm, I don't understand why it has no arguments here

stories/RangeInput.stories.js
```diff
@@ -14,4 +14,22 @@ storiesOf('RangeInput', module)
...
+            <label> 
+              <input type="number" :max="this.max"  :placeholder="this.max" :value="currentRefinements && currentRefinements[0]" @change="updateMin($event.currentTarget.value)"/>
```
Haroenv: did you change currentRefinements?

stories/RangeInput.stories.js
bobylito: The update and the search are done in two steps. I use updateMin and updateMax to update the value and refine to refine the search state with the updated values and trigger the search.

stories/RangeInput.stories.js
bobylito: I'm not sure I get your question. There is no currentRefinements in the connector. This is the values that you get in the slots and that are provided by the connector (start property)

src/components/RangeInput.vue
```diff
@@ -1,18 +1,16 @@
...
-    :updateMin="this.updateMin"
-    :updateMax="this.updateMax"
     :currentRefinements="this.state.start"
     :refine="this.refine"
   >
-  <form :class="suit('form')" @submit.prevent="refine()">
+  <form :class="suit('form')" @submit.prevent="refine(minInput, maxInput)">
     <label :class="suit('label')">
-      <input :class="[suit('input'), suit('input', 'min')]" type="number" :min="this.min"  :placeholder="this.min" :value="this.state.start && this.state.start[0]" @change="updateMin($event.currentTarget.value)"/>
+      <input :class="[suit('input'), suit('input', 'min')]" type="number" :min="this.min"  :placeholder="this.min" :value="this.state.start && this.state.start[0]" @change="minInput = $event.currentTarget.value"/>
```
Haroenv: last Sunday I discovered `valueAsNumber` is a thing! https://www.w3.org/TR/2011/WD-html5-20110405/common-input-element-attributes.html#dom-input-valueasnumber it's not supported on IE11 though for example :(

bobylito: That's amazing! :) Should I implement it then, even with ie11?

Haroenv: no, you should probably not use it, because the support is hard to find, while the value is fairly minimal

src/components/RangeInput.vue
```diff
@@ -1,15 +1,20 @@
...
 <div v-if="state" :class="suit()">
-  <form :class="suit('form')" @submit.prevent="refine()">
+  <slot
+    :currentRefinements="this.state.start"
+    :refine="this.refine"
+  >
+  <form :class="suit('form')" @submit.prevent="refine(minInput, maxInput)">
     <label :class="suit('label')">
-      <input :class="[suit('input'), suit('input', 'min')]" type="number" ref="minInput" :min="this.min" :placeholder="this.min"/>
+      <input :class="[suit('input'), suit('input', 'min')]" type="number" :min="this.min"  :placeholder="this.min" :value="this.state.start && this.state.start[0]" @change="minInput = $event.currentTarget.value"/>
```
samouss: The `min` & `max` value should be the one of the range rather than the provided values. We can also use this range to set the placeholder (see [React](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RangeInput&selectedStory=default&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs) & [IS](https://community.algolia.com/instantsearch.js/v2/dev-novel/?selectedStory=RangeInput.default) behaviour).

src/components/RangeInput.vue
```diff
@@ -1,15 +1,20 @@
...
     </label>
     <span :class="suit('separator')">to</span>
     <label :class="suit('label')">
-      <input :class="[suit('input'), suit('input', 'max')]" type="number" ref="maxInput" :max="this.max" :placeholder="this.max"/>
+      <input :class="[suit('input'), suit('input', 'max')]" type="number" :max="this.max"  :placeholder="this.max" :value="this.state.start && this.state.start[1]" @change="maxInput = $event.currentTarget.value"/>
     </label>
     <button :class="suit('submit')" type="submit">Go</button>
```
samouss: We can use a slot to allow the customisation of the label.

src/components/RangeInput.vue
```diff
@@ -48,19 +55,20 @@ export default {
...
       };
     },
   },
   methods: {
-    refine() {
-      const minValue = this.$refs.minInput.value && parseInt(this.$refs.minInput.value, 10);
-      const maxValue = this.$refs.maxInput.value && parseInt(this.$refs.maxInput.value, 10);
+    refine(min, max) {
+      const minValue = min && parseInt(min, 10);
+      const maxValue = max && parseInt(max, 10);
```
samouss: We don't need to parse the value the logic is already handled by the connector.

src/components/RangeInput.vue
```diff
@@ -1,15 +1,20 @@
...
+  >
+  <form :class="suit('form')" @submit.prevent="refine(minInput, maxInput)">
     <label :class="suit('label')">
-      <input :class="[suit('input'), suit('input', 'min')]" type="number" ref="minInput" :min="this.min" :placeholder="this.min"/>
+      <input :class="[suit('input'), suit('input', 'min')]" type="number" :min="this.min"  :placeholder="this.min" :value="this.state.start && this.state.start[0]" @change="minInput = $event.currentTarget.value"/>
     </label>
     <span :class="suit('separator')">to</span>
     <label :class="suit('label')">
-      <input :class="[suit('input'), suit('input', 'max')]" type="number" ref="maxInput" :max="this.max" :placeholder="this.max"/>
+      <input :class="[suit('input'), suit('input', 'max')]" type="number" :max="this.max"  :placeholder="this.max" :value="this.state.start && this.state.start[1]" @change="maxInput = $event.currentTarget.value"/>
```
samouss: We can add the `step` on both inputs.

src/components/RangeInput.vue
```diff
@@ -1,15 +1,20 @@
 <template>
 <div v-if="state" :class="suit()">
-  <form :class="suit('form')" @submit.prevent="refine()">
+  <slot
+    :currentRefinements="this.state.start"
```
samouss: `this` can be omit in the template

src/components/RangeInput.vue
samouss: The behaviour is different from the React & IS version, the values should be clamped to the range. It's not implemented in the connector for backward compatibility issue. But the behaviour exists in the [widget](https://github.com/algolia/instantsearch.js/blob/cd8f17ecf5b6e7aef14a54a9bec51fa8cd167f69/src/widgets/range-input/range-input.js#L42-L45).

docs/src/components/RangeInput.md
```diff
@@ -43,4 +43,10 @@ Class name | Description
 `ais-RangeInput-input` | An input
```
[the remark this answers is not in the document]
bobylito: I will remove the old file.

src/components/RangeInput.vue
```diff
@@ -1,15 +1,20 @@
...
+    :refine="this.refine"
```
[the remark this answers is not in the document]
bobylito: Is that for implementing a range slider?

src/components/RangeInput.vue
bobylito: Do you want an empty slot for strictly adding a label or do you want a slot on `<label>`?

src/components/RangeInput.vue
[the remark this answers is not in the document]
samouss: The connector expose the actual `range` available for the refinement, internally we should use it for the placeholder, min & max. We can also expose this range to the slot in order to allow the user to use it.

src/components/RangeInput.vue
samouss: Rather than an hard coded value for the text we can create a slot under the button to allow the user to customise it.

```html
<button :class="suit('submit')" type="submit">
  <slot name="nameOfTheSlots">Go</slot>
</button>
```

src/components/RangeInput.vue
samouss: It's already kinda configurable since we exposed the precision property on the connector we can use this value to compute the `step`. We already do it in the [RIS](https://github.com/algolia/react-instantsearch/blob/b37bf5aea9aacff87a85606ae4dc8b2a0812124c/packages/react-instantsearch-dom/src/components/RangeInput.js#L93) & [IS](https://github.com/algolia/instantsearch.js/blob/62bac743ab3aa679a4e5f10c6e8534b43e80ffcf/src/widgets/range-input/range-input.js#L39) implementation.

src/components/RangeInput.vue
```diff
@@ -1,15 +1,20 @@
 <template>
 <div v-if="state" :class="suit()">
```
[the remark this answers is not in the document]
samouss: It's there when the refinement is not possible.

stories/RangeInput.stories.js
```diff
@@ -14,4 +17,28 @@ storiesOf('RangeInput', module)
...
   .add('with min and max', () => ({
     template: `<ais-range-input attribute="price" :min=10 :max=50></ais-range-input>`,
+  }))
+  .add('using the default slot', () => ({
+    template: `
+      <ais-range-input attribute="price">
+        <template slot-scope="{refine, currentRefinements}">
+          <form  @submit.prevent="refine(min, max)" >
+            <label>
+              <input type="number" :max="this.max"  :placeholder="this.max" :value="currentRefinements && currentRefinements[0]" @change="min = $event.currentTarget.value"/>
```
Haroenv: no big deal at all, but if it gets over 80 lines I personally move the props to their own lines.

Haroenv:
Didn't realise, but the tests are failing now https://travis-ci.org/algolia/vue-instantsearch/builds/420008852#L534

bobylito:
> Didn't realise, but the tests are failing now https://travis-ci.org/algolia/vue-instantsearch/builds/420008852#L534

On it

stories/RangeInput.stories.js
bobylito: I was wondering about that

src/components/RangeInput.vue
```diff
@@ -48,21 +80,33 @@ export default {
...
-        this.$refs.maxInput.value && parseInt(this.$refs.maxInput.value, 10);
-
-      this.state.refine([minValue, maxValue]);
+    refine(min, max) {
+      this.state.refine([min, max]);
     },
   },
-};</script>
+};
+</script>
```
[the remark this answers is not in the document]
bobylito: My bad, I used prettier in my text editor without the eslint config.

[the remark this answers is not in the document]
Haroenv: Yep, me too, it's really annoying, but `yarn lint --fix` before committing helps

PR #494, opened 2018-08-20
feat(HierarchicalMenu): implement widget

samouss:
**Summary**

This PR implements the `HierarchicalMenu` widget. The PR also adds the ability to override the filters section of the preview wrapper from the stories. I've updated the `Breadcrumb` stories to use a `HierarchicalMenu` rather than the `Configure` widget.

I've also updated the Jest plugin that preprocess the Vue files from [Jest Vue Preprocessor](https://github.com/vire/jest-vue-preprocessor) to [Vue Jest](https://github.com/vuejs/vue-jest). The latter has a better support for functional components (it's the official plugin for Jest).

screen shot 2018-08-20 at 16 10 45

You can use the widget on [Storybook](https://deploy-preview-494--vue-instantsearch.netlify.com/stories/?selectedKind=HierarchicalMenu).

### DIFF
--- a/src/components/HierarchicalMenu.vue
+++ b/src/components/HierarchicalMenu.vue
@@ -1,19 +1,49 @@
 <template>
-  <div>
-    <slot v-bind="state">
-      <json-tree :level="2" :data="state"></json-tree>
+  <div
+    v-if="state"
+    :class="[suit(''), !canRefine && suit('', 'noRefinement')]"
+  >
+    <slot
+      :items="items"
+      :can-refine="canRefine"
+      :can-toggle-show-more="canToggleShowMore"
+      :is-showing-more="isShowingMore"
+      :refine="state.refine"
+      :create-URL="state.createURL"
+      :toggle-show-more="toggleShowMore"
+    >
+      <ais-hierarchical-menu-list
+        :items="items"
+        :level="0"
+        :suit="suit"
+        :refine="state.refine"
+        :createURL="state.createURL"
+      />
+
+      <button
+        v-if="showMore"
+        :class="[suit('showMore'), !canToggleShowMore && suit('showMore', 'disabled')]"
+        :disabled="!canToggleShowMore"
+        @click.prevent="toggleShowMore"
+      >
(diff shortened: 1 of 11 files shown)


Haroenv:
The chevron that points to the right to show if there's any children on a level doesn't seem to be here; is that on purpose?

samouss:
Yes because we don't have the informations until we click on the facet.

src/components/HierarchicalMenu.vue
```diff
@@ -1,55 +1,85 @@
...
-import algoliaComponent from '../component';
 import { connectHierarchicalMenu } from 'instantsearch.js/es/connectors';
+import algoliaComponent from '../component';
+import HierarchicalMenuList from './HierarchicalMenuList.vue';
 
 export default {
-  components: { 'json-tree': JsonTree },
   mixins: [algoliaComponent],
+  components: {
+    'ais-hierarchical-menu-list': HierarchicalMenuList,
```
Haroenv: if you imported it as `AisHierarchicalMenuList`, the mapping would be unnecessary and you could have written `components: { AisHierarchicalMenuList }`

stories/utils.js
```diff
@@ -22,6 +22,10 @@ export const previewWrapper = ({
       </li>
     </ol>
   `,
+  filters = `
+    <!-- @TODO: replace with a RefinementList  -->
+    <ais-menu attribute="brand" />
```
Haroenv: nice idea, but why does it even need a default?

src/components/HierarchicalMenu.vue
samouss: Yep indeed but I prefer to keep the same name for the import. Since this widget is a private API we can remove the prefix and use the component as it is `components: { HierarchicalMenuList }`. WDYT?

stories/utils.js
samouss: Mainly because in a couple of widgets like `CurrentRefinements`, `ClearRefinements` even `Hits` it's interesting to have more refinement capabilities to test with.

PR #496, opened 2018-08-27
fix(SearchBox): hide reset by default

samouss:
**Summary**

This PR hides the reset button when the query is empty (or the search is stalled). The behaviour match what we have on the other flavours ([React](https://community.algolia.com/react-instantsearch/storybook/?knob-showLoadingIndicator=true&selectedKind=SearchBox&selectedStory=default&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs) & [Vanilla](https://community.algolia.com/instantsearch.js/v2/dev-novel/?selectedStory=SearchBox.default) at least).

**Before**

before

**After**

after

You can use play with the widget on [Storybook](https://deploy-preview-496--vue-instantsearch.netlify.com/stories/?selectedKind=SearchBox&selectedStory=default&full=0&addons=1&stories=1&panelRight=0).


src/components/SearchBox.vue
```diff
@@ -53,6 +53,7 @@
           type="reset"
           :title="clearTitle"
           :class="suit('reset')"
+          :hidden="!currentRefinement || state.isSearchStalled"
```
Haroenv: why does it need to be hidden if the search is stalled?

samouss: Don't know but it's the behaviour on the other implementation (thoughts @bobylito?). We can merge it as it is (consistency). But we can review the behaviour latter on all the implem, WDYT?

Haroenv: yes, that makes sense

PR #498, opened 2018-08-27
fix(RangeInput): clamp values on the range

samouss:
**Summary**

This PR removes a warning that was display because of the `-Infinity / Infinity` value provided to the number inputs. The PR also align the behaviour with [React](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RangeInput&selectedStory=with%20min%20boundaries&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs) & [Vanilla](https://community.algolia.com/instantsearch.js/v2/dev-novel/?selectedStory=RangeInput.with%20min%20boundary). In case the value and the range are equal we don't display the value in the input. Last thing we use directly the range for min, max and placeholder since the logic to pick the correct value is already in the connector.

**Before**

before

**After**

after

You can use play with the widget on [Storybook]().


samouss:
Agree we should review the connector implementation in the future.

PR #499, opened 2018-08-27
chore(eslint): update ESLint plugins

samouss:
**Summary**

This PR updates all the ESLint plugins + our own configuration that handle Vue files a bit better. List of changes:

- pinned the devDependencies - once the BETA is on `master` we should setup Renovate
- added default props that were absent - we already have default value in the connector but those props can be used as "documentation" for the Vue users that are not aware of the connectors
- ran Prettier & fixed the issues related to the update of ESLint plugins


.eslintrc.js
```diff
@@ -3,6 +3,13 @@ module.exports = {
   rules: {
     'no-warning-comments': 'warn', // we have many Todo:, this will remind us to deal with them
     'no-use-before-define': 'off',
+    'vue/attribute-hyphenation': [
+      'error',
+      'always',
+      {
+        ignore: ['createURL'],
```
samouss: I've ignore this one for simplicity and coherence between the connector & Vue component. Otherwise the prop name would be `createUrl` rather than `createURL`.

src/components/PoweredBy.vue
```diff
@@ -1,20 +1,33 @@
...
-</div>
+  <div :class="suit()">
+    <a
+      :class="suit('link')"
+      :href="algoliaUrl"
+      target="_blank"
+      rel="noopener"
+      aria-label="search by Algolia"
+    >
+      <!-- TODO: remove inline style when correct CSS is released -->
```
Haroenv: https://github.com/algolia/instantsearch.css/pull/57, but I wonder if it's already released, anyway, this comment should have the issue

src/components/SearchBox.vue
```diff
@@ -81,9 +88,20 @@
...
-                  <circle stroke-opacity=".5" cx="18" cy="18" r="18"/>
+              <g
+                fill="none"
+                fill-rule="evenodd"
+              >
+                <g
+                  transform="translate(1 1)"
+                  stroke-width="2"
+                >
+                  <circle
```
Haroenv: kind of sad that this doesn't work like prettier but oh well

Haroenv: is there an option for minimal # of props before it's moved vertically?

samouss: Yes it's [1](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/max-attributes-per-line.md) by default.

src/components/PoweredBy.vue
samouss: Don't think it's released yet.

src/components/SearchBox.vue
Haroenv: probably too tricky (with some longer attributes), but I would prefer `singleline: 2` I think

samouss: This PR doesn't update the general configuration, we can create a PR on [eslint-config-algolia](https://github.com/algolia/eslint-config-algolia) for this.

PR #500, opened 2018-08-27
fix(component): avoid update on first render

samouss:
**Summary**

This PR update the `component` mixin to only apply the state update on the second render. The first render is mainly triggered to create the DOM / setup listeners. Since Vue handled that for us it can leads to some issue (like flash of content). e.g. the first render displays the empty state - then the search is will trigger the second render with the "real" state. In the Vanilla implementation we early return to avoid the render (see [RefinementList](https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/refinement-list/refinement-list.js#L46-L54), [Menu](https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/menu/menu.js#L40-L48), [RangeInput](https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/range-input/range-input.js#L26-L34)).

**Before**

stats-before
range-before

**After**

stats-after
range-after

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -40,9 +40,8 @@ export default {
     },
     updateState(state = {}, isFirstRender) {
       if (!isFirstRender) {
-        // Avoid to update the state on first render
-        // otherwise we have a flash from empty state
-        // to the next state
+        // Avoid updating the state on first render
+        // otherwise there will be a flash of placeholder data
         this.state = state;
       }
     },


src/__tests__/component.js
```diff
@@ -0,0 +1,162 @@
...
+  localVue.component('Test', {
+    render: () => null,
+  });
+
+const createFakeInstance = () => ({
+  addWidget: jest.fn(),
+  removeWidget: jest.fn(),
+});
+
+it('adds a wdiget on create', () => {
```
Haroenv: widget

src/__tests__/component.js
```diff
@@ -0,0 +1,162 @@
...
+      widgetParams,
+    }),
+  });
+
+  expect(connector).toHaveBeenCalled();
+  expect(factory).toHaveBeenCalledWith(widgetParams);
+  expect(instance.addWidget).toHaveBeenCalledWith(widget);
+});
+
+it('removes a wdiget on destroy', () => {
```
Haroenv: idem

src/component.js
```diff
@@ -19,27 +19,32 @@ export default {
...
     suit(...args) {
       return suit(this.widgetName, ...args);
     },
-    updateData(state = {}) {
-      this.state = state;
+    updateState(state = {}, isFirstRender) {
+      if (!isFirstRender) {
+        // Avoid to update the state on first render
+        // otherwise we have a flash from empty state
+        // to the next state
```
Haroenv: Avoid updating the state on first render, otherwise there will be a flash of placeholder data

src/component.js
```diff
@@ -19,27 +19,32 @@ export default {
...
   beforeDestroy() {
     this.instantSearchInstance.removeWidget(this.widget);
   },
   watch: {
-    widgetParams(newVal) {
-      const oldWidget = this.widget;
-      this.widget = this.widgetFactory(newVal);
+    widgetParams(nextWidgetParams) {
+      this.state = null;
+      this.instantSearchInstance.removeWidget(this.widget);
```
Haroenv: I wonder if this will cause a flash of double applied search parameters?

Haroenv: if that was the case, in the previous implementation, there would be a flash of empty parameters

samouss: The behaviour is the same as before, you can test on a slow network to see what happens.

PR #502, opened 2018-08-30
fix(Index): missing parameters / props

bobylito:
This PR updates the Index component so that it supports the same set of options as instantSearch(). It also removes the deprecated apiKey / appId parameters.

I don't know what would be a good strategy for testing the Index. Happy to pair with someone to figure out what would be a good approach.

### DIFF
--- a/docs/src/components/index.md
+++ b/docs/src/components/index.md
@@ -49,8 +49,8 @@ export default {
 |------------------|---------|---------|----------------------------------------------------------------------------------------------------------------------------------------------------|
 | searchClient      | Object  | ``      | The instance of a search client like algoliaSearch   |
 | index-name       | String  | ``      | The index name                                                                                                                                     |
-| query            | String  | ``      | The search query                                                                                                                                   |
-| query-parameters | Object  | ``      | The search query parameters. Available options are [documented here](https://www.algolia.com/doc/api-reference/search-api-parameters/). |
+| searchParameters | Object  | ``      | The search query parameters. Available options are [documented here](https://www.algolia.com/doc/api-reference/search-api-parameters/). |
+| searchFunction | Object  | ``      | A hook that will be called each time a search needs to be done, with the helper as a parameter. It’s your responsibility to call helper.search(). This option allows you to avoid doing searches at page load for example. |
 | stalledSearchDelay | number | `200`  | Time before the search is considered unresponsive. Used to display a loading indicator. |
 
 ## Slots
--- a/docs/src/components/index.md
+++ b/docs/src/components/index.md
@@ -18,41 +18,39 @@ This component automatically provides the search state to all its children.
 Basic usage:
 
 ```html
-<ais-index index-name="your_indexName"
-           app-id="YourAppID"
-           api-key="YourSearchAPIKey"
->
-  <!-- Add your InstantSearch components here. -->
-</ais-index>
+<template>
+  <ais-index index-name="your_indexName"
+            :searchClient="searchClient"
+  >
+    <!-- Add your InstantSearch components here. -->
+  </ais-index>
+</template>
+
+<!-- You need to instantiate the search client in your script -->
+<script>
+import algoliasearch from 'algoliasearch/lite';
+
+export default {
+  data() {
+    return {
+      searchClient: algoliasearch(
+        'latency',
+        '3d9875e51fbd20c7754e65422f7ce5e1'
+      ),
+    };
+  },
+};
+</script>
 ```
-
-Provide search query parameters:
-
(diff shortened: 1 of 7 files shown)


docs/src/components/index.md
```diff
@@ -18,41 +18,39 @@ This component automatically provides the search state to all its children.
...
+  },
+};
+</script>
 ```
 
-
 ## Props
 
-| Name             | Type    | Default | Description                                                                                                                                        |
+  | Name             | Type    | Default | Description                                                                                                                                        |
```
Haroenv: could this table be reformatted to have the same info and formatting as the other ones?

docs/src/components/index.md
```diff
@@ -18,41 +18,39 @@ This component automatically provides the search state to all its children.
...
-| query-parameters | Object  | ``      | The search query parameters. Available options are [documented here](https://www.algolia.com/doc/api-reference/search-api-parameters/). |
-| cache            | Boolean | `true`  | Whether to cache results or not. See [the documentation](https://www.algolia.com/doc/tutorials/getting-started/quick-start-with-the-api-client/javascript/#cache)       |
-| auto-search      | Boolean | `true`  | Whether to initiate a query to Algolia when this component is mounted                                                                               |
+| searchParameters | Object  | ``      | The search query parameters. Available options are [documented here](https://www.algolia.com/doc/api-reference/search-api-parameters/). |
+| searchFunction | Object  | ``      | A hook that will be called each time a search needs to be done, with the helper as a parameter. It’s your responsibility to call helper.search(). This option allows you to avoid doing searches at page load for example. |
```
Haroenv: it's still missing `routing` right?

src/components/Index.vue
```diff
@@ -12,24 +12,20 @@ import suit from '../suit';
...
+      searchFunction: this.searchFunction,
+      searchParameters: this.searchParameters,
     });
 
     return {
       instantSearchInstance: this.instantSearchInstance,
     };
   },
   props: {
-    apiKey: {
```
Haroenv: this change will need to be reflected in the storybooks as well, which used the api key form

src/components/Index.vue
```diff
@@ -12,24 +12,20 @@ import suit from '../suit';
 export default {
   provide() {
     this.instantSearchInstance = instantsearch({
-      appId: this.appId,
-      apiKey: this.apiKey,
```
Haroenv: searchClient is missing

Haroenv:
So actually using `searchClient` is what's causing `helper` to be null, which then causes an error in `removeWidget` (line 375 of `InstantSearch.js`). The code actually completely works and it searches etc. it's just that things are not removed (maybe only Configure.vue, but I don't see how it's different than the other ones)

src/component.js
```diff
@@ -34,6 +34,8 @@ export default {
     widgetParams: {
       handler(nextWidgetParams) {
         this.state = null;
+        // TODO: this line here breaks for unmounting a Configure :(
+        // only when `searchClient` is used; not if `apiKey+appId`
```
Haroenv: the offending line

bobylito:
Thanks for the analysis @Haroenv Working on it!

src/component.js
bobylito: Do you have a test to reproduce the error? Or a sample?

Haroenv: It happens in the storybook, when configure unmounts

bobylito:
I'm 100% sure why it was happening but here is what I noticed: 
 - helper is `null`
 - the only moment this happens is when InstantSearch is instanciated
 - when I set a break point in the `provide` it does not trigger the issue
 - the bug is erratic. With a conditional breakpoint set on the line that was failing, it was very random and not only limited to configure.

Feels like a race-condition but the fix might be bad practice. Happy to have your thoughts on it @Haroenv https://github.com/algolia/vue-instantsearch/pull/502/commits/a836e487527bad7b9b2e03c580814b6ada155a91

Haroenv:
I tried That earlier And then I was still missing something, seems like you’re on to something. I think in a situation where something unmounts before it’s fully mounted we need a check like that

bobylito:
That's a great insight @Haroenv. Thanks. I've made some changes accordingly.

I'm still not sure how it is related to `searchClient` though 😅

src/components/Index.vue
bobylito: You've done that right? in `utils.js`?

Haroenv:
Yep, I wasn't sure why this was happening because of this diff, but I guess something is different in the life cycle of InstantSearch when search client is generated or passed

PR #504, opened 2018-09-06
feat: allow search box to be controlled -> multi query!

Haroenv:
We allow the search box to be controlled via v-model. If you do that, you can sync between multiple search boxes across multiple indices

There's still an issue that this does 2n requests (not sure why, but at least it works). This would be caught by the newer request cache (not enabled by default)

### DIFF
--- a/src/components/SearchBox.vue
+++ b/src/components/SearchBox.vue
@@ -89,13 +89,13 @@ export default {
           this.$emit('input', this.value);
           this.state.refine(this.value);
         }
-        return this.value || this.localValue;
+        return this.value || this.localValue || this.state.query || '';
       },
       set(val) {
         this.localValue = val;
+        this.state.refine(val);
         if (this.isControlled) {
           this.$emit('input', val);
-          this.state.refine(val);
         }
       },
     },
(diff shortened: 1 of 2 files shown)


Haroenv:
https://deploy-preview-504--vue-instantsearch.netlify.com/stories/?selectedKind=MultiIndex&selectedStory=simple%20usage&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs is the link I think (it's not live yet)

Haroenv:
I won't generalise the solution for two reasons

1. we don't have a solid use case for shared widgets, most use cases focus on one searchbox - multiple hits
2. it add a second query being done. This is avoidable with `_useRequestCache`, but I don't think it's a good idea to release a feature that relies on that feature being true if it's not the default implementation so far.

I should probably add a test for this yes. Gotta figure out what a relevant test would be.

I think it _might_ be possible to make this inside the Component mixin, but I feel like it might add more overhead than worth it now, but I'm willing to reconsider that decision later.

Haroenv:
I've tried a few test strategies, but they all don't work, because vue-test-utils doesn't allow proper testing of v-model on something which isn't built in

PR #505, opened 2018-09-08
feat(conditional-rendering): add widget `ais-search-state`

Haroenv:
I'm still looking for a better name for this widget. Potential names include:

* `ais-search-state`
* `ais-state`
* `ais-results`
* `ais-result-state`
* `ais-state-results`

- [x] tests
- [x] docs
- [x] naming: `ais-search-state`


Haroenv:
It would be nice to get search state (uiState) like we do in React InstantSearch & routing, but this is a good temporary source of data IMO

PR #506, opened 2018-09-09
docs(https): fix https mixed content

vvo:
I get a lot of emails from Netlify about this:

image

Worth checking other *-instantsearch websites for this:
image


Haroenv:
correct, fixed this in v2 already

PR #510, opened 2018-09-11
feat(index): allow dynamically changing some of the props

Haroenv:
Left out: searchFunction & routing. These don't have an API for modifying in InstantSearch, since they get transformed in the constructor. 

The fact that they are left out is a reasonable, since there's few use cases for that. I made it throw an error that suggest opening an issue when someone wants to change it after the fact.


src/components/__tests__/Index.js
```diff
@@ -99,3 +114,80 @@ it('renders correctly (with slot used)', () => {
...
+      searchClient: {},
+      indexName: 'bla',
+    },
+  });
+
+  wrapper.setProps({
+    indexName: 'doggie_bowl',
+  });
+
+  expect(instantsearch.__helper.setIndex).toHaveBeenCalledTimes(1);
```
francoischalifour: Wouldn't it make more sense to expect `instantsearch.__helper.getIndex()` to be `doggie_bowl` instead of the two `setIndex` lines that are more testing the helper implementation instead? ([see reference](https://community.algolia.com/algoliasearch-helper-js/reference.html#AlgoliaSearchHelper#getIndex))

src/components/__tests__/Index.js
```diff
@@ -99,3 +114,80 @@ it('renders correctly (with slot used)', () => {
...
+    },
+  });
+
+  const newClient = { cats: 'rule', dogs: 'drool' };
+
+  wrapper.setProps({
+    searchClient: newClient,
+  });
+
+  expect(instantsearch.__helper.setClient).toHaveBeenCalledTimes(1);
```
francoischalifour: Same as above with [`helper.getClient()`](https://community.algolia.com/algoliasearch-helper-js/reference.html#AlgoliaSearchHelper#getClient)?

src/components/__tests__/Index.js
```diff
@@ -99,3 +114,80 @@ it('renders correctly (with slot used)', () => {
...
+  });
+
+  expect(global.console.error.mock.calls[0][0]).toMatchInlineSnapshot(`
+[Error: searchFunction configuration can not be changed dynamically at this point.
+
+Please open a new issue: https://github.com/algolia/vue-instantsearch/issues/new?template=feature.md]
+`);
+});
+
+it('Does not allows a change in `routing`', () => {
```
francoischalifour: allows → allow

src/components/__tests__/Index.js
```diff
@@ -99,3 +114,80 @@ it('renders correctly (with slot used)', () => {
...
+  wrapper.setProps({
+    searchClient: newClient,
+  });
+
+  expect(instantsearch.__helper.setClient).toHaveBeenCalledTimes(1);
+  expect(instantsearch.__helper.setClient).toHaveBeenCalledWith(newClient);
+  expect(instantsearch.__helper.search).toHaveBeenCalledTimes(1);
+});
+
+it('Does not allows a change in `search-function`', () => {
```
francoischalifour: allows → allow

src/components/Index.vue
```diff
@@ -50,6 +50,32 @@ export default {
...
+      this.instantSearchInstance.helper.setClient(searchClient).search();
+    },
+    indexName(indexName) {
+      this.instantSearchInstance.helper.setIndex(indexName).search();
+    },
+    stalledSearchDelay(stalledSearchDelay) {
+      // private InstantSearch.js API:
+      this.instantSearchInstance._stalledSearchDelay = stalledSearchDelay;
+    },
+    routing() {
```
francoischalifour: I think this is good 👍

src/components/__tests__/Index.js
```diff
@@ -99,3 +114,80 @@ it('renders correctly (with slot used)', () => {
...
+    },
+  });
+
+  wrapper.setProps({
+    indexName: 'doggie_bowl',
+  });
+
+  expect(instantsearch.__helper.setIndex).toHaveBeenCalledTimes(1);
+  expect(instantsearch.__helper.setIndex).toHaveBeenCalledWith('doggie_bowl');
+  expect(instantsearch.__helper.search).toHaveBeenCalledTimes(1);
```
francoischalifour: Maybe assert that the results are the one you expected instead of relying on `helper.search` to be called?

src/components/__tests__/Index.js
Haroenv: yeah sure, that's possible :)

Haroenv:
Those changes require the mock to be on the helper instead. I think it's possible, but quite a bit more work to reimplement the helper correctly. WDYT?

PR #513, opened 2018-09-12
fix(vue-autosuggest) fix vue-autosuggest story

darrenjennings:
* change suggestions to be array of object with format { data: [] }
* use slots


Haroenv:
thanks Darren ❤️, I missed that part about `[{data: []}]` in suggestions

PR #516, opened 2018-09-14
feat(geoSearch): add widget

Haroenv:
TODO: 

- [ ] main implementation
- [ ] finalise docs
- [ ] tests

### DIFF
--- a/.storybook/styles.css
+++ b/.storybook/styles.css
@@ -98,3 +98,8 @@ body {
 .playground-hits-desc > p:last-child {
   margin: 0;
 }
+
+#map {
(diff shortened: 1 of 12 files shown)


samouss:
The API of the connector will probably change to match the React implementation (drop the support for the search parameters they can be set with `Configure` amongst other things). Maybe it worth postpone the implementation?

Haroenv:
Yes, if someone is interested in the implementation of a geo search widget, upon seeing this closed PR, you can take a look at the initial work done here, and ping us in a new issue if you need that help. 

Have a nice day 👋

PR #520, opened 2018-09-19
feat: allow users to override classes with `class-names` prop

### LINKED ISSUE #517: v2 - Overriding class names not working
**Feedback on the v2 alpha version 🆕**

When you pass the prop `class-names` to  the components, as described in here [https://deploy-preview-416--vue-instantsearch.netlify.com/getting-started/styling.html#overriding-class-names](url)
 the classes are not beeing overriding, instead it add the attribute as `class-names="[object Object]" `

### Reproduction
https://codesandbox.io/s/l7jp6rxr3m
Inspect the element on you will see the attribute `class-names="[object Object]" `

v2.0.0-alpha.1

Haroenv:
3 separate things happening here:

1. refactor to put all classname logic in suit mixin
2. add this new prop to all docs
3. update styling doc

Also some smaller things I saw I've updated inline (two new stories too)

closes #517


src/mixins/suit.js
```diff
@@ -1,9 +1,19 @@
...
+    classNames: {
+      type: Object,
+      default: undefined,
+    },
+  },
   methods: {
     suit(...args) {
-      return suit(this.widgetName, ...args);
+      const className = suit(name, ...args);
+      if (this.classNames && this.classNames[className]) {
```
samouss: We should avoid to rely on the generated class name for the keys of the object. We should provide to the users stable keys. This way we don't break their if we choose to update our class. This is a approach really similar of what we have inside InstantSearch (see below).

https://github.com/algolia/instantsearch.js/blob/ef731cc48c7640aac71e09db799b476ef2474f3b/src/widgets/refinement-list/refinement-list.js#L271-L282

stories/Hits.stories.js
```diff
@@ -57,4 +57,18 @@ storiesOf('Hits', module)
         <template slot="footer">Footer</template>
       </ais-panel>
     `,
+  }))
+  .add('inline template', () => ({
```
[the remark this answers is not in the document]
Haroenv: I wrote it to test out if it still worked as expected, but doesn't need to be committed, indeed!

src/mixins/suit.js
Haroenv:
```diff
diff --git a/src/mixins/suit.js b/src/mixins/suit.js
index af5f283..fc51857 100644
--- a/src/mixins/suit.js
+++ b/src/mixins/suit.js
@@ -8,12 +8,12 @@ export const createSuitMixin = ({ name }) => ({
     },
   },
   methods: {
-    suit(...args) {
-      const className = suit(name, ...args);
-      if (this.classNames && this.classNames[className]) {
+    suit(element, modifier) {
+      const identifier = [element, modifier].filter(Boolean).join('-');
+      if (this.classNames && this.classNames[identifier]) {
         return this.classNames[className];
       }
-      return className;
+      return suit(name, element, modifier);
     },
   },
 });
```

Does this make sense, or rather in the components themselves? I think I prefer it here, since if there's only one place it's written, there's less chance of a typo happening. 

If a typo happens in the class name, it's as much of a breaking change as in the option, and we could add custom logic here to keep it non-breaking if necessary IMO

samouss: Use the full class then!

PR #522, opened 2018-09-20
docs(examples): build all on netlify

Haroenv:
1. make links relative in the example
2. build on netlify


Haroenv:
actually now that I think of this, it should of course be done in codesandbox 🤦‍♂️

PR #524, opened 2018-09-20
refactor(mixin): make into a factory function

Haroenv:
This has the advantage of making it easier to make a custom component, since more speaks out of itself

### DIFF
--- a/docs/src/getting-started/migration.md
+++ b/docs/src/getting-started/migration.md
@@ -26,6 +26,8 @@ Some components have been renamed to be be more consistent with other InstantSea
 * `ais-rating` -> `ais-rating-menu`
 * `ais-sort-by-selector` -> `ais-sort-by`
 
+The `Component` mixin has been renamed to `createWidgetMixin({ connector })`. Read more about that in the [custom component guide](/advanced/custom-components.html).
+
 ### New components
 
 Find more information on these in their respective documentation.


src/components/Autocomplete.vue
```diff
@@ -21,11 +21,11 @@
 </template>
 
 <script>
-import algoliaComponent from '../mixins/component';
+import { createAlgoliaMixin } from '../mixins/component';
```
samouss: We can find another name that don't mention Algolia like `createWidgetMixon` or something else? We can also rename the file with the name found.

src/components/PoweredBy.vue
```diff
@@ -31,10 +31,10 @@
 </template>
 
 <script>
-import algoliaComponent from '../mixins/component';
+import { createAlgoliaMixin } from '../mixins/component';
 
 export default {
-  mixins: [algoliaComponent],
+  mixins: [createAlgoliaMixin()],
```
samouss: Once we merge the `suit` mixin we can drop this one.

src/components/SearchInput.vue
```diff
@@ -114,10 +114,10 @@
 </template>
 
 <script>
-import algoliaComponent from '../mixins/component';
+import { createAlgoliaMixin } from '../mixins/component';
 
 export default {
-  mixins: [algoliaComponent],
+  mixins: [createAlgoliaMixin()],
```
samouss: Once we merge the `suit` mixin we can drop this one.

src/mixins/component.js
```diff
@@ -1,6 +1,6 @@
 import suit from '../util/suit';
 
-export default {
+export const createAlgoliaMixin = ({ connector } = {}) => ({
```
samouss: Once we have the dedicated `suit` mixin we can maybe throw an error when there is no connector. The mixin makes no sense without it.

Haroenv: Actually I just thought about that, and it allows you to make a fully custom widget, using the helper like this (pushed an example)

samouss: It's a bit risky though because you have to manage what should be removed etc.. when it's unmounted. But sill it's a great capability.

samouss:
The CI is failing though.

Haroenv:
Yep, I had a followup commit :)

PR #525, opened 2018-09-20
docs(range-input): add two slider examples

Haroenv:
adds two examples: vuetify and vue-slider-component

Both aren't really that great, but they definitely work. For both I had to patch the currentRefinement not to be -Infinity and Infinity. Maybe that's something that should be done in the connector or render method instead?


samouss:
Yes it's worth moving it in the connector. We didn't do it yet mainly for backward compatibility reason IIRC.

PR #533, opened 2018-09-24
feat(Configure): don't add attributes to the DOM

Haroenv:
see https://vuejs.org/v2/api/#inheritAttrs for more info. This version is from July 2017, so more than old enough for us

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "lodash": "^4.17.11"
   },
   "peerDependencies": {
-    "vue": "^2.2.2"
+    "vue": "^2.4.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -48,7 +48,7 @@
     "lodash": "^4.17.11"
   },
   "peerDependencies": {
-    "vue": "^2.2.2"
+    "vue": "^2.4.0"
```
francoischalifour: Do you have an idea of how many users it would impact?

Haroenv: likely none, it's a quite old version and this is a major change anyway :)

Haroenv: (also we already required it via another thing earlier 😉, it just wasn't changed yet)

PR #529, opened 2018-09-21
refactor: rename things

Haroenv:
1. rename `ais-index` to `ais-instant-search`
2. rename exported components to have `Ais` prefix (so that they have the same name if you import them, as well as avoiding a conflict for InstantSearch)
3. rename stories to those names too

These changes are put in the migration guide

PR #534, opened 2018-09-24
feat(theme): implement demo

samouss:
localhost_8080_


examples/default-theme/src/App.vue
```diff
@@ -0,0 +1,134 @@
...
+            >
+              <div class="product-desc-wrapper">
+                <div class="product-name">
+                  <ais-highlight attribute="name" :hit="item" />
+                </div>
+                <div class="product-type">
+                  <ais-highlight attribute="type" :hit="item" />
+                </div>
+                <div class="product-footer">
+                  <div class="ais-RatingMenu-link">
```
Haroenv: I wonder if we should expose this star rating thing somehow? probably out of scope for now though!

examples/default-theme/src/App.vue
```diff
@@ -0,0 +1,134 @@
+<template>
+  <div id="root">
+    <ais-index
```
Haroenv: needs to be renamed after #529, so I'll want to merge that one first :)

PR #537, opened 2018-10-01
feat(InstantSearch): call addAlgoliaAgent if possible

Haroenv:
~~still to do: read it from the package.json etc. I have to figure out how it definitely doesn't go fully in the source code. I have a babel macro written for it, but we use buble and not babel here~~ 

EDIT: package.json import will leave only the `version` in the bundle via rollup, so that's good to go since we increment version before releasing


src/components/InstantSearch.vue
```diff
@@ -103,6 +104,9 @@ export default {
       );
     },
   },
+  created() {
+    addAlgoliaAgent(this.searchClient);
```
samouss: Since the logic is short & simple, I would rather go for inline it. We are not gonna re-use this logic anywhere in the code.

src/components/InstantSearch.vue
```diff
@@ -9,7 +9,7 @@
 import instantsearch from 'instantsearch.js/es/';
 import { createSuitMixin } from '../mixins/suit';
 import { warn } from '../util/warn';
-import { addAlgoliaAgent } from '../util/client';
+import { version } from '../../package.json'; // rollup does pick only what needed from json
```
samouss: It would be better to not rely on JSON but rather relying on a plain JS file. We can drop the dependency on the JSON Rollup plugin then. But I'm fine if it's stays like this though.

Haroenv: I wanted to avoid adding another build step in the release process. IMO if it would show any problems at some point, we can still switch at that point.

src/components/InstantSearch.vue
```diff
@@ -103,6 +104,11 @@ export default {
       );
     },
   },
+  created() {
+    if (this.searchClient.addAlgoliaAgent) {
```
francoischalifour: We should go for `typeof this.searchClient.addAlgoliaAgent === 'function'` to ensure that we can actually call it in the line below ([see on InstantSearch.js](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/InstantSearch.js#L111)).

Haroenv: I can't imagine a use case where someone adds that key without it being a function. Was aware of this tradeoff but will change it, since that's what we do in InstantSearch.js

PR #538, opened 2018-10-01
chore(release): allow prerelease

Haroenv:
The flow didn't change, except that now you can give a prerelease-style version and it will release on the right npm tag

### DIFF
--- a/scripts/get-npm-tag.js
+++ b/scripts/get-npm-tag.js
@@ -1,0 +1,12 @@
+#!/usr/bin/env node
+/* eslint-disable import/no-commonjs */
+/* eslint-disable no-console */
+const { version } = require('../package.json');
+
+const getReleaseTag = _version => {
+  const [match = 'latest'] = _version.match(/[a-z]+/) || [];
+  return match;
+};
+
+// give this on to the main script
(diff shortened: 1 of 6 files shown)


Haroenv:
I wrote it in JS because messing with regexes and stuff is really meh in bash, and doing scripts in JS is also quite meh

scripts/get-npm-tag.js
```diff
@@ -0,0 +1,12 @@
...
+/* eslint-disable no-console */
+const { version } = require('../package.json');
+
+const getReleaseTag = _version => {
+  const [match = 'latest'] = _version.match(/[a-z]+/) || [];
+  return match;
+};
+
+// give this on to the main script
+console.log(getReleaseTag(version));
```
samouss: There is maybe a more "cleaner" way to output a value?

Haroenv: alternative is `process.stdout.write`. I'll change to that, only difference is that console.log will send a newline :

samouss: I'm fine with this implementation then, `console.log` actually calls `process.stdout.write` internally.

Issue #544, opened 2018-10-04
v2 - advanced routing & vue router


Haroenv:
**Feedback on the v2 alpha version 🆕**

### Describe the bug / features

Currently there's three existing ways how to use the InstantSearch routing. 

The first option is putting `:routing="true"`. This will use the default serialising that doesn't lose any information, but might be a bit verbose

The second option is to put an object of config. This object can take `stateMapping`, with the functions `stateToRoute` and `routeToState` to serialise differently, but still use the default routing. This allows to rename things to make them easier to read, without touching how the serialising itself happens.

Finally, you can also change the URL to use full URLs, rather than just the query string. You need to change the `router` key inside the `routing` object. You can import the default (history) router from `import {history} from 'instantsearch.js/es/lib/routers'`, and modify, like in InstantSearch JS.

[![Edit vue-instantsearch-app](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/q8pmz6n7lj?module=%2Fsrc%2FApp.vue)

All docs for InstantSearch routing configuration are [here](https://community.algolia.com/instantsearch.js/v2/guides/routing.html).

Finally an option is to use Vue Router. All previous examples will _work_ using Vue Router, as long as they don't conflict, as long as you don't try to do specific Vue Router things which are controlled by InstantSearch, since that might cause infinite loops, and InstantSearch provided query strings won't be available to Vue Router, as soon as the routing changes from its initial deserialisation.

#### How to make the InstantSearch router use Vue Router? 

This is possible by passing a custom router to to InstantSearch. The router prop accepts an object looking like this: 

```js
const instantSearchRouting = {
  router: {
    read() {
      return router.currentRoute.query;
    },
    write(routeState) {
      router.push({
        query: routeState,
      });
    },
    createURL(routeState) {
      return router.resolve({
        query: routeState
      }).href;
    },
    onUpdate() {}, 
    dispose() {},
  },
}
```

The problem is that this doesn't work with Objects (https://github.com/vuejs/vue-router/issues/1268). The solution can be in two ways. 

1. use `stateMapping` to make the `routeState` fully flat, or only using arrays. These arrays can be nested two times, but not more
2. modify Vue Router to work with objects (https://github.com/vuejs/vue-router/issues/1259#issuecomment-306000548)

```js
import qs from 'qs';

const router = new Router({
    routes: [
        // ...
    ],
    // set custom query resolver
    parseQuery(query) {
[...]

PR #548, opened 2018-10-05
docs(router): how to use it

Haroenv:
note i don't use vue router all the time in the guide, but i think that's fine in _most_ use cases

see also #544 for seeing if we need to ship this config in the library or not. So far I haven't found a sufficiently good way of including it, so left it in user land, mostly because of the "flat" or "nested arrays" problem.

### DIFF
--- a/docs/src/advanced/vue-router-url-sync.md
+++ b/docs/src/advanced/vue-router-url-sync.md
@@ -8,126 +8,67 @@ navWeight: 3
 editable: true
 githubSource: docs/src/advanced/vue-router-url-sync.md
 ---
-> NOTE: this guide has not yet been updated for v2
+> NOTE: this guide **has** been updated for v2
 
-In this guide you'll learn how to synchronize your InstantSearch experience with the [official Vue-Router](https://github.com/vuejs/vue-router).
+Currently there's three existing ways how to use InstantSearch routing.
 
-There are two things you'll want to setup in order to have the Vue router play nicely with InstantSearch.
+The first option is putting `:routing="true"` on `ais-instant-search`. This will use the default serialising that doesn't lose any information, but might be a bit verbose
 
-1. Extract some parameters from the router and initialize the search store with it. That way, on the initial load of the search experience, it will display results based on the current route.
+The second option is to put an object of configuration. This object can take `stateMapping`, with the functions `stateToRoute` and `routeToState` to serialise differently, but still use the default routing. This allows to rename things to make them easier to read, without touching how the serialising itself happens.
 
-2. Push the new state of the search to the router as the user interacts with the InstantSearch components.
+Finally, you can also change the URL to use full URLs, rather than just the query string. You need to change the `router` key inside the `routing` object. You can import the default (history) router from `import {history} from 'instantsearch.js/es/lib/routers'`, and modify, like in InstantSearch JS.
 
-## Example app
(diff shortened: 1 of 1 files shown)


docs/src/advanced/vue-router-url-sync.md
```diff
@@ -8,126 +8,67 @@ navWeight: 3
...
 
-## Example app
+[![Edit vue-instantsearch-app](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/q8pmz6n7lj?module=%2Fsrc%2FApp.vue)
 
-A fully working example of synchronizing the Vue router with Vue InstantSearch can be found here: https://github.com/algolia/vue-instantsearch-examples/tree/master/examples/vue-router
+All docs for InstantSearch routing configuration are [here](https://community.algolia.com/instantsearch.js/v2/guides/routing.html).
 
-## Initialize InstantSearch with route parameters
+Finally an option is to use Vue Router. All previous examples will _work_ using Vue Router, as long as they don't conflict, as long as you don't try to do specific Vue Router things which are controlled by InstantSearch, since that might cause infinite loops, and InstantSearch provided query strings won't be available to Vue Router, as soon as the routing changes from its initial deserialization.
```
samouss: Which kind of setup could trigger an infinite loop? Is it already happened?

docs/src/advanced/vue-router-url-sync.md
```diff
@@ -8,126 +8,67 @@ navWeight: 3
...
-```html
-<template>
-  <ais-instant-search :search-store="searchStore" :query="query">
-    <!-- ... -->
-  </ais-instant-search>
-</template>
-```
-
-The search store being bound to the `data` option, we can now observe the changes to reflect them in the router.
+Note that here I only use the `query` key, but other keys can also be used. It's advised here to use a `stateMapping` that changes nested objects into flat objects, or as a second option to modify Vue Router to allow nested objects in the query string:
```
samouss: We should explain why we have to use flat objects rather than nested one. WDYT?

docs/src/advanced/vue-router-url-sync.md
Haroenv: It was actually with the wrong config, I’ll remove it since it’s not really likely

docs/src/advanced/vue-router-url-sync.md
```diff
@@ -50,7 +50,9 @@ const instantSearchRouting = {
 }
 ```
 
-Note that here I only use the `query` key, but other keys can also be used. It's advised here to use a `stateMapping` that changes nested objects into flat objects, or as a second option to modify Vue Router to allow nested objects in the query string:
+Note that here I only use the `query` key, but other keys can also be used. It's advised here to use a `stateMapping` that changes nested objects into flat objects. The reason why you need flat objects, is because by default Vue Router will serialize an object as value for a query string object as `[object Object]`. This can be avoided by either _not_ having deep objects (possibly replaced by arrays, or a flat version with only what you need in your app), or by modifying the Vue Router
```
samouss: the Vue Router [PERIOD]

docs/src/advanced/vue-router-url-sync.md
```diff
@@ -50,7 +50,9 @@ const instantSearchRouting = {
...
+
+That second option is to modify Vue Router to allow nested objects in the query string:
```
samouss: Since we modify the previous sentence this one seems a bit redundant.

PR #553, opened 2018-10-19
feat(SearchBox): forward class-names & slots

Haroenv:
also rename clearTitle -> resetTitle


src/components/__tests__/SearchBox.js
```diff
@@ -166,3 +166,63 @@ test('refine on empty string on form reset', () => {
...
+
+<button type="submit"
+        title="Search"
+        hidden="hidden"
+        class="ais-SearchBox-submit"
+>
+  <span>
+    SUBMIT
+  </span>
+</button>
```
samouss: For those snapshots we can use a dedicated file.

PR #556, opened 2018-10-24
fix(example): add codesandbox config

Haroenv:
see https://github.com/CompuIves/codesandbox-client/issues/1179#issuecomment-432305547

working urls: 

* https://codesandbox.io/s/github/algolia/vue-instantsearch/tree/fix/codesandbox-examples/examples/default-theme
* https://codesandbox.io/s/github/algolia/vue-instantsearch/tree/fix/codesandbox-examples/examples/ecommerce
https://codesandbox.io/s/github/algolia/vue-instantsearch/tree/fix/codesandbox-examples/examples/media

### DIFF
--- a/examples/default-theme/sandbox.config.json
+++ b/examples/default-theme/sandbox.config.json
@@ -1,0 +1,3 @@
+{
+  "template": "nuxt"
+}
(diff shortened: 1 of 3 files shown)


samouss:
It means that we can drop the script that publish the examples on Netlify right?

PR #557, opened 2018-10-30
feat(CurrentRefinements): give createURL to slot

Haroenv:
The function is prefilled this with the item.

I'm doubting if maybe it should be prefilled with the url in the `item` itself instead? This is a larger change and we can do it another time though


Haroenv:
Will do

PR #560, opened 2018-11-01
feat(RangeInput): refine & currentRefinement sync

Haroenv:
1. currentRefinements -> currentRefinement
2. currentRefinement: object
3. refine: object

The tests aren't changed; since the internal refine is tested, not the exposed one

IFW-196

### DIFF
--- a/docs/src/components/RangeInput.md
+++ b/docs/src/components/RangeInput.md
@@ -78,7 +78,7 @@ Note that you can pass the prop `class-names`, with an object of class names and
 
 Name | Scope | Description
 ---|---|---
-default | `{ refine: ({min, max}) => void, currentRefinement: {min: number, max: number}, noRefinements, range: {min: number, max: number}}` | Slot to override the DOM output. `refine` takes the updated values and then trigger the search. The `currentRefinements` contains the min and the max as set in the search state (and should be used to render the input values).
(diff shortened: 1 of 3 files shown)


docs/src/components/RangeInput.md
```diff
@@ -78,7 +78,7 @@ Note that you can pass the prop `class-names`, with an object of class names and
 
 Name | Scope | Description
 ---|---|---
-default | `{ refine: (min, max) => void, currentRefinements: [number, number], noRefinements, range: {min: number, max: number}}` | Slot to override the DOM output. `refine` takes the updated values and then trigger the search. The `currentRefinements` contains the min and the max as set in the search state (and should be used to render the input values).
+default | `{ refine: ({min, max}) => void, currentRefinement: {min: number, max: number}, noRefinements, range: {min: number, max: number}}` | Slot to override the DOM output. `refine` takes the updated values and then trigger the search. The `currentRefinements` contains the min and the max as set in the search state (and should be used to render the input values).
```
samouss:
```suggestion
default | `{ refine: ({ min, max }) => void, currentRefinement: { min: number, max: number }, noRefinements, range: { min: number, max: number }}` | Slot to override the DOM output. `refine` takes the updated values and then trigger the search. The `currentRefinements` contains the min and the max as set in the search state (and should be used to render the input values).
```

samouss:
You can still add a test for the slot that will assert that the exposed API is the expected one (this is what we do in the stories, agree but better safe than sorry).

PR #562, opened 2018-11-02
feat(router): force object form

Haroenv:
IFW-221

We force the object form, because that way there's less options, using it as a widget would indeed be nicer than as option, but we can do that later, I didn't completely grasp how to do that with InstantSearch.js, it seems to be setting some internal things, more than just a widget

### DIFF
--- a/docs/src/advanced/vue-router-url-sync.md
+++ b/docs/src/advanced/vue-router-url-sync.md
@@ -11,25 +11,46 @@ githubSource: docs/src/advanced/vue-router-url-sync.md
 
 > NOTE: this guide **has** been updated for v2
 
-Currently there's three existing ways how to use InstantSearch routing.
 
-The first option is putting `:routing="true"` on `ais-instant-search`. This will use the default serialising that doesn't lose any information, but might be a bit verbose
 
-The second option is to put an object of configuration. This object can take `stateMapping`, with the functions `stateToRoute` and `routeToState` to serialise differently, but still use the default routing. This allows to rename things to make them easier to read, without touching how the serialising itself happens.
+The `routing` prop on `ais-instant-search` accepts an object. The simplest way to get started without customising URLs at all is the following:
 
-Finally, you can also change the URL to use full URLs, rather than just the query string. You need to change the `router` key inside the `routing` object. You can import the default (history) router from `import {history} from 'instantsearch.js/es/lib/routers'`, and modify, like in InstantSearch JS.
+```vue
+<template>
+  <ais-instant-search
+    :routing="{
+      router: historyRouter(),
+      stateMapping: simpleMapping(),
+    }"
+  >
+    <!-- Your search components go in here -->
+  </ais-instant-search>
+</template>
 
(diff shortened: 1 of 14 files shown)


docs/src/advanced/vue-router-url-sync.md
```diff
@@ -11,25 +11,46 @@ githubSource: docs/src/advanced/vue-router-url-sync.md
...
+      historyRouter,
+      simpleMapping,
+    };
+  },
+};
+</script>
+```
 
-Finally an option is to use Vue Router. All previous examples will _work_ using Vue Router, as long as they don't conflict, as long as you don't try to do specific Vue Router things which are controlled by InstantSearch, since InstantSearch provided query strings won't be available to Vue Router, as soon as the routing changes from its initial deserialization.
+They're routing object contains two keys: "history" and "stateMapping". the history object is used for writing and reading to the URL, and the state mapping is used for mapping the InstantSearch state towards the state that will be read and written to the URL.
```
tkrugg: I would suggest splitting this in a bullet list
```suggestion
They're routing object contains two keys: `history` and `stateMapping`:

- **history**: used for writing and reading to the URL,
- **stateMapping**: used for mapping the InstantSearch state towards the state that will be read and written to the URL.
```

docs/src/advanced/vue-router-url-sync.md
```diff
@@ -11,25 +11,46 @@ githubSource: docs/src/advanced/vue-router-url-sync.md
...
 
-## How **do** I use Vue Router?
+If you want to customise which things are written in the URL but don't want to customise how exactly the URL looks you will use state mapping. The way to do this is replacing the call to `stateMapping` with an object with the functions `stateToRoute` and `routeToState`.
```
tkrugg: I think an example can help illustrate this

docs/src/components/InstantSearch.md
```diff
@@ -53,7 +53,7 @@ Name | Type | Default | Description | Required
 `index-name` | String  | | The index to target for the search | yes
 `searchFunction` | `(AlgoliaHelper) => void` | | A hook that will be called each time a search needs to be done, with the [helper](https://community.algolia.com/algoliasearch-helper-js/) as a parameter. It’s your responsibility to call `helper.search()`. This option allows you to avoid doing searches at page load for example. | no
 `stalled-search-delay` | Number | `200`  | Time (in ms) before the search is considered unresponsive. Used to display a loading indicator. | no
-`routing` | Boolean or Object | `false` | Enable the default routing feature by passing `true`. More advanced usage is documented [here](https://community.algolia.com/instantsearch.js/v2/guides/routing.html). | no
+`routing` | Object | | documented [here](/advanced/vue-router-url-sync.html). | no
```
tkrugg: it might be worth detailing the shape of Object? in this case `{ history, stateMapping }`

docs/src/advanced/vue-router-url-sync.md
Haroenv: There is an example lower on the page on how to override stateMapping

PR #565, opened 2018-11-05
docs(SearchClient): preventing (initial & empty) request

Haroenv:
This is the most correct way of preventing initial search, both in InstantSearch v1 and v2

fixes no specific open issue, but people have been asking this capability time and time again.

IFW-220

### DIFF
--- a/docs/src/advanced/conditional-requests.md
+++ b/docs/src/advanced/conditional-requests.md
@@ -69,7 +69,7 @@ Functionally this is not any different from directly passing the created search 
 
 ## Mocking a search request
 
-Since we don't want to do a request when the query is `""`, we first need to detect it:
+Since we don't want to do a request when the query is empty (`""`), we first need to detect it:
 
 ```js
 const searchClient = {
--- a/docs/src/advanced/conditional-requests.md
+++ b/docs/src/advanced/conditional-requests.md
@@ -84,7 +84,7 @@ const searchClient = {
 
 There can be multiple requests done in case a user for example clicks a hierarchical menu open. For now this doesn't really have a big impact, except that we have to make sure to check that every query is empty before we intercept.
 
-Then we need to return a [Response](https://www.algolia.com/doc/api-reference/api-methods/search/?language=javascript#response). This is an array of objects, of the same length as the `requests` array. At the bare minimum, each object needs to contain: `processingTimeMS`, `nbHits`, `hits` and `facets`:
+Then, we need to return a [formatted response](https://www.algolia.com/doc/api-reference/api-methods/search/?language=javascript#response). This is an array of objects of the same length as the `requests` array. At the bare minimum, each object needs to contain: `processingTimeMS`, `nbHits`, `hits` and `facets`:
 
 ```js
 const searchClient = {
--- a/docs/src/advanced/conditional-requests.md
+++ b/docs/src/advanced/conditional-requests.md
@@ -82,7 +82,7 @@ const searchClient = {
 }
 ```
 
-There can be multiple requests done in case a user for example clicks a hierarchical menu open. For now this doesn't really have a big impact, except that we have to make sure to check that every query is empty before we intercept.
+There can be multiple requests done in case a user for example clicks a hierarchical menu open. For now, this doesn't really have a big impact except that we have to make sure to check that every query is empty before we intercept the function call.
 
 Then, we need to return a [formatted response](https://www.algolia.com/doc/api-reference/api-methods/search/?language=javascript#response). This is an array of objects of the same length as the `requests` array. At the bare minimum, each object needs to contain: `processingTimeMS`, `nbHits`, `hits` and `facets`:
 
--- a/docs/src/advanced/conditional-requests.md
+++ b/docs/src/advanced/conditional-requests.md
@@ -1,0 +1,168 @@
+---
+title: Conditionally doing search requests
+mainTitle: Advanced
+layout: main.pug
+category: Advanced
+withHeadings: true
+navWeight: 7
+editable: true
+githubSource: docs/src/advanced/conditional-requests.md
+---
+
+> NOTE: this guide **has** been updated for v2
+
+InstantSearch will do a requests to Algolia servers on every key stroke, which means that it's fast and users will see updated results instantly. This also means that a request is being done to show the initial results on an empty query. While this usually isn't an issue and even a feature, since it warms up the network connection so subsequent requests will be done faster too.
+
+However, there are cases in which you do not want to do extra network requests than strictly necessary, for example when you are using a custom backend, or when the initial results are never displayed.
+
+## How it's made possible
+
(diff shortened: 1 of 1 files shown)


docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+category: Advanced
+withHeadings: true
+navWeight: 7
+editable: true
+githubSource: docs/src/advanced/conditional-requests.md
+---
+
+> NOTE: this guide **has** been updated for v2
+
+InstantSearch will do a requests to Algolia servers on every key stroke, which means that it's fast and users will see updated results instantly. This also means that a request is being done to show the initial results on an empty query. While this usually isn't an issue and even a feature, since it warms up the network connection so subsequent requests will be done faster too.
```
francoischalifour:
```suggestion
InstantSearch sends a request to Algolia servers on every keystroke, which means that it's fast and users will see updated results instantly. A request is also sent to show the initial results on an empty query. This is the original behavior since it warms up the network connection so subsequent requests will be done faster too.
```

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+
+However, there are cases in which you do not want to do extra network requests than strictly necessary, for example when you are using a custom backend, or when the initial results are never displayed.
+
+## How it's made possible
+
+InstantSearch is the UI part on top a search client with a state managed by a [JavaScript Helper](https://github.com/algolia/algoliasearch-helper-js). These three layers are composable and can be interchanged to leverage the InstantSearch widgets system with different search clients.
```
francoischalifour:
```suggestion
InstantSearch is the UI part on top of a search client with a state managed by a [JavaScript Helper](https://github.com/algolia/algoliasearch-helper-js). These three layers are composable and can be interchanged to leverage the InstantSearch widgets system with different search clients.
```

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+<script>
+import algoliasearch from "algoliasearch/lite";
+
+const algoliaClient = algoliasearch(
+  "latency",
+  "6be0576ff61c053d5f9a3225e2a90f76"
+);
+
+const searchClient = {
+  async search(requests) {
```
francoischalifour: Is `async` recommended with the common Vue tooling?

It is usually not easily compilable down to ES5.

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+    :search-client="searchClient"
+    index-name="instant_search"
+  >
+    <ais-search-box />
+    <ais-hits />
+  </ais-instant-search>
+</template>
+
+<script>
+import algoliasearch from "algoliasearch/lite";
```
francoischalifour: Did you format the file so it uses single quotes?

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+  data() {
+    return {
+      searchClient,
+    };
+  }
+};
+</script>
+```
+
+Functionally this is not any different from directly passing the created search client from the `algoliasearch` call, but what it empowers us is conditionally **not** calling the Algolia search request, and thus not doing the query.
```
francoischalifour:
```suggestion
Functionally, this is not any different from directly passing the created search client from the `algoliasearch` call, but what it empowers us is conditionally **not** sending the Algolia search request, and thus not performing the query.
```

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Mocking a search request
+
+Since we don't want to do a request when the query is `""`, we first need to detect it:
```
francoischalifour:
```suggestion
Since we don't want to do a request when the query is empty (`""`), we first need to detect it:
```

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+  async search(requests) {
+    if (requests.every(({ params: { query } }) => Boolean(query) === false)) {
+      // here we need to do something else
+    }
+    return algoliaClient.search(requests);
+  },
+}
+```
+
+There can be multiple requests done in case a user for example clicks a hierarchical menu open. For now this doesn't really have a big impact, except that we have to make sure to check that every query is empty before we intercept.
```
francoischalifour:
```suggestion
There can be multiple requests done in case a user for example clicks a hierarchical menu open. For now, this doesn't really have a big impact except that we have to make sure to check that every query is empty before we intercept the function call.
```

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+
+Then we need to return a [Response](https://www.algolia.com/doc/api-reference/api-methods/search/?language=javascript#response). This is an array of objects, of the same length as the `requests` array. At the bare minimum, each object needs to contain: `processingTimeMS`, `nbHits`, `hits` and `facets`:
```
francoischalifour:
```suggestion
Then, we need to return a [formatted response](https://www.algolia.com/doc/api-reference/api-methods/search/?language=javascript#response). This is an array of objects of the same length as the `requests` array. At the bare minimum, each object needs to contain: `processingTimeMS`, `nbHits`, `hits` and `facets`:
```

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+
+const algoliaClient = algoliasearch(
+  "latency",
+  "6be0576ff61c053d5f9a3225e2a90f76"
+);
+
+const searchClient = {
+  async search(requests) {
+    // eslint-disable-next-line no-console
+    console.log(
```
francoischalifour: This should be a comment IMO.

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
...
+      "change conditional if any of the other facets are faked",
+      requests
+    );
+    if (requests.every(({ params: { query } }) => Boolean(query) === false)) {
+      return {
+        results: requests.map(params => {
+          // eslint-disable-next-line no-console
+          console.log("fake something of the result if necessary", params);
```
francoischalifour: This should be a comment IMO.

docs/src/advanced/conditional-requests.md
```diff
@@ -0,0 +1,168 @@
+---
+title: Conditionally doing search requests
```
tkrugg: is "making" a better term? I'll trust your judgment.
```suggestion
title: Conditionally making search requests
```

docs/src/advanced/conditional-requests.md
tkrugg: search is already returning a promise? 
In that case async is unneeded anyway.

Haroenv: Regular vue tooling transpiles async indeed, reason I did is to avoid writing `Promise.resolve` in the interception code

PR #576, opened 2018-11-12
fix(RefinementList): allow overriding of searchable placeholder

### LINKED ISSUE #572: v2 - RefinementList - Add slot for searchable

Please add the ability to change the placeholder text

v2

Haroenv:
Also noticed that the test was broken for `searchable`, so fixed that one too.

New prop: `searchable-placeholder` on `ais-refinement-list`

fixes #572

### DIFF
--- a/docs/src/components/RefinementList.md
+++ b/docs/src/components/RefinementList.md
@@ -25,6 +25,7 @@ Name | Type | Default | Description | Required
 ---|---|---|---|---
 attribute | string | | The attribute to refine on click | yes
 searchable | boolean | `false` | You can also search within the options of this | no
+placeholder | string | `` | If searchable is `true`, this will be the placeholder for that search box | no
 operator | "or" / "and" | "or" | How to apply refinements | no
 limit | number | 10 | Number of items to show | no
 showMoreLimit | number | 20 | Number of items to show when the user clicked on “show more items” | no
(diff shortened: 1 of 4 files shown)


docs/src/components/RefinementList.md
```diff
@@ -25,6 +25,7 @@ Name | Type | Default | Description | Required
 ---|---|---|---|---
 attribute | string | | The attribute to refine on click | yes
 searchable | boolean | `false` | You can also search within the options of this | no
+placeholder | string | "Search here …" | If searchable is `true`, this will be the placeholder for that search box | no
```
samouss: Placeholder is a bit broad in the context of a RefinementList. Why not `searchablePlaceholder` ?

francoischalifour: We indeed call this `searchablePlaceholder` in InstantSearch.js 3 ([see new options](https://github.com/algolia/instantsearch.js/blob/838abecceacad5bb72587697eb9c64642c325882/src/widgets/refinement-list/refinement-list.js#L130)).

src/components/RefinementList.vue
```diff
@@ -121,6 +124,11 @@ export default {
       type: Boolean,
       default: false,
     },
+    searchablePlaceholder: {
+      default: 'Search here…',
```
francoischalifour: This is not part of the [RefinementList spec](https://instantsearch-css.netlify.com/widgets/refinement-list/) yet, I'm not sure we want a default placeholder here?

In InstantSearch.js 3 it's `"Search..."` for now but I would rather go without any.

Haroenv: there will always be a placeholder since we use ais-search-box, we should probably spec it

also, go back to rest 👨‍⚕️

PR #577, opened 2018-11-12
fix(CurrentRefinements): compute unique key

### LINKED ISSUE #574: v2 - Vue Warning: Duplicate keys detected when using CurrentRefinements component

Vue Warning: Duplicate keys detected

### Reproduction

Use the CurrentRefinements component with a disjunctive facet. Select multiple entries for the disjunctive facet.

### How to resolve or implement the fix?

Choose a unique key here https://github.com/algolia/vue-instantsearch/blob/2a67eeff89fd6ce3b0628f24244b13133bd114c0/src/components/CurrentRefinements.vue#L14 . Perhaps a combination of the attribute and label values appended together.

v2.0.0-beta.1

Haroenv:
fixes #574

### DIFF
--- a/src/components/CurrentRefinements.vue
+++ b/src/components/CurrentRefinements.vue
@@ -11,7 +11,7 @@
       <ul :class="suit('list')">
         <li
           v-for="item in refinements"
-          :key="item.attribute"
+          :key="item.key"
         >
           <slot
             name="item"
@@ -69,12 +69,16 @@ export default {
       return this.refinements.length === 0;
     },
     refinements() {
-      let refinements = this.state.refinements.map(item => ({
-        type: item.type,
-        attribute: item.type === 'query' ? 'query' : item.attributeName,
-        label: item.computedLabel,
-        value: item,
-      }));
+      let refinements = this.state.refinements.map(item => {
+        const attribute = item.type === 'query' ? 'query' : item.attributeName;
+        return {
+          type: item.type,
+          attribute,
+          label: item.computedLabel,
+          value: item,
+          key: `${attribute}-${item.computedLabel}`,
+        };
+      });
 
       if (this.includedAttributes) {
         refinements = refinements.filter(
--- a/src/components/__tests__/CurrentRefinements.js
+++ b/src/components/__tests__/CurrentRefinements.js
@@ -100,6 +100,29 @@ it('renders correctly (with refinements)', () => {
   expect(wrapper.html()).toMatchSnapshot();
 });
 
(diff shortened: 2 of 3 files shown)


src/components/CurrentRefinements.vue
```diff
@@ -69,12 +69,16 @@ export default {
...
     },
     refinements() {
-      let refinements = this.state.refinements.map(item => ({
-        type: item.type,
-        attribute: item.type === 'query' ? 'query' : item.attributeName,
-        label: item.computedLabel,
-        value: item,
-      }));
+      let refinements = this.state.refinements.map(item => {
+        const attribute = item.type === 'query' ? 'query' : item.attributeName;
```
tkrugg: is `this.state.refinements` the result of [getRefinements](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/utils.js#L227) ?
If yes, items with `type: "query"` now also have `attributeName: "query"`

Haroenv: I'm aware, but we can't rely on a specific InstantSearch version to be installed.

src/components/CurrentRefinements.vue
```diff
@@ -69,12 +69,16 @@ export default {
...
+        return {
+          type: item.type,
+          attribute,
+          label: item.computedLabel,
+          value: item,
+          key: `${attribute}-${item.computedLabel}`,
```
tkrugg: computedLabel can change right? (in case of a query)
Isn't the key meant to be constant in vue?

src/components/CurrentRefinements.vue
tkrugg: Can you explain why we can't force >= version on InstantSearchJS?

src/components/CurrentRefinements.vue
Haroenv: if it's a different query, it's a different element, we still need a unique key somehow, and the query will always have that issue

Haroenv: Another option is to only add the suffix where duplicates are possible (disjunctive)

francoischalifour: FYI this is how I computed the key in InstantSearch.js 3:

https://github.com/algolia/instantsearch.js/blob/838abecceacad5bb72587697eb9c64642c325882/src/components/CurrentRefinements/CurrentRefinements.js#L5-L9

Haroenv: Thanks @francoischalifour, we will revisit this when InstantSearch v3 will be incorporated here

PR #578, opened 2018-11-12
fix(PoweredBy): new design & colors

### LINKED ISSUE #571: v2 - update poweredBy

logo is old in poweredBy (colors are decided differently)

### How to resolve or implement the fix?

replace the svg by the one described in https://instantsearch-css.netlify.com/widgets/powered-by/

> Always try the latest one before opening an issue.

v2.0.0-beta.1

Haroenv:
fixes #571


src/components/PoweredBy.vue
```diff
@@ -7,23 +7,26 @@
...
+          :fill="theme === 'light' ? '#5D6494' : '#FFF'"
```
francoischalifour: You've got a validator in your component so I'm not sure this applies:

The way this is computed doesn't make `"light"` the default value, but `"dark"`.

IMO it would make more sense to reverse the condition ([this is what I did in InstantSearch.js 3](https://github.com/algolia/instantsearch.js/blob/838abecceacad5bb72587697eb9c64642c325882/src/components/PoweredBy/PoweredBy.js#L15)). This way, if the user gives some wrong string, the default theme would be `"light"`.

PR #592, opened 2018-12-12
Fixed: values.slice is not a function

u12206050:


u12206050:
Found this bug but has simple fix.

Haroenv:
Hey @u12206050, thanks for the PR! In which case did you come up with this error message?

u12206050:
I got it when trying to use the ais-refinement-list component with the category field which looks like this: ['cat1', 'cat2', 'cat3'] or ['cat1'] or [ ]

Sorry though after further research this PR seems to break other components so seems like the try catch block is actually used quite often.

I moved it: 

  ```
var values;
  try {
    values = this._helper.lastResults.getFacetValues(attribute, {
      sortBy: sortBy,
    })

    if (limit === -1) {
      return values;
    }

    return values.slice(0, limit);
  } catch (e) {
    return [];
  }
```

Haroenv:
Thanks for your feedback. Note that the search store part will disappear in v2. You can look at migration here: https://v2--vue-instantsearch.netlify.com/getting-started/migration.html

PR #596, opened 2018-12-20
feat: server side rendering

Haroenv:
1. new component: `ais-instant-search-ssr` which you use instead of `ais-instant-search` (both client and server)
2. extra methods to patch on the InstantSearch instance

## The API

This example is Nuxt, but equivalent for other SSR

```vue
<template>
  <ais-instant-search-ssr>
    <!-- regular components -->
  </ais-instant-search-ssr>
</template>

<script>
import algoliasearch from 'algoliasearch/lite';

const searchClient = algoliasearch(
  'latency',
  '6be0576ff61c053d5f9a3225e2a90f76'
);

const { instantsearch, rootMixin } = createInstantSearch({
  searchClient,
  indexName: 'movies',
});

export default {
  asyncData() {
    return instantsearch
      .findResultsState({
        query: 'hi',
        hitsPerPage: 5,
        disjunctiveFacets: ['genre'],
        disjunctiveFacetsRefinements: { genre: ['Comedy'] },
      })
      .then(() => ({
        instantSearchState: instantsearch.getState(),
      }));
  },
  beforeMount() {
    instantsearch.hydrate(this.instantSearchState);
  },
  mixins: [rootMixin],
};
</script>
```

## What's done?

- ssr example
- request on server
- serialise
- read from serialised on client
- proper names for all APIs
- move APIs from user land to library land

## What's not done? 

- avoid request on frontend (out of scope for now)


examples/ssr/src/instantsearch.js
```diff
@@ -0,0 +1,106 @@
...
+
+  search.injectToRootOrProvideOrSomethingMaybeAMixin = () => ({
+    provide() {
+      return {
+        $_ais: search,
+      };
+    },
+  });
+
+  search.injectSsrOrSomethingServer = ({ components, context }) => {
```
Haroenv: inject search results done in SSR into the server rendered page (to be used frontend)

examples/ssr/src/instantsearch.js
```diff
@@ -0,0 +1,106 @@
...
+        lastResults._state
+      );
+      search.helper.lastResults = lastResults;
+      // TODO: set the search results, and use those for the initial "search"
+      // not doing that yet is what causes the flash of no content
+      // delete window.__ALGOLIA_STATE__;
+    }
+  };
+
+  search.injectToRootOrProvideOrSomethingMaybeAMixin = () => ({
```
Haroenv: provides this InstantSearch instance from the root in main.js towards where it's consumed (frontend)

examples/ssr/src/instantsearch.js
```diff
@@ -0,0 +1,106 @@
...
+      templatesConfig: {},
+      createURL: () => '#',
+      instantSearchInstance: search,
+      searchMetadata: {
+        isSearchStalled: false,
+      },
+    });
+  };
+
+  search.injectOrHydrate = () => {
```
Haroenv: frontend: read from serialised state to set the correct state in the InstantSearch instance

examples/ssr/src/instantsearch.js
```diff
@@ -0,0 +1,106 @@
...
+      templatesConfig: {},
+      createURL: () => '#',
+      instantSearchInstance: search,
+      searchMetadata: {
+        isSearchStalled: false,
+      },
+    });
+  };
+
+  search.giveRenderPropsClientOrSomething = widget => {
```
Haroenv: used internally only (client): make sure client is aware of state from results before start, so it can render without flash

examples/ssr/src/instantsearch.js
```diff
@@ -0,0 +1,106 @@
...
+    search.helper.lastResults = lastResults;
+
+    return {
+      ais: {
+        lastResults: JSON.parse(JSON.stringify(lastResults)),
+      },
+    };
+  };
+
+  search.giveRenderPropsServerOrSomething = widget => {
```
Haroenv: used internally only (server): put state from results into widgets, to prevent them rendering empty (start() is not called on server)

src/util/createInstantSearch.js
```diff
@@ -12,13 +12,16 @@ See more info here: https://community.algolia.com/vue-instantsearch/components/I
   );
 
 export const createInstantSearch = ({ ssr }) => ({
-  name: 'AisInstantSearch',
+  name: `AisInstantSearch${ssr ? 'Ssr' : ''}`,
```
francoischalifour: Instead of concatenating letters to the name, it's often better to rewrite the whole name like so:

```suggestion
  name: ssr ? 'AisInstantSearchSsr' : 'AisInstantSearch',
```

It makes it easier to read and to find when searching in the source code.

examples/ssr/public/index.html
```diff
@@ -0,0 +1,19 @@
...
+    <link rel="icon" href="<%= BASE_URL %>favicon.ico">
+    <title>{{ title }}</title>
+    {{{ renderResourceHints() }}}
+    {{{ renderStyles() }}}
+  </head>
+  <body>
+    <!--vue-ssr-outlet-->
+    {{{ renderState() }}}
+    {{{ renderState({ contextKey: 'apolloState', windowKey: '__APOLLO_STATE__' }) }}}
+    {{{ renderState({ contextKey: 'ais', windowKey: '__ALGOLIA_STATE__' }) }}}
```
samouss: The context key could be a bit more explicit like the Apollo one.

examples/ssr/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,134 @@
+<template>
```
samouss: Do we really have to keep those files inside the example?

examples/ssr/src/main.js
```diff
@@ -0,0 +1,48 @@
...
+import algoliasearch from 'algoliasearch/lite';
+
+const searchClient = algoliasearch(
+  'latency',
+  '6be0576ff61c053d5f9a3225e2a90f76'
+);
+
+Vue.config.productionTip = false;
+
+export async function createApp({
```
samouss: We use `async / await` inside the client, but IIRC it's a bit complex to have it correctly setup. 

Do we really want to advocate for this syntax?

examples/ssr/vue.config.js
```diff
@@ -0,0 +1,14 @@
...
+    ssr: {
+      nodeExternalsWhitelist: [
+        /\.css$/,
+        /\?vue&type=style/,
+        /vue-instantsearch/,
+        /instantsearch.js/,
+      ],
+    },
+  },
+  // TODO: re-enable this
```
samouss: Re-enable or remove the comment.

rollup.config.js
```diff
@@ -66,6 +67,7 @@ export default [
         transforms: {
           dangerousForOf: true,
         },
+        objectAssign: 'Object.assign',
```
samouss: [`Object.assign`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Object/assign) does not work on IE.

src/components/InstantSearchSsr.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+  <!-- AisSsr is an empty component that will hold other widgets -->
```
samouss:
```suggestion
  <!-- AisInstantSearchSsr is an empty component that will hold other widgets -->
```

src/components/__tests__/InstantSearchSsr.js
```diff
@@ -0,0 +1,20 @@
+import InstantSearchSsr from '../InstantSearchSsr.vue';
+import { mount } from '@vue/test-utils';
+
+it('to require an injected instantsearch instance ($_ais)', () => {
+  expect(() => mount(InstantSearchSsr)).toThrowErrorMatchingInlineSnapshot(
+    `"When using SSR, it is required to use the rootMixin"`
+  );
+});
+
+// it('to use the provided instantsearch instance', () => {
```
samouss: To enable or to remove.

src/mixins/widget.js
```diff
@@ -23,6 +23,10 @@ export const createWidgetMixin = ({ connector } = {}) => ({
       this.factory = connector(this.updateState, () => {});
       this.widget = this.factory(this.widgetParams);
       this.instantSearchInstance.addWidget(this.widget);
+
+      if (this.instantSearchInstance._isSsr) {
```
samouss: We do want to force render on the server and on the client, but only the first time. With the current implementation a widget that mount after the initial render is calling `__forceRender`.

src/util/createInstantSearch.js
```diff
@@ -0,0 +1,115 @@
+import instantsearch from 'instantsearch.js/es/index';
+import algoliaHelper from 'algoliasearch-helper';
+const { SearchParameters, SearchResults } = algoliaHelper;
+
+export const createInstantSearch = ({ searchClient, indexName, options }) => {
+  const search = instantsearch({
+    searchClient,
+    indexName,
+    ...options,
```
samouss: We can move the spread operator above the two other, otherwise the options can accidentally override the provided value.

src/util/createInstantSearch.js
```diff
@@ -0,0 +1,115 @@
...
+
+  search._isSsr = true;
+
+  // main API for SSR, called in asyncData of a root component which contains instantsearch
+  search.findResultsState = params => {
+    search.helper = algoliaHelper(searchClient, indexName, {
+      // parameters set by default
+      highlightPreTag: '__ais-highlight__',
+      highlightPostTag: '__/ais-highlight__',
+      ...params,
```
samouss: We can move the spread operator above the two other, otherwise the options can accidentally override the provided value.

src/util/createInstantSearch.js
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    return search.helper.searchOnce().then(({ content: lastResults }) => {
+      search.helper.lastResults = lastResults;
```
samouss: We can add a comment that explain why we have to do that.

examples/ssr/src/entry-server.js
```diff
@@ -0,0 +1,41 @@
...
+          if (Component.asyncData) {
+            return Component.asyncData({
+              instantsearch,
+              route: router.currentRoute,
+            });
+          }
+        })
+      )
+        .then(components =>
+          instantsearch.findRoot({ components, context })
```
samouss: We can avoid this function IMO - we can expose a function like `getState()` whatever the name to retrieve the results that we get from `asyncData`. This is how Vuex works, so we might expect that people know what to do. It would looks like:

```js
Promise.all(
  matchedComponents.map(Component => {
    if (Component.asyncData) {
      return Component.asyncData({
        instantsearch,
        route: router.currentRoute
      });
    }
  })
).then(() => {
  // Probably not the best name but you got the idea
  context.instantsearchState = instantsearch.getState();

  resolve(app);
});
```

src/util/createInstantSearchComponent.js
```diff
@@ -0,0 +1,140 @@
...
+const oldApi = () =>
+  warn(
+    `Vue InstantSearch: You used the prop api-key or api-key.
+These have been replaced by search-client.
+
+See more info here: https://community.algolia.com/vue-instantsearch/components/InstantSearch.html#usage`
+  );
+
+export const createInstantSearchComponent = ({ ssr }) => ({
+  name: ssr ? 'AisInstantSearchSsr' : 'AisInstantSearch',
```
samouss: Rather than conditions that are a bit hard to read we could move each configuration that is dedicated to a specific instance inside his own file and keep all the share logic here. The function could accept a subset of the declaration and merge it to create the declaration of the component.

```js
createInstantSearchComponent({
  data() {
    return {
      instantsearchInstance: instantsearch({ ... }),
    },
  }
})
```

src/components/InstantSearchSsr.vue
Haroenv: This comment is removed with the .js file

examples/ssr/src/main.js
Haroenv: This isn't my code; but rather the one from Vue CLI, so it seems fine for me? 👍

src/components/InstantSearch.js
```diff
@@ -0,0 +1,92 @@
...
+
+const oldApi = () =>
+  warn(
+    `Vue InstantSearch: You used the prop api-key or api-key.
+These have been replaced by search-client.
+
+See more info here: https://community.algolia.com/vue-instantsearch/components/InstantSearch.html#usage`
+  );
+
+export default _objectSpread({}, createInstantSearchComponent(), {
```
samouss: I would have invert the two: 

```js
// InstantSearch.js
export default createInstantSearchRoot({
  name: '...',
  // ...
})

// createInstantSearchRoot.js
export const createInstantSearchRoot = (options) => ({
  ...options,
  common: {
    // ...
  }
})
```

src/mixins/widget.js
```diff
@@ -22,6 +22,11 @@ export const createWidgetMixin = ({ connector } = {}) => ({
       this.factory = connector(this.updateState, () => {});
       this.widget = this.factory(this.widgetParams);
       this.instantSearchInstance.addWidget(this.widget);
+
+      const { hydrated, started } = this.instantSearchInstance;
+      if ((!started && hydrated) || this.$isServer) {
```
samouss: `hydrated` is never set back to `false`, it can leads to some issues since the InstantSearch instance is global and never reset to its initial state (an other potential solution is to reset the instance on `dispose`).

```
/search (server) -> SSR -> hydrate -> forceRender -> /about (client) -> /search -> forceRender
```

The last `forceRender` should not happen because the last results might be different from the one we want to render on a fresh page. We can have a blink with the stalled results from previous search and then the fresh results for the current page. Note that the blink only occurs when the request is not cached. When it's cached it looks like Vue is batching the two renders.

blink

src/util/createInstantSearch.js
```diff
@@ -0,0 +1,115 @@
+import instantsearch from 'instantsearch.js/es';
+import algoliaHelper from 'algoliasearch-helper';
+const { SearchParameters, SearchResults } = algoliaHelper;
+import { _objectSpread } from './polyfills';
+import { warn } from './warn';
+
+export const createInstantSearch = ({ searchClient, indexName, options }) => {
```
samouss: Why do we change the signature of `instantsearch`? Is there a reason we want to scope `options`?

examples/ssr/src/entry-server.js
```diff
@@ -0,0 +1,40 @@
...
+          }
+        })
+      )
+        .then(() => {
+          // After all preFetch hooks are resolved, our store is now
+          // filled with the state needed to render the app.
+          // When we attach the state to the context, and the `template` option
+          // is used for the renderer, the state will automatically be
+          // serialized and injected into the HTML as `window.__INITIAL_STATE__`.
+          if (context.url === '/search') {
```
Haroenv: I wonder if we need a native condition in InstantSearch here 🤔

src/util/__tests__/createInstantSearch.test.js
```diff
@@ -0,0 +1,399 @@
...
+    const { rootMixin, instantsearch } = createInstantSearch({
+      searchClient: {},
+      indexName: 'bla',
+    });
+    expect(rootMixin.provide().$_ais).toBe(instantsearch);
+  });
+});
+
+describe('findResultsState', () => {
+  it('returns `ais` with lastResults', async () => {
```
Haroenv:
```suggestion
  it('modifies the state of InstantSearch', async () => {
```

examples/nuxt/pages/search.vue
```diff
@@ -0,0 +1,101 @@
...
+  asyncData() {
+    return instantsearch
+      .findResultsState({
+        query: 'hi',
+        hitsPerPage: 5,
+        disjunctiveFacets: ['genre'],
+        disjunctiveFacetsRefinements: { genre: ['Comedy'] },
+      })
+      .then(() => ({
+        instantSearchState: instantsearch.getState(),
```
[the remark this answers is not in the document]
Haroenv: choosing `algoliaState`

src/mixins/widget.js
Haroenv: TODO: clear helper on dispose

src/util/createInstantSearch.js
Haroenv: TODO: make signature the same as InstantSearch.js

examples/ssr/src/entry-server.js
Haroenv: TODO: drop warning in `getState`

src/components/InstantSearch.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+    routing: {
+      default: null,
+      validator(value) {
+        if (
+          typeof value === 'boolean' ||
+          !value.router ||
+          !value.stateMapping
+        ) {
+          warn('routing should be an object, with `router` and `stateMapping`');
```
francoischalifour: We should link to the documentation: https://www.algolia.com/doc/api-reference/widgets/instantsearch/vue/#widget-param-routing

Also perhaps warn a more human-friendly message.

src/components/InstantSearch.js
```diff
@@ -0,0 +1,91 @@
+import instantsearch from 'instantsearch.js/es/';
+import { createInstantSearchComponent } from '../util/createInstantSearchComponent';
+import { warn } from '../util/warn';
+
+const oldApi = () =>
+  warn(
+    `Vue InstantSearch: You used the prop api-key or api-key.
+These have been replaced by search-client.
+
+See more info here: https://community.algolia.com/vue-instantsearch/components/InstantSearch.html#usage`
```
francoischalifour: This link is broken. We should use the new Algolia website.

src/components/InstantSearch.js
```diff
@@ -0,0 +1,91 @@
+import instantsearch from 'instantsearch.js/es/';
+import { createInstantSearchComponent } from '../util/createInstantSearchComponent';
+import { warn } from '../util/warn';
+
+const oldApi = () =>
```
francoischalifour: Instead of creating an implicit function for that, we can store the message in a variable and warn it everywhere we want.

src/components/InstantSearchSsr.js
```diff
@@ -0,0 +1,32 @@
...
+
+export default createInstantSearchComponent({
+  name: 'AisInstantSearchSsr',
+  inject: {
+    // should be possible to configure this with {camelcase: ['error', {allow: ['^\\$_']}]}
+    // but that didn't work
+    // eslint-disable-next-line camelcase
+    $_ais: {
+      default() {
+        throw new Error('When using SSR, it is required to use the rootMixin');
```
francoischalifour:
```suggestion
        throw new Error('It is required to use `rootMixin` when using SSR.');
```

src/util/createInstantSearch.js
```diff
@@ -0,0 +1,108 @@
...
+const { SearchParameters, SearchResults } = algoliaHelper;
+import { _objectSpread } from './polyfills';
+import { warn } from './warn';
+
+export const createInstantSearch = instantSearchOptions => {
+  const search = instantsearch(instantSearchOptions);
+  const { searchClient, indexName } = instantSearchOptions;
+
+  // main API for SSR, called in asyncData of a root component which contains instantsearch
+  search.findResultsState = params => {
```
francoischalifour: Already discussed but `getResultsState` would make more sense IMO (although it's named `findResultsState` in React InstantSearch for now).

src/util/createInstantSearch.js
```diff
@@ -0,0 +1,108 @@
...
+      search.helper.lastResults = lastResults;
+    });
+  };
+
+  // make sure correct data is available in each widget's state
+  // called in widget mixin
+  search.__forceRender = widget => {
+    if (!search.helper) {
+      warn(
+        'You did not call `instantsearch.findResultsState`, which is required for ais-instant-search-ssr'
```
francoischalifour: Try not addressing directly to the user, it's almost insulting haha

```suggestion
        '`instantsearch.findResultsState()` needs to be called when using `ais-instant-search-ssr`.'
```

src/util/createInstantSearch.js
```diff
@@ -0,0 +1,108 @@
...
+    return {
+      lastResults: JSON.parse(JSON.stringify(search.helper.lastResults)),
+    };
+  };
+
+  // called before app mounts on client
+  search.hydrate = instantSearchState => {
+    if (!instantSearchState || !instantSearchState.lastResults) {
+      warn(
+        'You did not pass the result of `findResultsState` to `hydrate`, which is required'
```
francoischalifour: Same here.

```suggestion
        'The result of `findResultsState()` needs to be passed to `hydrate()`.'
```

src/util/createInstantSearchComponent.js
```diff
@@ -0,0 +1,60 @@
+import { createSuitMixin } from '../mixins/suit';
+import { version } from '../../package.json'; // rollup does pick only what needed from json
```
francoischalifour: We might not need this comment.

src/util/createInstantSearchComponent.js
```diff
@@ -0,0 +1,60 @@
...
+        indexName(indexName) {
+          this.instantSearchInstance.helper.setIndex(indexName).search();
+        },
+        stalledSearchDelay(stalledSearchDelay) {
+          // private InstantSearch.js API:
+          this.instantSearchInstance._stalledSearchDelay = stalledSearchDelay;
+        },
+        routing() {
+          throw new Error(
+            'routing configuration can not be changed dynamically at this point.' +
```
francoischalifour: Why is that?

Maybe phrase it:

> Routing is not supported [...]

Haroenv: because I couldn't figure out how to monkey patch InstantSearch to take the new Routing configuration

src/util/createInstantSearchComponent.js
Haroenv: It's more here so that someone later doesn't move it to a file for no reason

src/util/createInstantSearchComponent.js
Haroenv: (this is not from this PR by the way)

src/mixins/widget.js
samouss: We don't have the blink anymore but we still call `forceRender` even when we don't have to. The `hydrated` value is never reset to `false`. So even without a value we pay the cost of an extra render. We can check the value inside the helper or (to avoid to rely on implementation detail) sync the `hydrated` value with the helper.

src/util/testutils/client.js
```diff
@@ -0,0 +1,7 @@
+export const fakeClient = {
+  search: jest.fn(requests =>
```
samouss: Prefer the usage of a factory otherwise we share the same `jest.fn` for all the tests.

src/components/__tests__/InstantSearchSsr.js
```diff
@@ -0,0 +1,127 @@
...
+    slots: {
+      default: SearchBox,
+    },
+  });
+
+  // does not call again, since we remove `hydrated` flag on unmount
+  expect(forceRenderSpy).toHaveBeenCalledTimes(1);
+});
+
+it('does not start too many times', async () => {
```
samouss: The common lifecycles should be tested in `createInstantSearchComponent`. Only the specifics of each component should be in this test file.

src/util/createInstantSearchComponent.js
```diff
@@ -0,0 +1,67 @@
...
+        // "Note that mounted does not guarantee that all child components have also been mounted. If you want to
+        // wait until the entire view has been rendered, you can use vm.$nextTick inside of mounted"
+        this.$nextTick(() => {
+          if (!this.instantSearchInstance.started) {
+            this.instantSearchInstance.start();
+          }
+        });
+      },
+      beforeDestroy() {
+        this.instantSearchInstance.dispose();
```
samouss: We could add a condition around this because if we `dispose` before `start` it throws an error (inside `dispose` we do call `helper.getState()` but the helper is not created at that time).

PR #605, opened 2019-01-16
fix(core): make InstantSearch(Ssr) into regular .js files

Haroenv:
building further on #603


src/components/InstantSearchSsr.js
```diff
@@ -0,0 +1,4 @@
+import { createInstantSearchComponent } from '../util/createInstantSearchComponent';
+
+export default createInstantSearchComponent({ ssr: true });
+
```
samouss:
```suggestion
```

src/components/__tests__/__snapshots__/InstantSearchSsr.js.snap
```diff
@@ -0,0 +1,18 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders correctly (empty) 1`] = `
+
+<div class="ais-InstantSearch ais-InstantSearch-ssr">
```
samouss: It should be `ais-InstantSearch ais-InstantSearch--ssr`

PR #606, opened 2019-01-17
fix(components): remove faulty whitespace

Haroenv:
If an inline element has whitespace, and then a text node, Vue will preserve this whitespace inside the component, and thus it will take up bundle size (a tiny little bit)

### DIFF
--- a/src/mixins/widget.js
+++ b/src/mixins/widget.js
@@ -31,9 +31,7 @@ it's not the recommended way of making custom components.
 
 If you want to disable this message, pass { connector: true } to the mixin.
 
-Read more on using connectors:
-http://alg.li/vue-custom
-`
+Read more on using connectors: http://alg.li/vue-custom`
       );
     }
   },
--- a/src/mixins/widget.js
+++ b/src/mixins/widget.js
@@ -31,7 +31,7 @@ it's not the recommended way of making custom components.
 
 If you want to disable this message, pass { connector: true } to the mixin.
 
-Read more on using connectors: http://alg.li/vue-custom`
+Read more on using connectors: https://alg.li/vue-custom`
       );
     }
   },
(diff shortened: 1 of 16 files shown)


src/mixins/widget.js
```diff
@@ -31,9 +31,7 @@ it's not the recommended way of making custom components.
 
 If you want to disable this message, pass { connector: true } to the mixin.
 
-(TODO: update v2 link)
-Read more on using connectors: https://community.algolia.com/vue-instantsearch/getting-started/custom-components.html
-        `
+Read more on using connectors: http://alg.li/vue-custom`
```
Haroenv: I made this error message shorter while removing the newline

src/mixins/widget.js
```diff
@@ -31,9 +31,7 @@ it's not the recommended way of making custom components.
 
 If you want to disable this message, pass { connector: true } to the mixin.
 
-(TODO: update v2 link)
-Read more on using connectors: https://community.algolia.com/vue-instantsearch/getting-started/custom-components.html
-        `
+Read more on using connectors: https://alg.li/vue-custom`
```
francoischalifour: Should this be part of this PR?

Also, why using a URL shortener here and not the actual link?

Haroenv: because the real link (not the one here) was really long. Could technically be a different PR, but it isn't really that relevant

Haroenv:
no automatic way exists to do these changes

src/mixins/widget.js
francoischalifour: Right, if we want to make the link a bit more explicit, `https://alg.li/vue-custom-components` could be another name.

francoischalifour:
I came across [this thread](https://github.com/vuejs/vue/issues/3934) but it might be too complex for now.

samouss:
I would have prefer to use [`whitespace`](https://github.com/vuejs/vue/blob/dev/packages/vue-template-compiler/README.md) on it's available rather than template that are harder to read.

Haroenv:
We are not using the `vue-template-compiler` here, but good call, I'll witch to that at some later point 😎 @samouss

samouss:
We do though the Rollup plugin.

PR #609, opened 2019-01-22
More clearly clarify how to use URL Synchronization

jryd:
The docs, as they are written and structured currently, suggest that synchornization is only possible with Vue Router.

The proposed changes make this more generic both in the navigation menu and in the documentation itself.

Out of the box, if you just create a `routing` object with `historyRouter()` and `simpleMapping()` then this will work whether you are using Vue Router or not.

I've then separated the section with a heading for times when you may want to customise this functionality (including use with Vue Router).

### DIFF
--- a/docs/src/advanced/url-sync.md
+++ b/docs/src/advanced/url-sync.md
@@ -1,19 +1,19 @@
 ---
-title: Synchronize with Vue Router
+title: URL Synchronization
 mainTitle: Advanced
 layout: main.pug
 category: Advanced
 withHeadings: true
 navWeight: 3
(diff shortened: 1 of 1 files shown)


Haroenv:
This guide has been updated in the new documentation (linked from the top). Does that also solve your suggestions? https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/vue/

Haroenv:
Perfect. Soon we will redirect this website completely to the main one, but I haven't yet found the time to do the mapping of one to the other.

PR #611, opened 2019-01-24
chore(examples): use published version of Vue InstantSearch

Haroenv:
Still to do: make Nuxt work with nodeExternalsWhitelist


Haroenv:
This will be working with Nuxt 2.4.0, which should be released soon (see https://github.com/nuxt/nuxt.js/pull/4850)

Haroenv:
(I merged so that at least the examples are correct, even if the Nuxt one doesn't work yet)

PR #612, opened 2019-01-25
feat(core): update to InstantSearch.js v3

Haroenv:
### `ais-current-refinements`

DOM is now nested, new slot: `refinement` for inner item (`item` still exists, and is outer item)

### `ais-breadcrumbs`

template / scope "options" now uses `label / value` instead of `name / value`


src/components/ClearRefinements.vue
```diff
@@ -38,17 +38,25 @@ export default {
   props: {
     excludedAttributes: {
       type: Array,
-      default: () => ['query'],
+      // explicitly no default, since in/ex are incompatible
```
samouss: We could use the real word, I've struggled to understand at first.

src/components/CurrentRefinements.vue
```diff
@@ -5,28 +5,41 @@
...
+          <span :class="suit('label')">{{ item.attribute | capitalize }}: </span>
+          <span
+            v-for="refinement in item.refinements"
+            :key="createItemKey(refinement)"
+            :class="suit('category')"
           >
-            <span :class="suit('item')">
-              <span :class="suit('label')">{{ item.attribute | capitalize }}: {{ item.label }}</span>
+            <slot
+              name="item"
```
samouss: The slots could be moved on the `li`, users might want to be able to customize the label.

Haroenv: Could probably be two slots, one for the inner item, one for the containing attribute

src/components/CurrentRefinements.vue
```diff
@@ -66,41 +79,21 @@ export default {
...
-            : true),
+        includedAttributes: this.includedAttributes,
+        excludedAttributes: this.excludedAttributes,
+        transformItems: this.transformItems,
       };
     },
   },
+  methods: {
+    createItemKey({ attribute, value, type, operator }) {
+      return [attribute, type, value, operator].join(':');
```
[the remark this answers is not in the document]
Haroenv: Ah, that's why you added the boolean filter. However if you join `undefined` it will simply be coerced to an empty string, and thus it won't have an impact wether the string is unique within the list or not

src/components/CurrentRefinements.vue
```diff
@@ -5,26 +5,46 @@
...
-            :createURL="() => state.createURL(item.value)"
+            :createURL="state.createURL"
           >
-            <span :class="suit('item')">
-              <span :class="suit('label')">{{ item.attribute | capitalize }}: {{ item.label }}</span>
-              <button
-                :class="suit('delete')"
-                @click="state.refine(item.value)"
-              >✕</button>
+            <span :class="suit('label')">{{ item.attribute | capitalize }}: </span>
```
francoischalifour: If you rely on [InstantSearch.js 3.1.0](https://unpkg.com/instantsearch.js@3.1.0/), you can now use `item.label`.

src/components/__tests__/NumericMenu.js
```diff
@@ -78,14 +78,14 @@ it('accepts an items prop', () => {
 
   const wrapper = mount(NumericMenu, {
     propsData: props,
-  });
+  }); // ok
```
samouss:
```suggestion
  });
```

stories/CurrentRefinements.stories.js
```diff
@@ -60,24 +60,69 @@ storiesOf('ais-current-refinements', module)
       transformItems(items) {
         return items.map(item =>
           Object.assign({}, item, {
-            label: item.label.toUpperCase(),
+            attribute: item.attribute.toLocaleUpperCase(),
```
samouss: You can use `label` rather than `attribute`.

src/components/HierarchicalMenu.vue
```diff
@@ -110,7 +110,9 @@ export default {
     widgetParams() {
       return {
         attributes: this.attributes,
-        limit: this.showMore ? this.showMoreLimit : this.limit,
+        limit: this.limit,
+        showMore: this.showMore,
```
samouss: We have to use the `showMore` capabilities of the connector.

stories/CurrentRefinements.stories.js
Haroenv:
```suggestion
            attribute: item.label.toLocaleUpperCase(),
```

src/components/__tests__/HierarchicalMenu.js
```diff
@@ -113,7 +113,10 @@ const defaultState = {
     microsoft,
   ],
   refine: () => {},
-  createURL: () => {},
+  createURL: () => {}, // TODO: turn this into a string
```
francoischalifour: Is that still to do?

Haroenv: good catch, not necessary

```suggestion
  createURL: () => {},
```

PR #621, opened 2019-02-14
docs(website): redirect Netlify website to Algolia

francoischalifour:
This redirects the Vue InstantSearch Netlify website index page to the Algolia documentation website.

I just copied the Netlify `_redirects` file to the website `dist/` folder during the build process. This should be enough for now since we'll get rid of the docgen website.

### DIFF
--- a/docs/_redirects
+++ b/docs/_redirects
@@ -1,0 +1,1 @@
+/ https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/vue/
--- a/docs/package.json
+++ b/docs/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "dev": "babel-node start.js",
-    "build:dev": "babel-node build.js && cd .. && yarn storybook:build && yarn examples:build",
+    "build:dev": "babel-node build.js && (cd .. && yarn storybook:build && yarn examples:build) && cp _redirects ./dist/",
     "build": "NODE_ENV='production' babel-node build.js && cd .. && yarn storybook:build && yarn examples:build",
     "deploy": "yarn run build && gh-pages -d dist"
   },


docs/_redirects
```diff
@@ -0,0 +1 @@
+/ https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/vue/
```
tkrugg: I think we talked about this but maybe I misunderstood it.
Shouldn't there be a redirect for all the other pages:
example
-  `/getting-started/installing.html` -> corresponding page in the new docs
- `/getting-started/using-components.html` -> corresponding page in the new docs
...

docs/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "dev": "babel-node start.js",
-    "build:dev": "babel-node build.js && cd .. && yarn storybook:build && yarn examples:build",
+    "build:dev": "babel-node build.js && (cd .. && yarn storybook:build && yarn examples:build) && cp _redirects ./dist/",
```
Haroenv: I think we can avoid this by putting it in `assets` which already is copied?

docs/package.json
```diff
@@ -5,7 +5,7 @@
...
     "build": "NODE_ENV='production' babel-node build.js && cd .. && yarn storybook:build && yarn examples:build",
```
Haroenv: regular build should add these redirects too?

docs/package.json
tkrugg: I was going to say 
> this should be on build rather than build:dev no?
but I just noticed Netlify actually ran build:dev. Isn't that a little strange?
probably not worth changing a this point though 😄

docs/_redirects
francoischalifour: Not for Vue ([see reference](https://docs.google.com/document/d/1B-eWm7WJapqE5fMmI8HwsT_hZF_z6s0A0QWyA08SxNk/edit#heading=h.fcxqldkv871a)).

docs/package.json
francoischalifour: It's not used by Netlify.

docs/package.json
francoischalifour: Yeah I thought it was weird but that's how it is right now (not worth changing now).

francoischalifour: The `_redirects` file need to be at the root (it's not picked up by Netlify if it's in `assets/`).

Haroenv: It's doing build:dev because the prod build was too long and it times out

francoischalifour: The `build:dev` already turned my laptop's fan like crazy haha

PR #622, opened 2019-02-15
docs(v1): add redirect to root & noindex to rest

Haroenv:
depends on https://github.com/algolia/doc/pull/2662 to be released first

This doesn't need to be put on gh-pages, since it will be replaced by netlify

### DIFF
--- a/docs/src/_redirects
+++ b/docs/src/_redirects
@@ -1,0 +1,1 @@
+/ https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/vue/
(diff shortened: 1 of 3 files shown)


docs/src/_redirects
```diff
@@ -0,0 +1 @@
+/ https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/vue/
```
francoischalifour: Because of this redirection, the v1 website won't be easily accessible via the URL (https://community.algolia.com/vue-instantsearch/ points to Algolia).

Do we want to create the endpoint `https://community.algolia.com/vue-instantsearch/v1`? This is what we do for [InstantSearch.js v2](https://community.algolia.com/instantsearch.js/v2).

Haroenv: that's a possibility, but would break all URLs. What I did for now is change the URL we use to the getting started URL and thus keep all existing URLs. Maybe some kind of redirection of all things to v1 would be possible as you said. 🤔

samouss:
@Haroenv could we move forward with this?

PR #627, opened 2019-02-20
feat(StateResults): give access to search parameters

Haroenv:
This changes the internal `connectStateResults` connector so that it exposes the state too, but without the methods.

This is not a breaking change in the widget, so should be straightforward to migrate to the new usage:

```diff
<ais-state-results>
- <template slot-scope="{ query, hits }">
+ <template slot-scope="{ results: { query, hits } }">
  </template>
</ais-state-results>
```

It allows you to read from the state as well, avoiding race conditions if for example hiding hits based on the query.

```diff
<ais-state-results>
- <div slot-scope="{ query }">
+ <div slot-scope="{ state: { query } }">
    <ais-hits v-if="query" />
  </div>
</ais-state-results>
```

TO DO:
- [ ] decide if the connector should be moved to InstantSearch.js instead
- [x] alternative for Object.assign
- [x] prevent breaking change


src/components/StateResults.vue
```diff
@@ -1,56 +1,73 @@
...
-        widgetParams,
+        widgetParams
       },
       true
     );
   },
 
-  render({ results, instantSearchInstance }) {
+  render({ results, instantSearchInstance, state }) {
     const resultsCopy = Object.assign({}, results);
```
Haroenv: let's find an alternative for shallow copying here, since it's not available in IE11

src/components/StateResults.vue
```diff
@@ -1,56 +1,73 @@
...
+  name: "AisStateResults",
   mixins: [
     createWidgetMixin({ connector: connectStateResults }),
-    createSuitMixin({ name: 'StateResults' }),
+    createSuitMixin({ name: "StateResults" })
   ],
+  computed: {
+    stateResults() {
+      // TODO: just pass results & state separately in the next major version
+      const stateResults = this.state.results;
```
Haroenv: This is recursively setting the properties now, in contrast with the Object.assign version, which only set it one level deep.

src/components/StateResults.vue
samouss: Switch to Babel?

src/components/StateResults.vue
```diff
@@ -1,56 +1,73 @@
...
+      <pre>results: {{ Object.keys(state.results) }}</pre>
+      <pre>state: {{ Object.keys(state.state) }}</pre>
     </slot>
   </div>
 </template>
 
 <script>
-import { createSuitMixin } from '../mixins/suit';
-import { createWidgetMixin } from '../mixins/widget';
+import { createSuitMixin } from "../mixins/suit";
```
samouss: The Prettier settings seems not correct.

src/components/StateResults.vue
```diff
@@ -1,56 +1,73 @@
...
     createWidgetMixin({ connector: connectStateResults }),
-    createSuitMixin({ name: 'StateResults' }),
+    createSuitMixin({ name: "StateResults" })
   ],
+  computed: {
+    stateResults() {
+      // TODO: just pass results & state separately in the next major version
+      const stateResults = this.state.results;
+      stateResults.results = this.state.results;
+      stateResults.state = this.state.state;
```
samouss: I would have create a new object even if we already do it inside the connector. It helps for the comprehension, at first I thought we were mutating the helper results.

stories/StateResults.stories.js
```diff
@@ -84,14 +84,80 @@ storiesOf('ais-state-results', module)
...
+            <p v-if="hits.length === 0">
+              No results found for the query: <q>{{ query }}</q>
+            </p>
+          </template>
+        </ais-state-results>
         <ais-hits />
+      </div>
+    `,
+  }))
+  .add('no results (old usage)', () => ({
```
samouss: We can drop this usage to avoid people finding it.

src/components/__tests__/StateResults.js
```diff
@@ -21,33 +24,87 @@ it("doesn't render if no results", () => {
...
+        expect(props).toEqual(expect.objectContaining(results));
+        expect(props.results).toEqual(results);
+        expect(props.state).toEqual(state);
+      },
     },
   });
 });
 
-it('allows default slot to render whatever they want (truthy query)', () => {
+it('allows default slot to render whatever they want (modern scope)', () => {
```
samouss: We could have kept the tests that were present before but scope them under `describe('legacy usage')`. Then create new ones with the same state for the new usage.

Haroenv:
In the mean time, here's a sandbox on how to make this in userland: https://codesandbox.io/s/vue-instantsearch-app-hk54z (and use it for debugging state, and setting default value)

francoischalifour:
What's the cost of waiting for Vue InstantSearch v3 to add this functionality and having this as a breaking change?

Haroenv:
I’d prefer people being able to use this new version before adding it as breaking change in v3

Haroenv:
I’ve also been wanting to use the state frequently, and have had to rewrite this in user land a few times the last few weeks for support questions

src/connectors/connectStateResults.js
```diff
@@ -0,0 +1,45 @@
...
+        instantSearchInstance,
+        widgetParams,
+      },
+      true
+    );
+  },
+
+  render({ results, instantSearchInstance, state }) {
+    const resultsCopy = Object.keys(results).reduce((acc, key) => {
+      if (key !== '_state') {
```
francoischalifour: Any reasons for removing the `_state` key? React InstantSearch doesn't:

SearchResults in RIS

Haroenv: I want to avoid giving access to the helper functions here, only the values. Might be overkill to remove it, but it doesn't cost much to remove it.

francoischalifour: It would have been nice to align the two but I guess we can do that once we have this kind of connector in InstantSearch.js.

src/connectors/connectStateResults.js
```diff
@@ -0,0 +1,45 @@
+const connectStateResults = (renderFn, unmountFn) => (widgetParams = {}) => ({
```
francoischalifour: There's not tests for this connector. It would be for asserting `isFirstRender` values and calling `unmountFn`. Is that on purpose?

src/connectors/connectStateResults.js
```diff
@@ -0,0 +1,45 @@
...
+        results: undefined,
+        instantSearchInstance,
+        widgetParams,
+      },
+      true
+    );
+  },
+
+  render({ results, instantSearchInstance, state }) {
+    const resultsCopy = Object.keys(results).reduce((acc, key) => {
```
yannickcr: This could be simplified using destructuring:

```js
const { _state, ...resultsCopy } = results;
```

src/connectors/connectStateResults.js
```diff
@@ -0,0 +1,45 @@
...
+      if (key !== '_state') {
+        // eslint-disable-next-line no-param-reassign
+        acc[key] = results[key];
+      }
+      return acc;
+    }, {});
+
+    const stateCopy = Object.keys(state).reduce((acc, key) => {
```
yannickcr: This could be simplified using the `_objectSpread` helper:

```js
const stateCopy = _objectSpread({}, state);
```

src/connectors/connectStateResults.js
Haroenv: Vue InstantSearch does not have babel set up for object.spread, because it uses bublé instead

yannickcr: Yes, but here we're using the rest syntax, not spread.

Haroenv: rest & spread both aren't supported

yannickcr: I tested and it gets compiled to:

```js
var _state = results._state;
var rest = objectWithoutProperties( results, ["_state"] );
var resultsCopy = rest;
```

Seems correct to me :thinking:

yannickcr: Another solution using `_objectSpread` helper:

```js
const resultsCopy = _objectSpread({}, results);
delete resultsCopy._state;
```

src/connectors/connectStateResults.js
```diff
@@ -12,19 +18,9 @@ const connectStateResults = (renderFn, unmountFn) => (widgetParams = {}) => ({
...
 
   render({ results, instantSearchInstance, state }) {
-    const resultsCopy = Object.keys(results).reduce((acc, key) => {
-      if (key !== '_state') {
-        // eslint-disable-next-line no-param-reassign
-        acc[key] = results[key];
-      }
-      return acc;
-    }, {});
+    const resultsCopy = _objectSpread({}, results);
```
yannickcr: You do not remove the `_state` key from the copy anymore, is it normal?

Haroenv: Yes, too much discussion on it, and the chance that someone will abuse it is small I think

src/connectors/connectStateResults.js
```diff
@@ -0,0 +1,41 @@
+import { _objectSpread } from '../util/polyfills';
+
+const connectStateResults = (renderFn, unmountFn = () => {}) => (
+  widgetParams = {}
+) => ({
+  $$type: 'vue.stateResults',
```
francoischalifour: Why is this the only VIS widget having the `$$type` key? (we're still using IS.js v3 here)

Haroenv: I added the type in preparation for v4, so we don't need to add it later.

francoischalifour: It means that it's going to be inconsistent in this version.

(We should probably go for `avis` since it's `ais` in IS.js.)

Haroenv: Avis is a regular word, so would confuse me. 

You're right that it's inconsistent, but I don't see the issue, since it isn't used anywhere. WDYT?

francoischalifour: Inconsistency brings confusion, there's no reasons to have this key in this VIS version :/

Besides, we rushed the decision to prefix it with `vue.`.

If we're at VIS 5 and look back at VIS 2 and see this key, we'll be confused: "What is this for?" "Did people rely on it already?" "We've never used this prefix elsewhere."

Haroenv: That makes sense, do you suggest no $$type at all, also in v3?

francoischalifour: Yeah, let's not introduce it until we need it. We'll add `connectStateResults` in IS.js at some point anyway (perhaps with a better name?).

Haroenv: https://github.com/algolia/vue-instantsearch/commit/fcf6469fff974bb4f218bbc53db6c426e32c2f5d

PR #635, opened 2019-03-06
fix(umd): prevent inclusion of Vue in bundle

Haroenv:
This is fixed in two spots:

1. add Vue to externals for the UMD (used in Panel)
2. use the Vue object passed to the install function instead of importing it ourselves (in the plugin)

Furthermore this also fixes the plugin which needlessly exposes all widgets (you can already import them easily, since the default export is being put on the window)

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -1,12 +1,11 @@
 /* eslint import/namespace: ['error', { allowComputed: true }]*/
-import Vue from 'vue';
 
 import * as widgets from './widgets';
 
-export const plugin = Object.assign({}, widgets, {
-  install() {
+export const plugin = {
+  install(Vue) {
     Object.keys(widgets).forEach(widgetName => {
       Vue.component(widgets[widgetName].name, widgets[widgetName]);
     });
   },
-});
+};
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -3,9 +3,9 @@
 import * as widgets from './widgets';
 
 export const plugin = {
-  install(Vue) {
+  install(localVue) {
     Object.keys(widgets).forEach(widgetName => {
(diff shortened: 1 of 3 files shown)


francoischalifour:
This warning appears while building.

```
src/instantsearch.umd.js → dist/vue-instantsearch.js...
│   Bundle Size:  224.06 KB                  │
│   Minified and Gzipped Size:  57.93 KB     │
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
vue (guessing 'Vue')
```

Haroenv:
good call, I didn't have it earlier, but can replicate now @francoischalifour

Haroenv:
by the way, nice effect of this:

## before

```
src/instantsearch.umd.js → dist/vue-instantsearch.js...
│   Bundle Size:  284.17 KB                  │
│   Minified Size:  281.05 KB                │
│   Gzipped Size:  79.27 KB                  │
```

## after

```
src/instantsearch.umd.js → dist/vue-instantsearch.js...
│   Bundle Size:  224.09 KB                  │
│   Minified Size:  221.15 KB                │
│   Gzipped Size:  57.93 KB                  │
```

francoischalifour:
Nice decrease!

src/plugin.js
```diff
@@ -1,12 +1,11 @@
 /* eslint import/namespace: ['error', { allowComputed: true }]*/
-import Vue from 'vue';
 
 import * as widgets from './widgets';
 
-export const plugin = Object.assign({}, widgets, {
-  install() {
+export const plugin = {
+  install(Vue) {
```
samouss:
```suggestion
  install(localVue) {
```

We could rename it to indicate that it's clearly not the global one, it avoids a bit of confusion.

samouss:
It's better without Vue embed!

PR #637, opened 2019-03-06
ci: configure bundlesize

Haroenv:
on purpose expecting this branch to fail before #635


package.json
```diff
@@ -45,6 +46,7 @@
   },
   "dependencies": {
     "algoliasearch-helper": "^2.26.1",
+    "bundlesize": "^0.17.1",
```
francoischalifour: This should be in `devDependencies`.

package.json
```diff
@@ -52,6 +53,7 @@
     "vue": "^2.4.0"
   },
   "devDependencies": {
+    "bundlesize": "0.17.1",
```
francoischalifour: It should be right after `babel-preset-es2015`.

Haroenv: I expected yarn to order this and didn't check :(

package.json
```diff
@@ -32,7 +32,8 @@
     "docs:build:dev": "cd docs && yarn && yarn build:dev",
     "docs:watch": "cd docs && yarn && yarn dev",
     "docs:deploy": "cd docs && yarn && yarn deploy",
-    "test": "jest --coverage && yarn run lint",
+    "test": "jest --coverage && yarn run lint && yarn run test:size",
```
francoischalifour: Probably aside from this PR but I think that the `test` script owns too many responsibilities.

Haroenv: Fairly nitpicky, but this is on purpose. Running `yarn test` will verify everything, if you just want to run the tests, I always just run `yarn jest`. We can align on that over all libraries though

PR #648, opened 2019-03-18
fix(stories): update range slider component to v3

### LINKED ISSUE #647: ais-range-slider does not work out of the box
**Bug 🐞**

When using the range slider, the examples in the documentation do not work.
https://www.algolia.com/doc/api-reference/widgets/range-slider/vue/

The slider does not filter the results.   

Are we meant to pass a `v-model` value to the range slider?

https://codesandbox.io/s/98k5vqx5lr

When using the slider in the docs, it should filter the results.

    "algoliasearch": "3.30.0",
    "instantsearch.css": "7.1.1",
    "vue": "2.5.17",
    "vue-instantsearch": "2.0.0",
    "vue-slider-component": "^3.0.15"

> Always try the latest one before opening an issue.

Thanks in advance

Haroenv:
fixes #647 (the other fix is in the doc repo: https://github.com/algolia/doc/pull/2818)

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "vue-json-tree": "0.3.3",
     "vue-loader": "14.2.2",
     "vue-slider-component": "3.0.15",
-    "vue-template-compiler": "2.5.17",
+    "vue-template-compiler": "2.5.18",
     "vuetify": "1.5.3"
   },
   "bundlesize": [
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "vue-jest": "2.6.0",
     "vue-json-tree": "0.3.3",
     "vue-loader": "14.2.2",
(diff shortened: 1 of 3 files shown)


francoischalifour:
The story doesn't display anything: https://deploy-preview-648--vue-instantsearch.netlify.com/stories/?selectedKind=ais-range-input&selectedStory=with%20vue-slider-component.

Vuetify requires Vue >= 2.5.18.

Haroenv:
I see, it's hidden because the CSS wasn't imported

francoischalifour:
I don't think we should recommend using the `change` event because it triggers a huge number of requests.

vue-slider

The same goes for the story "with vuetify slider".

Haroenv:
That was already the case before this PR, and I don't really see a nice way to implement "on submit" on either of the slider examples

PR #649, opened 2019-03-21
fix(toggle-refinement): display label

francoischalifour:
The required `label` of the [`ais-toggle-refinement`](https://www.algolia.com/doc/api-reference/widgets/toggle-refinement/vue/) widget was not used in the template, thus ignored. The [deployed story](https://vue-instantsearch.netlify.com/stories/?selectedKind=ais-toggle-refinement&selectedStory=default) should display "Free shipping" but displays the attribute name "free_shipping".

This issue was reported by a user.

### DIFF
--- a/src/components/ToggleRefinement.vue
+++ b/src/components/ToggleRefinement.vue
@@ -18,7 +18,7 @@
           :checked="state.value.isRefined"
           @change="state.refine(state.value)"
         >
-        <span :class="suit('labelText')">{{ state.value.name }}</span>
+        <span :class="suit('labelText')">{{ label }}</span>
         <span
           v-if="state.value.count !== null"
           :class="suit('count')"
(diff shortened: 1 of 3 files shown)


Haroenv:
Weird, I solved that issue I thought

Haroenv:
I'm surprised that this didn't break any tests 🤔

francoischalifour:
So am I. The Vue testing environment seems to be different:

https://github.com/algolia/vue-instantsearch/blob/e3010f61a397da721b75256925f27e41b7992c29/src/components/__tests__/ToggleRefinement.js#L72-L80

https://github.com/algolia/vue-instantsearch/blob/e3010f61a397da721b75256925f27e41b7992c29/src/components/__tests__/__snapshots__/ToggleRefinement.js.snap#L3-L16

Haroenv:
Can you change the test so that the name isn't the same as the label? 

https://github.com/algolia/vue-instantsearch/blob/e3010f61a397da721b75256925f27e41b7992c29/src/components/__tests__/ToggleRefinement.js#L9

francoischalifour:
All right – the test covers more cases now.

Haroenv:
Netlify failures are the fault of their cache, it works consistently on my machine

Haroenv:
This is technically a breaking change, but small enough to ignore IMO

PR #650, opened 2019-03-22
fix(ua): change the User-Agent to use the new specs lib (version)

robertmogos:


src/components/__tests__/InstantSearch.js
```diff
@@ -245,9 +246,10 @@ it('will call client.addAlgoliaAgent if present', () => {
     },
   });
 
-  expect(client.addAlgoliaAgent).toHaveBeenCalledTimes(1);
-  expect(client.addAlgoliaAgent.mock.calls[0][0]).toMatch(
-    /Vue InstantSearch \([a-z0-9.-]+\)/
+  expect(client.addAlgoliaAgent).toHaveBeenCalledTimes(2);
+  expect(client.addAlgoliaAgent).toHaveBeenCalledWith(`Vue (${Vue.version})`);
```
francoischalifour: I think you can chain the calls like so:

```js
expect(client.addAlgoliaAgent).toHaveBeenCalledWith([
  `Vue (${Vue.version})`,
  `Vue InstantSearch (${version})`
]);
```

robertmogos: Nope, it doesn't work

francoischalifour: Sorry, I meant:

```js
expect(client.addAlgoliaAgent).toHaveBeenCalledWith(
  `Vue (${Vue.version})`,
  `Vue InstantSearch (${version})`
);
```

robertmogos: That one neither :)

PR #652, opened 2019-03-28
feat(query-rules): add new widgets

Haroenv:
- [x] initial implementation
- [x] add tests
- [x] correct DOM for CustomData
- [x] correct slots for CustomData
- [x] change import path
- [x] require a compatible version of InstantSearch.js
- [x] compatible stories with InstantSearch

### DIFF
--- a/src/components/QueryRuleContext.js
+++ b/src/components/QueryRuleContext.js
@@ -1,6 +1,6 @@
 import { createSuitMixin } from '../mixins/suit';
 import { createWidgetMixin } from '../mixins/widget';
-import { connectQueryRules } from '../../../instantsearch.js/es/connectors';
+import { connectQueryRules } from 'instantsearch.js/es/connectors';
 
 export default {
   name: 'AisQueryRuleContext',
--- a/src/components/QueryRuleContext.js
+++ b/src/components/QueryRuleContext.js
@@ -1,0 +1,35 @@
+import { createSuitMixin } from '../mixins/suit';
+import { createWidgetMixin } from '../mixins/widget';
+import { connectQueryRules } from '../../../instantsearch.js/es/connectors';
+
+export default {
+  name: 'AisQueryRuleContext',
+  mixins: [
+    createSuitMixin({ name: 'QueryRuleContext' }),
+    createWidgetMixin({
+      connector: connectQueryRules,
+    }),
+  ],
+  props: {
+    trackedFilters: {
+      type: Object,
+      required: true,
+    },
+    transformRuleContexts: {
+      type: Function,
+      required: false,
+      default: ruleContext => ruleContext,
+    },
+  },
+  computed: {
+    widgetParams() {
+      return {
+        trackedFilters: this.trackedFilters,
+        transformRuleContexts: this.transformRuleContexts,
+      };
+    },
+  },
+  render() {
+    return null;
+  },
+};
--- a/src/components/QueryRuleContext.js
+++ b/src/components/QueryRuleContext.js
@@ -18,7 +18,7 @@ export default {
     transformRuleContexts: {
       type: Function,
       required: false,
-      default: ruleContext => ruleContext,
+      default: ruleContexts => ruleContexts,
     },
   },
   computed: {
--- a/src/components/QueryRuleCustomData.vue
+++ b/src/components/QueryRuleCustomData.vue
@@ -13,7 +13,7 @@
 <script>
 import { createSuitMixin } from '../mixins/suit';
 import { createWidgetMixin } from '../mixins/widget';
-import { connectQueryRules } from '../../../instantsearch.js/es/connectors';
+import { connectQueryRules } from 'instantsearch.js/es/connectors';
 
 export default {
   name: 'AisQueryRuleCustomData',
--- a/src/components/QueryRuleCustomData.vue
+++ b/src/components/QueryRuleCustomData.vue
@@ -1,9 +1,18 @@
 <template>
-  <div v-if="state" :class="suit()">
+  <div
+    v-if="state"
+    :class="suit()"
+  >
     <slot :items="state.items">
-      <div v-for="(item, key) in state.items" :key="key">
-        <slot name="item" :item="item">
-          <pre>{{item}}</pre>
+      <div
+        v-for="(item, key) in state.items"
+        :key="key"
+      >
+        <slot
+          name="item"
+          :item="item"
+        >
+          <pre>{{ item }}</pre>
         </slot>
       </div>
     </slot>
--- a/src/components/QueryRuleCustomData.vue
+++ b/src/components/QueryRuleCustomData.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="state">
+  <div v-if="state" :class="suit()">
     <slot :items="state.items">
       <div v-for="(item, key) in state.items" :key="key">
         <slot name="item" :item="item">
@@ -27,7 +27,7 @@ export default {
     transformItems: {
       type: Function,
       required: false,
-      default: customData => customData,
+      default: items => items,
     },
   },
   computed: {
--- a/src/components/QueryRuleCustomData.vue
+++ b/src/components/QueryRuleCustomData.vue
@@ -46,8 +46,5 @@ export default {
       };
     },
   },
-  render() {
-    return null;
-  },
 };
 </script>
--- a/src/components/QueryRuleCustomData.vue
+++ b/src/components/QueryRuleCustomData.vue
@@ -1,0 +1,42 @@
+<template>
+  <div v-if="state">
+    <slot :items="state.items">
+      <div v-for="(item, key) in state.items" :key="key">
+        <pre>{{item}}</pre>
+      </div>
+    </slot>
+  </div>
+</template>
+
+<script>
+import { createSuitMixin } from '../mixins/suit';
+import { createWidgetMixin } from '../mixins/widget';
+import { connectQueryRules } from '../../../instantsearch.js/es/connectors';
+
+export default {
+  name: 'AisQueryRuleCustomData',
+  mixins: [
+    createSuitMixin({ name: 'QueryRuleCustomData' }),
+    createWidgetMixin({
+      connector: connectQueryRules,
+    }),
+  ],
+  props: {
+    transformItems: {
+      type: Function,
+      required: false,
+      default: customData => customData,
+    },
+  },
+  computed: {
+    widgetParams() {
+      return {
+        transformItems: this.transformItems,
+      };
+    },
+  },
+  render() {
+    return null;
+  },
+};
+</script>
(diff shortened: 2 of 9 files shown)


src/components/QueryRuleContext.js
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  props: {
+    trackedFilters: {
+      type: Object,
+      required: true,
+    },
+    transformRuleContexts: {
+      type: Function,
+      required: false,
+      default: ruleContext => ruleContext,
```
francoischalifour:
```suggestion
      default: ruleContexts => ruleContexts,
```

src/components/QueryRuleCustomData.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <div v-if="state">
```
francoischalifour: The `ais-QueryRuleCustomData` CSS class needs to be added.

src/components/QueryRuleCustomData.vue
```diff
@@ -0,0 +1,44 @@
...
+    createSuitMixin({ name: 'QueryRuleCustomData' }),
+    createWidgetMixin({
+      connector: connectQueryRules,
+    }),
+  ],
+  props: {
+    transformItems: {
+      type: Function,
+      required: false,
+      default: customData => customData,
```
francoischalifour:
```suggestion
      default: items => items,
```

src/components/QueryRuleCustomData.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <div v-if="state">
+    <slot :items="state.items">
+      <div v-for="(item, key) in state.items" :key="key">
+        <slot name="item" :item="item">
+          <pre>{{item}}</pre>
```
francoischalifour: We don't provide any default template in InstantSearch.js. Do we need to add one there?

src/widgets.js
```diff
@@ -22,6 +22,10 @@ export { default as AisNumericMenu } from './components/NumericMenu.vue';
 export { default as AisPagination } from './components/Pagination.vue';
 export { default as AisPanel } from './components/Panel.vue';
 export { default as AisPoweredBy } from './components/PoweredBy.vue';
+export { default as AisQueryRuleContext } from './components/QueryRuleContext';
```
francoischalifour: Missing `.vue` extension.

stories/QueryRuleContext.stories.js
```diff
@@ -0,0 +1,136 @@
...
+    `,
+    data() {
+      return {
+        trackedFilters: {
+          genre: () => ['Thriller', 'Drama'],
+        },
+      };
+    },
+  }))
+  .add('with filter out "thriller" in generated rules', () => ({
```
francoischalifour: If we want to create more stories, we can also reproduce the previous story with an initial rule context (set with `transformRuleContexts`), but without refinement.

src/components/QueryRuleCustomData.vue
Haroenv: I found it really useful for debugging before committing to a certain way to display them, I implemented the same in Angular InstantSearch

src/widgets.js
Haroenv: Nope, this is a .js file 😃

francoischalifour:
[InstantSearch.js 3.3.0](https://unpkg.com/instantsearch.js@3.3.0/) has been released. We can move forward with this PR.

stories/QueryRuleContext.stories.js
francoischalifour: We can replace this story by "with initial rule context".

See in Angular InstantSearch: https://github.com/algolia/angular-instantsearch/blob/37512b2732be6e0b1c6320d92ea1d87fb1f2b7c6/examples/storybook/src/stories/QueryRuleContext.stories.ts#L117

stories/QueryRuleCustomData.stories.js
```diff
@@ -0,0 +1,122 @@
...
+            <header>
+              <strong><ais-highlight attribute="title" :hit="item"/></strong>
+            </header>
+          </article>
+        </li>
+      </ol>
+      `,
+    })
+  )
+  .add('simple usage', () => ({
```
francoischalifour:
```suggestion
  .add('default', () => ({
```

src/components/QueryRuleCustomData.vue
```diff
@@ -0,0 +1,53 @@
...
+    },
+  },
+  computed: {
+    widgetParams() {
+      return {
+        transformItems: this.transformItems,
+      };
+    },
+  },
+  render() {
```
samouss: This is not supposed to be there correct?

src/components/__tests__/QueryRuleContext.js
```diff
@@ -0,0 +1,33 @@
...
+});
+
+it('accepts only trackedFilters and transformRuleContexts', () => {
+  const trackedFilters = {};
+  const transformRuleContexts = jest.fn();
+  const wrapper = mount(QueryRuleContext, {
+    propsData: {
+      trackedFilters,
+      transformRuleContexts,
+      lolExtra: 'hehe not a widget param',
```
samouss: I'm not convinced it brings value to the test.

stories/QueryRuleCustomData.stories.js
```diff
@@ -0,0 +1,122 @@
...
+          </a>
+        </template>
+      </ais-query-rule-custom-data>
+    </div>
+    `,
+    data() {
+      return {
+        transformItems: items => {
+          if (items.length > 0) {
+            return [items[0]];
```
samouss: Is there a reason to only pick the first item rather than the full array?

stories/QueryRuleCustomData.stories.js
```diff
@@ -0,0 +1,122 @@
...
+        Type <q>not a banner</q> and nothing will appear.
+      </p>
+      <ais-query-rule-custom-data
+        :transform-items="transformItems"
+      />
+    </div>`,
+    data() {
+      return {
+        transformItems: items =>
+          items.filter(item => item.banner !== undefined),
```
samouss:
```suggestion
          items.filter(item => typeof item.banner !== 'undefined'),
```

stories/QueryRuleCustomData.stories.js
Haroenv: This is an example so you always only show one banner, which might be the case for most designs

stories/QueryRuleCustomData.stories.js
Haroenv: I've looked up a lot about this for the helper, and couldn't find a recent reason why to use typeof in this case, WDYT?

samouss: It's mostly a best practice to avoid runtime error but in that case it can't throw. I would have used `Boolean(...)`.

> https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined

stories/QueryRuleContext.stories.js
```diff
@@ -0,0 +1,136 @@
...
+          genre: () => ['Thriller', 'Drama'],
+        },
+      };
+    },
+  }))
+  .add('with initial filter', () => ({
+    template: `
+    <div>
+      <ul>
+        <li>On empty query & "Drama" category: "The Shawshank Redemption" appears</li>
```
francoischalifour: We can add an item to the list like:

> "Drama" is filtered by default and "The Shawshank Redemption" appears

stories/QueryRuleCustomData.stories.js
```diff
@@ -0,0 +1,124 @@
...
+      </ol>
+      `,
+    })
+  )
+  .add('default', () => ({
+    template: `
+    <div>
+      <ul
+        <li>On query <q>music</q>: "This is it" appears.</li>
+        <li>On query <q>not a banner</q>: empty banner appears.</li>
```
francoischalifour: That's confusing. Do you mean no banners appear on other queries? We could remove this `li` then.

stories/QueryRuleCustomData.stories.js
```diff
@@ -0,0 +1,124 @@
...
+      </ais-query-rule-custom-data>
+    </div>
+    `,
+  }))
+  .add('with default banner', () => ({
+    template: `
+    <div>
+      <ul>
+        <li>On query <q>music</q>: "This is it" appears.</li>
+        <li>On query <q>not a banner</q>: empty banner appears.</li>
```
francoischalifour: Same here.

stories/QueryRuleCustomData.stories.js
Haroenv: no, there's literally an empty banner that appears in the DOM, just not visibly

francoischalifour: I just saw that another Query Rule was added to the `instant_search_movies` index.

image

We shouldn't update the index like this since it's used to demo features now.

If you want to demo something else than banners, we can for instance create an ad, which is more meaningful.

Haroenv: yep, that's a good idea

src/components/__tests__/QueryRuleContext.js
```diff
@@ -0,0 +1,33 @@
+import { mount } from '@vue/test-utils';
+import QueryRuleContext from '../QueryRuleContext';
+import { __setState } from '../../mixins/widget';
+jest.mock('../../mixins/widget');
```
francoischalifour:
```suggestion

jest.mock('../../mixins/widget');
```

src/components/__tests__/QueryRuleCustomData.js
```diff
@@ -0,0 +1,44 @@
+import { mount } from '@vue/test-utils';
+import QueryRuleCustomData from '../QueryRuleCustomData.vue';
+import { __setState } from '../../mixins/widget';
+jest.mock('../../mixins/widget');
```
francoischalifour:
```suggestion

jest.mock('../../mixins/widget');
```

Haroenv:
applied your feedback @francoischalifour

src/components/__tests__/QueryRuleCustomData.js
```diff
@@ -30,6 +32,41 @@ it('renders in a list of <pre> by default', () => {
...
+  expect.assertions(items.length);
+  __setState({
+    items,
+  });
+
+  mount(QueryRuleCustomData, {
+    scopedSlots: {
+      item(props) {
+        expect(props).toEqual({
+          item: expect.objectContaining({ text: expect.any(String) }),
```
francoischalifour: I suppose this could have been done with a spy:

```js
const itemScope = jest.fn();

mount(QueryRuleCustomData, {
  scopedSlots: {
    item: itemScope,
  },
});

expect(itemScope).toHaveBeenCalledWith({ item: items[0] });
expect(itemScope).toHaveBeenCalledWith({ item: items[1] });
```

samouss:
BTW it was on purpose that we didn't implement a default slot for the `ais-query-rule-context`?

Haroenv:
Because the InstantSearch.js one also has no template I think. I couldn’t find a reason for displaying the context. This is more like configure

samouss:
Yes, but we do provide a slot for `ais-configure`. Anyway we can add one later if we have a need for it.

Issue #654, opened 2019-03-30
Multi Index Search in SSR version of InstantSearch within Nuxt.


JoaoPedroAS51:
**Feature ⚡️**

### What is your use case for such a feature?

Be able to use [Multi Index Search](https://www.algolia.com/doc/guides/building-search-ui/resources/ui-and-ux-patterns/in-depth/multi-index-search/vue/) in SSR version of InstantSearch 2 within Nuxt. It's a missing feature since 2.0 according to this https://github.com/algolia/vue-instantsearch/issues/348#issuecomment-338928514

### What is your proposal

Hi,
My proposal is use multi index with InstantSearchSsr, changing the injection `$_ais` to `$_ais-indexname`. The problem is that InstantSearchSrr component doesn't allow to set the injection name dynamically since props are not accessible inside inject, so it always inject the same index using `$_ais`.

The only solution I found was create the InstantSearchSrr component manually.

To avoid problems, I recommend add a new param ([multiIndex](https://github.com/JoaoPedroAS51/vue-instantsearch/blob/master/src/util/createInstantSearch.js#L9)) to instantSearchOptions in the file [createInstantSearch](https://github.com/JoaoPedroAS51/vue-instantsearch/blob/master/src/util/createInstantSearch.js):

```
const { searchClient, indexName, multiIndex } = instantSearchOptions;
```

Then edit the [rootMixin](https://github.com/JoaoPedroAS51/vue-instantsearch/blob/master/src/util/createInstantSearch.js#L91) to change $_ais name based in the index name, if multiIndex is set to true. That way it's only needed to create the custom component if enabled:

```
const rootMixin = {
    provide() {
      const name = multiIndex ? `$_ais-${indexName}` : '$_ais';
      return {
        // should be possible to configure this with {camelcase: ['error', {allow: ['^\\$_']}]}
        // but that didn't work
        // eslint-disable-next-line camelcase
        [name]: search,
      };
    },
  };
```

And [export](https://github.com/JoaoPedroAS51/vue-instantsearch/blob/master/src/instantsearch.js#L5) the function [createInstantSearchComponent](https://github.com/JoaoPedroAS51/vue-instantsearch/blob/master/src/util/createInstantSearchComponent.js). So users area able to create their own [InstantSearchSsr](https://github.com/JoaoPedroAS51/vue-instantsearch/blob/master/src/components/InstantSearchSsr.js) components:

```
export {
  createInstantSearchComponent,
[...]

PR #656, opened 2019-04-01
feat(server side rendering): allow injecting multiple separate InstantSearch instances

JoaoPedroAS51:
closes https://github.com/algolia/vue-instantsearch/issues/654

Hi,

I had to update the dependencies to be able to use my fork. Should I revert the **package.json** even though everything has worked after updating all the dependencies?

Thanks!


package.json
```diff
@@ -46,56 +46,57 @@
...
     "rollup-plugin-terser": "4.0.4",
-    "rollup-plugin-vue": "3.0.0",
+    "rollup-plugin-vue": "4.7.2",
     "rollup-watch": "4.3.1",
-    "vue": "2.5.18",
-    "vue-autosuggest": "1.8.1",
-    "vue-jest": "2.6.0",
+    "vue": "2.6.10",
+    "vue-autosuggest": "1.8.3",
+    "vue-jest": "3.0.4",
```
Haroenv: Most of the major changes you did here had a negative impact on bundle size; but that could have been with a different patch version. It's fairly hard to review like this, so could you make the PR with only your changes?

package.json
```diff
@@ -46,56 +46,57 @@
   },
   "dependencies": {
     "algoliasearch-helper": "^2.26.1",
-    "instantsearch.js": "^3.1.1"
+    "core-js": "^2.6.5",
```
Haroenv: This likely also isn't necessary, but could be because of babel? We didn't need it before though

src/util/createInstantSearch.js
```diff
@@ -90,11 +90,12 @@ export const createInstantSearch = instantSearchOptions => {
   // we can then reuse that InstantSearch instance seamlessly from `ais-ssr`
   const rootMixin = {
     provide() {
+      const name = multiIndex ? `$_ais-${indexName}` : '$_ais';
```
Haroenv: we could do this without the new option without breaking change

package.json
JoaoPedroAS51: Sure. Sorry for change this file. I had to do that to be able to use my fork due the error I was receiving on build the dist files. I will revert it.

package.json
JoaoPedroAS51: One of the updated dependencies was requiring the new version of it.

src/util/createInstantSearch.js
JoaoPedroAS51: I added this option because as the multi index need a custom ssr component, those who don't need it could keep using the original ssr component. If we remove this option, the original ssr component will stop working. Wouldn't that be a problem?

Haroenv: Ah yes, I see the issue; the server side rendering component can no longer read from it, since it doesn't have access to the index name. This could be solved by making that a prop again, but as you pointed out, it would become a breaking change for users who use the current version

JoaoPedroAS51: Also we can't access props inside of inject. I tried.

Haroenv: How are you configuring the component in your case then? You're overriding it in your custom ais-instant-search-ssr?

JoaoPedroAS51: Yes, as I exported the function **createInstantSearchComponent**, I'm able to create my own components and setting the injection name manually. It was the only way I found that works.

JoaoPedroAS51: Would be much better and easier with the prop, but unfortunately isn't possible.

DavidRouyer:
@Haroenv  I want to do the same thing and I can't wait a few months... Can we at least expose the `createInstantSearchComponent` method in ` src/instantsearch.js`?

Haroenv:
That can be done @DavidRouyer, but won't you want to modify more things inside it as well? As soon as you want to edit things, it makes more sense that you'd copy the file into your own code base. Or is the changes in this PR strictly enough for you?

DavidRouyer:
I tried that approach but `createInstantSearchComponent` has a reference to `createSuitMixin` which is also internal... So I want to avoid duplicating half of the codebase in my own.

I'm trying to do the same thing as @JoaoPedroAS51, it doesn't require anything else I think, I already made a custom `src/util/createInstantSearch.js`

Haroenv:
Sure that makes sense

Haroenv:
Can you fully use `createInstantSearchComponent` without changes in it though? I'm just trying to make sure this isn't exposed without modification, if it doesn't solve your issue.

You can get around the suit mixin by not importing it, it's only used for class names

PR #664, opened 2019-04-23
feat(insights): add insightsClient support on storybook

tkrugg:
**Summary**
This prepares for the insights stories to be added illustrating the usage of Hits and InfiniteHits widgets with insights.

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -1,2 +1,3 @@
 import '@storybook/addon-knobs/register';
 import '@storybook/addon-options/register';
+import '@storybook/addon-actions/register';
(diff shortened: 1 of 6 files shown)


stories/utils.js
```diff
@@ -2,6 +2,7 @@ import algoliasearch from 'algoliasearch/lite';
 
 export const previewWrapper = ({
   searchClient = algoliasearch('latency', '6be0576ff61c053d5f9a3225e2a90f76'),
+  insightsClient = null,
```
Haroenv:
```suggestion
  insightsClient,
```

.storybook/addons.js
```diff
@@ -1,2 +1,3 @@
 import '@storybook/addon-knobs/register';
 import '@storybook/addon-options/register';
+import '@storybook/addon-actions/register';
```
Haroenv: what's the reason this was already added?

tkrugg: I'll later mock insights with actions. 
It's like adding the insightsClient prop to the storybook previewWrapper while no story are using it yet: it's preparing for the next two PR (`hits with insights` and `infiniteHits with insights`) which will be using it

PR #665, opened 2019-04-23
feat(insights): add insights support to Hits widget

tkrugg:
**Summary**

This integrates insights with Vue Instantsearch to help using Click Analytics in Hits.

Using `connectHitsWithInsights` instead of `connectHits` exposed a new property `insights` to the internal state, which user can call to emit insights events.

**Result**

```html
<ais-hits>
  <div slot-scope="{ items, insights }">
    <div v-for="item in items">
      <button @click="insights('clickedObjectIDsAfterSearch', { eventName: 'Add to cart', objectIDs: [item.objectID] })">Add to cart</button>
    </div>
  </div>
</ais-hits>
```

### DIFF
--- a/src/components/Hits.vue
+++ b/src/components/Hits.vue
@@ -3,7 +3,10 @@
     v-if="state"
     :class="suit()"
   >
-    <slot :items="items">
+    <slot
+      :items="items"
+      :insights="state.insights"
+    >
       <ol :class="suit('list')">
         <li
           v-for="(item, itemIndex) in items"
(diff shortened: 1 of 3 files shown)


stories/Hits.stories.js
```diff
@@ -58,3 +59,29 @@ storiesOf('ais-hits', module)
...
+    <div>
+        <ais-configure :clickAnalytics="true" />
+        <ais-hits>
+          <div slot-scope="{ items, insights }">
+            <div
+              v-for="item in items"
+              :key="item.objectID"
+            >
+              custom objectID: {{item.objectID}}
+              <button @click="insights('clickedObjectIDsAfterSearch', { eventName: 'Add to cart', objectIDs: [item    .objectID] })">Add to cart</button>
```
Haroenv:
```suggestion
              <button @click="insights('clickedObjectIDsAfterSearch', { eventName: 'Add to cart', objectIDs: [item.objectID] })">Add to cart</button>
```

src/components/Hits.vue
```diff
@@ -3,7 +3,10 @@
     v-if="state"
     :class="suit()"
   >
-    <slot :items="items">
+    <slot
+      :items="items"
+      :insights="state.insights"
```
samouss: We could also add the function to the `item` slot. It would avoid to recreate the full UI to use insights.

tkrugg: added here: https://github.com/algolia/vue-instantsearch/commit/80446a6196168cedba999cff0121b875c94cf836

PR #666, opened 2019-04-23
feat(insights): add insights support to InfiniteHits widget

tkrugg:
⚠️ This PR depends on #664, it can't work until it's merged. It will stay as draft until then.

**Summary**
This integrates insights with Vue Instantsearch to help using Click
Analytics in InfiniteHits.

Using `connectInfiniteHitsWithInsights` instead of `connectInfiniteHits`
exposed a new property `insights` to the internal state, which user can
call to emit insights events.

**Result**
```html
<ais-infinite-hits>
  <div slot-scope="{ items, refine, insights }">

    <div v-for="item in items">
      <button @click="insights('clickedObjectIDsAfterSearch', { eventName: 'Add to cart', objectIDs: [item.objectID] })">Add to cart</button>
    </div>

    <button @click="refine()"> Load more </button>
  </div>
</ais-infinite-hits>
```

### DIFF
--- a/src/components/InfiniteHits.vue
+++ b/src/components/InfiniteHits.vue
@@ -37,6 +37,7 @@
             name="item"
             :item="item"
             :index="index"
+            :insights="state.insights"
           >objectID: {{ item.objectID }}, index: {{ index }}</slot>
         </li>
       </ol>
--- a/src/components/InfiniteHits.vue
+++ b/src/components/InfiniteHits.vue
@@ -25,6 +25,7 @@
       :refine-previous="refinePrevious"
       :refine-next="refineNext"
       :refine="refineNext"
+      :insights="state.insights"
     >
       <ol :class="suit('list')">
         <li
@@ -59,13 +60,13 @@
 
 <script>
 import { createWidgetMixin } from '../mixins/widget';
-import { connectInfiniteHits } from 'instantsearch.js/es/connectors';
+import { connectInfiniteHitsWithInsights } from 'instantsearch.js/es/connectors';
 import { createSuitMixin } from '../mixins/suit';
 
 export default {
   name: 'AisInfiniteHits',
   mixins: [
-    createWidgetMixin({ connector: connectInfiniteHits }),
+    createWidgetMixin({ connector: connectInfiniteHitsWithInsights }),
     createSuitMixin({ name: 'InfiniteHits' }),
   ],
   props: {
--- a/stories/InfiniteHits.stories.js
+++ b/stories/InfiniteHits.stories.js
@@ -133,9 +133,9 @@ storiesOf('ais-infinite-hits', module)
         action(`[InsightsClient] sent ${method} with payload`)(payload),
     })
   )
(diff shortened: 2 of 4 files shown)


Haroenv:
looks good to me

Haroenv:
this adds almost a whole kb, expected?

tkrugg:
> this adds almost a whole kb, expected?

I increased the bundlesize.
Strangely a local build tells me bundle size is 65.11 kB
```
│   Bundle Size:  245.82 KB                  │
│   Minified Size:  242.8 KB                 │
│   Gzipped Size:  65.11 KB                  │
(!) vue plugin: The ongenerate hook used by plugin vue is deprecated. The generateBundle hook should be used instead.
created dist/vue-instantsearch.js in 8.9s
```

src/components/InfiniteHits.vue
```diff
@@ -25,6 +25,7 @@
       :refine-previous="refinePrevious"
       :refine-next="refineNext"
       :refine="refineNext"
+      :insights="state.insights"
```
samouss: We could also add the function to the `item` slot. It would avoid to recreate the full UI to use insights.

stories/InfiniteHits.stories.js
```diff
@@ -124,3 +125,35 @@ storiesOf('ais-infinite-hits', module)
...
+          <div slot-scope="{ items, refine, insights }">
+            <div
+              v-for="item in items"
+              :key="item.objectID"
+            >
+              custom objectID: {{item.objectID}}
+              <button @click="insights('clickedObjectIDsAfterSearch', { eventName: 'Add to cart', objectIDs: [item.objectID] })">Add to cart</button>
+            </div>
+            <button
+              :disabled="isLastPage"
```
samouss: This variable is not declared inside the `scope`.

src/components/InfiniteHits.vue
tkrugg: Yes we can!
At first I thought maybe it would confuse some users as on why you have to pass objectIDs on the item slot scope, but that's probably unfounded.

```ts
<ais-infinite-hits>
  <div slot-scope="{ item, refine, insights }">
      <button @click="insights('clickedObjectIDsAfterSearch', { eventName: 'Add to cart', objectIDs: [item.objectID] })">Add to cart</button>
  </div>
</ais-infinite-hits>
```

stories/InfiniteHits.stories.js
tkrugg: fixed here 3132a34

src/components/InfiniteHits.vue
samouss: Ahh I get the point, I don't think we matter too much in that case. The signature remains the same between the slots that's what is important. Worst case if we have users complain we could pre-apply the function with the `objectID`.

tkrugg: added here: b03c9c032ed1af5375d164b5e957bfec90913690

PR #670, opened 2019-04-30
feat(panel): provide noRefinement scope to slots

### LINKED ISSUE #347: Displaying some sort of placeholder instead of hiding component when there are no facets

Feature

**Feature: What is your use case for such a feature?**

When a component has no facets or results then they get hidden, but in my current implementation this makes it look like the interface broke, see screenshots for an example:

*With results (ok)*
image

*Without results (looks broken)*
image

I can't see any way to still display something when there are no results.

**Feature: What is your proposed API entry? The new option to add? What is the behavior?**

Add a `placeholder` slot or something comparable to still show some sort of message or button when the component would be hidden.

Latest

Haroenv:
This allows you to create interface like:

```vue
<ais-panel>
  <template slot="header">Hi</template>
  <template slot="default" slot-scope="{ noRefinement }">
    <p v-if="noRefinement">no results</p>
  </template>
</ais-panel>
```

sandbox with this: https://codesandbox.io/s/znw7ry52wl (which now obviously doesn't work).

fixes #347

### DIFF
--- a/src/components/Panel.vue
+++ b/src/components/Panel.vue
@@ -6,11 +6,11 @@
     >
       <slot
         name="header"
-        :no-refinement="!canRefine"
+        :has-refinements="canRefine"
       />
     </div>
     <div :class="suit('body')">
-      <slot :no-refinement="!canRefine" />
+      <slot :has-refinements="canRefine" />
     </div>
     <div
       v-if="$slots.footer || $scopedSlots.footer"
@@ -18,7 +18,7 @@
     >
       <slot
         name="footer"
-        :no-refinement="!canRefine"
+        :has-refinements="canRefine"
       />
     </div>
   </div>
--- a/src/components/Panel.vue
+++ b/src/components/Panel.vue
@@ -1,7 +1,7 @@
 <template>
   <div :class="[suit(), !canRefine && suit('', 'noRefinement')]">
     <div
-      v-if="$slots.header"
+      v-if="$slots.header || $scopedSlots.header"
       :class="suit('header')"
     >
       <slot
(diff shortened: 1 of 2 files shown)


src/components/__tests__/Panel.js
```diff
@@ -30,6 +30,21 @@ describe('default render', () => {
...
+      scopedSlots: {
+        default: defaultScopedSlot,
+      },
+    });
+
+    wrapper.setData({
+      canRefine: false,
+    });
+
+    expect(defaultScopedSlot).toHaveBeenCalledWith({ noRefinement: true });
```
francoischalifour: Would it make sense to test that it's been called once?

src/components/__tests__/Panel.js
```diff
@@ -30,6 +30,21 @@ describe('default render', () => {
     expect(wrapper.html()).toMatchSnapshot();
   });
 
+  it('passes data without refinement', () => {
```
francoischalifour: We can add another test where there are refinements and `noRefinement` is `false`.

src/components/Panel.vue
```diff
@@ -4,16 +4,22 @@
       v-if="$slots.header"
       :class="suit('header')"
     >
-      <slot name="header" />
+      <slot
+        name="header"
+        no-refinement="!canRefine"
```
francoischalifour: I think the proper English way is "no refinements". But this is a mistake that we made in InstantSearch.css.

src/components/__tests__/Panel.js
Haroenv: I don't think it's relevant, we can call the slot multiple times but it wouldn't change anything for users

francoischalifour:
Why not give `canRefine` instead of `noRefinement`?

Haroenv:
That way it's consistent with the CSS class

PR #679, opened 2019-05-31
fix(highlight): change expected tag to replace

### LINKED ISSUE #639: v2 – Highlight component ignores some props
**Bug 🐞**

The `ais-highlight` component does not pick up the `highlighted-tag-name` property and does not apply the CSS classes correctly.

[Sandbox →](https://codesandbox.io/s/jj23oy6919)

Based on this sandbox, the tag name should be `em` and should have the CSS class `ais-Highlight-highlighted` appended.

InstantSearch.js 3 [highlights with the `mark` tag by default](https://github.com/algolia/instantsearch.js/blob/2bb5203018774c2ec920a9ac1be4d828856f2bd8/src/helpers/highlight.js), while Vue InstantSearch [expects an `em` tag to transform](https://github.com/algolia/vue-instantsearch/blob/79d0573dc0b8b9685beecc114656288b7d03d6c9/src/components/Highlight.vue#L48).

Updating this matching with the `mark` tag should solve the problem.

Vue InstantSearch 2.0.0

Haroenv:
fixes #639 

to fully fix #639, we also want algolia/instantsearch.js#3830 to be merged, however those are completely separate issues

### DIFF
--- a/src/components/Highlight.vue
+++ b/src/components/Highlight.vue
@@ -10,6 +10,13 @@ import { createSuitMixin } from '../mixins/suit';
 import { getPropertyByPath } from '../util/object';
 import { warn } from '../util/warn';
 
+// not imported from instantsearch.js/es/lib/escape-highlight,
+// because that path can change adn break our implementation
+const TAG_REPLACEMENT = {
+  highlightPreTag: '<mark>',
+  highlightPostTag: '</mark>',
+};
+
 export default {
   name: 'AisHighlight',
   mixins: [createSuitMixin({ name: 'Highlight' })],
@@ -43,12 +50,19 @@ export default {
             'See https://www.algolia.com/doc/api-reference/api-parameters/attributesToHighlight/.'
         );
       }
-      return attributeValue
+
+      const final = attributeValue
         .replace(
-          new RegExp('<em>', 'g'),
+          new RegExp(TAG_REPLACEMENT.highlightPreTag, 'g'),
           `<${this.highlightedTagName} class="${this.suit('highlighted')}">`
         )
-        .replace(new RegExp('</em>', 'g'), `</${this.highlightedTagName}>`);
+        .replace(
+          new RegExp(TAG_REPLACEMENT.highlightPostTag, 'g'),
+          `</${this.highlightedTagName}>`
+        );
+
+      console.warn(attributeValue);
+      return final;
     },
   },
 };
--- a/src/components/Highlight.vue
+++ b/src/components/Highlight.vue
@@ -51,7 +51,7 @@ export default {
         );
       }
 
-      const final = attributeValue
+      return attributeValue
         .replace(
           new RegExp(TAG_REPLACEMENT.highlightPreTag, 'g'),
           `<${this.highlightedTagName} class="${this.suit('highlighted')}">`
@@ -60,8 +60,6 @@ export default {
           new RegExp(TAG_REPLACEMENT.highlightPostTag, 'g'),
           `</${this.highlightedTagName}>`
         );
-
-      return final;
     },
   },
 };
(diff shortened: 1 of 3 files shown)


src/components/Highlight.vue
```diff
@@ -43,12 +50,18 @@ export default {
             'See https://www.algolia.com/doc/api-reference/api-parameters/attributesToHighlight/.'
         );
       }
-      return attributeValue
+
+      const final = attributeValue
```
Haroenv:
```suggestion
      return attributeValue
```

src/components/Highlight.vue
```diff
@@ -43,12 +50,18 @@ export default {
...
+          new RegExp(TAG_REPLACEMENT.highlightPreTag, 'g'),
           `<${this.highlightedTagName} class="${this.suit('highlighted')}">`
         )
-        .replace(new RegExp('</em>', 'g'), `</${this.highlightedTagName}>`);
+        .replace(
+          new RegExp(TAG_REPLACEMENT.highlightPostTag, 'g'),
+          `</${this.highlightedTagName}>`
+        );
+
+      return final;
```
Haroenv:
```suggestion
```

src/components/Highlight.vue
```diff
@@ -43,12 +50,18 @@ export default {
...
-          new RegExp('<em>', 'g'),
+          new RegExp(TAG_REPLACEMENT.highlightPreTag, 'g'),
           `<${this.highlightedTagName} class="${this.suit('highlighted')}">`
         )
-        .replace(new RegExp('</em>', 'g'), `</${this.highlightedTagName}>`);
+        .replace(
+          new RegExp(TAG_REPLACEMENT.highlightPostTag, 'g'),
+          `</${this.highlightedTagName}>`
+        );
+
```
Haroenv:
```suggestion
```

src/components/Highlight.vue
```diff
@@ -10,6 +10,13 @@ import { createSuitMixin } from '../mixins/suit';
 import { getPropertyByPath } from '../util/object';
 import { warn } from '../util/warn';
 
+// not imported from instantsearch.js/es/lib/escape-highlight,
+// because that path can change adn break our implementation
+const TAG_REPLACEMENT = {
```
Haroenv: @francoischalifour WDYT about this duplication instead of import?

Haroenv:
I checked it, and this change also fixes it for searchable

src/components/Highlight.vue
francoischalifour: We can inline `mark` where we used `em` because we don't use this anywhere else.

francoischalifour:
[The stories](https://deploy-preview-679--vue-instantsearch.netlify.com/stories/?selectedKind=ais-hits&selectedStory=simple%20usage&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs) don't seem to work, they display `<mark>Text</mark>`.

Haroenv:
That’s because of InstantSearch’s PR not yet been used here (there’s two hits widgets). You can check a story with only one hits to see it working compared to previously.

Haroenv:
That’ll only work with escapeHTML: false, which is why I didn’t go for it

samouss:
You're sure? I've tested it without issues. We use the same strategy inside InstantSearch.js.

Haroenv:
It's possible that it works, but in the end it would be best if we migrate the InstantSearch highlight strategy to the one of React InstantSearch where there's no need for escaping or innerHTML, but you can map over the parts

benjam-es:
highlighted-tag-name still does not work for me. Is the above fix tagged?

I'm on  "vue-instantsearch": "^2.2.1"

francoischalifour:
@benjam-es The issue is fully fixed since #691 which updates the InstantSearch.js dependency. Vue InstantSearch wasn't released since. The `highlighted-tag-name` behavior will work as expected in the next version.

PR #687, opened 2019-06-21
chore(examples): implement e-commerce demo mobile design

eunjae-lee:
## Description
This implements the mobile design of the e-commerce demo. Based on https://github.com/algolia/instantsearch.js/pull/3866

## Notes
- Other than the changes for mobile design in InstantSearch.js, there are [a few changes](https://github.com/algolia/vue-instantsearch/pull/687/commits/48f76ef3329754018f9f852f982b9f60a25ccb49#diff-fdd071e87b0b77e2d913c02777ea0cf2) in style dedicated to Vue InstantSearch.
- The implementation for opening/close filters from the vanilla(`ui.js`) is applied directly to `App` component.

## Preview
[See preview](https://deploy-preview-687--vue-instantsearch.netlify.com/examples/ecommerce/index.html)


Haroenv:
Screenshot 2019-06-24 at 11 00 25

you can't scroll fully to the bottom on mobile

Haroenv:
I'd like it if we can make sure there's no moving around of text on selecting something on mobile, by using regular heights

eunjae-lee:
@Haroenv 

> I'd like it if we can make sure there's no moving around of text on selecting something on mobile, by using regular heights

Can you give me an example?

Haroenv:
compare kitchenaid which jumps because there's only one level on the hierarchical menu, vs samsung which doesn't have the issue because of the length which doesn't change

eunjae-lee:
@Haroenv by `regular heights`, do you mean something like this?
I'm not sure what's the best solution to this.

Kapture 2019-06-24 at 17 43 40

Haroenv:
I guess it's a  problem we can leave for now. A possible solution is to show the other levels greyed out, but we can't do that yet.

eunjae-lee:
Okay, when we have some improvements on vanilla, let's propagate them to the flavors.

eunjae-lee:
Have this PR been approved already? I don't see any sign it's approved but I somehow see green "squash and merge" button. why...?

examples/ecommerce/src/App.mobile.css
```diff
@@ -0,0 +1,261 @@
+@media (max-width: 899px) {
```
Haroenv: this can be done with a media query on the `link` too I think

eunjae-lee: You're saying something like this, right?
```
  <link rel="stylesheet" type="text/css" href="App.mobile.css" media="max-width: 899px">
```

I don't know how to do that from `import './App.mobile.css';`. This should be easier on vanilla where they directly use `link` tag.

examples/ecommerce/src/App.vue
```diff
@@ -78,14 +71,29 @@
                 </div>
               </template>
             </ais-clear-refinements>
+
+            <ais-stats data-layout="mobile">
```
francoischalifour: The text size is smaller than on the design on this widget.

examples/ecommerce/src/App.vue
```diff
@@ -210,6 +180,32 @@
               </template>
             </ais-panel>
           </div>
+
+          <footer class="container-filters-footer" data-layout="mobile">
+            <ais-clear-refinements class="container-filters-footer-button-wrapper">
```
francoischalifour: This button should:

- [x] Reset the filters
- [ ] Close the overlay
- [ ] Have the same design when not disabled than when disabled (because it's still clickable to close the overlay)
- [ ] Be renamed to "Reset filters"
- [ ] Not contain the image

examples/ecommerce/src/App.vue
```diff
@@ -97,77 +105,41 @@
...
               </template>
             </ais-panel>
 
-            <ais-panel
-              :class-names="{
-                'ais-Panel-header': 'panel-header',
-              }"
-            >
+            <ais-panel>
               <template slot="header">Rating</template>
```
francoischalifour:
```suggestion
               <template slot="header">Ratings</template>
```

This mistake appeared before this PR.

examples/ecommerce/src/App.vue
```diff
@@ -461,6 +475,17 @@ export default {
...
       ];
     },
+    openFilters() {
+      document.body.classList.add('filtering');
+      window.scrollTo(0, 0);
+      window.addEventListener('keyup', this.onKeyUp);
+    },
+    closeFilters() {
+      document.body.classList.remove('filtering');
+      const resultsContainer = document.querySelector('.container-results');
```
francoischalifour: This doesn't need to be computed every time, right? It could be stored outside of the function.

eunjae-lee: 4f042f9

examples/ecommerce/src/App.vue
```diff
@@ -482,8 +482,10 @@ export default {
     },
     closeFilters() {
       document.body.classList.remove('filtering');
-      const resultsContainer = document.querySelector('.container-results');
-      resultsContainer.scrollIntoView();
+      if (!this.resultsContainer) {
```
francoischalifour: I meant storing it in the `script` tag after the imports.

eunjae-lee: It won't successfully query the element until mounted. We can move it to `mounted` lifecycle method. What do you think?

francoischalifour: Ah right – I'm not proficient with Vue so do as seems good to you!

examples/ecommerce/src/App.vue
eunjae-lee: 6f1ed6c

examples/ecommerce/src/App.vue
eunjae-lee: a4e3e68

examples/ecommerce/src/App.vue
eunjae-lee: 5d72727

eunjae-lee:
> The stars in RatingMenu should be 24px size on mobile

d646e43

> The RangeSlider handles should be 24px size on mobile

1360272

eunjae-lee:
I've brought all the recent changes from InstantSearch.js.

examples/ecommerce/src/App.mobile.css
```diff
@@ -170,6 +171,7 @@
   .ais-RefinementList-checkbox {
     height: 1.5rem;
     width: 1.5rem;
```
francoischalifour: `width` is not required with `min-width`.

eunjae-lee: Right I was supposed to remove `width`. I'm doing the same to angular-IS. c1e37b2

francoischalifour:
@tkrugg I think we're unable to format the counts formatting in RatingMenu and HierarchicalMenu without connectors because there's no concept of formatting numbers/counts in Vue InstantSearch:

https://github.com/algolia/vue-instantsearch/blob/c023ba0fafbb32d556db2895aafec6c3742a1380/src/components/RatingMenu.vue#L71

That's something worth keeping in mind!

examples/ecommerce/src/widgets/PriceSlider.css
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.vue-slider-dot-tooltip-text .vue-slider-dot-tooltip-inner {
+  cursor: -webkit-grabbing;
+  cursor: -moz-grabbing;
+  text-align: center;
+  white-space: nowrap;
+}
+
+.vue-slider-dot-handle {
```
francoischalifour: The style for the handles on mobile are missing, they're supposed to be bigger.

eunjae-lee: Right, it was missing and now it's back. 001680d

PR #696, opened 2019-07-16
feat(RefinementList): forward class names to input

### LINKED ISSUE #644: Can't apply a class name to ais-SearchBox-input within ais-refinement-list
**Bug 🐞**

It is impossible to apply a class-name to `ais-SearchBox-input` within `ais-refinement-list`

https://4j2k883yp4.codesandbox.io/
If you inspect the ais-RefinementList-count, a "foo" class is applied. However, if do the same thing on ais-SearchBox-input, the "bar" class is not applied. 

We should be able to apply a style to the `ais-SearchBox-input` class by using the following code : 

```
<ais-refinement-list
attribute="label"
searchable
:class-names="{
	'ais-SearchBox-input': 'test',
}" />
```

It happens all the time.

2.0.0

Haroenv:
fixes #644

When we forward the class names from the refinement list itself, you can override the classes like this:

```vue
<template>
  <ais-refinement-list
    attribute="category"
    searchable
    :class-names="classNames"
  />
</template>

<script>
export default {
  data() {
    return {
      classNames: {
        'ais-RefinementList-list': 'refinement-list',
        'ais-SearchBox-input': 'refinement-input',
      },
    };
  },
};
</script>
```

### DIFF
--- a/src/components/RefinementList.vue
+++ b/src/components/RefinementList.vue
@@ -25,6 +25,7 @@
         <search-input
           v-model="searchForFacetValues"
           :placeholder="searchablePlaceholder"
+          :class-names="classNames"
         />
       </div>
       <slot
--- a/src/components/__tests__/RefinementList.js
+++ b/src/components/__tests__/RefinementList.js
@@ -103,6 +103,26 @@ it("allows override of placeholder when it's searchable", () => {
   );
 });
 
+it("allows search bar classes override when it's searchable", () => {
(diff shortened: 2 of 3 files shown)


Haroenv:
https://5d2dc49d71ebf9000bae6985--vue-instantsearch.netlify.com/ works (but there's no story showing this behaviour since it doesn't really warrant it)

PR #698, opened 2019-07-22
fix(ssr): clearer error message when using ssr without the component

### LINKED ISSUE #697: this.instantSearchInstance.__forceRender is not a function using Nuxt
**Bug 🐞**

Skærmbillede 2019-07-21 kl  12 52 58

``` vue
<template>
  <ais-instant-search :search-client="searchClient" index-name="instant_search">
    <ais-search-box />
    <ais-stats />
    <ais-hits>
      <template
        slot="item"
        slot-scope="{ item }"
      >
        <p>
          <ais-highlight
            attribute="name"
            :hit="item"
          />
        </p>
        <p>
          <ais-highlight
            attribute="brand"
            :hit="item"
          />
        </p>
      </template>
    </ais-hits>
    <ais-pagination />
  </ais-instant-search>
</template>

<script>
  import {
    AisInstantSearch,
    AisRefinementList,
    AisHits,
    AisHighlight,
    AisSearchBox,
    AisStats,
    AisPagination,
    createInstantSearch
  } from 'vue-instantsearch'
  import algoliasearch from 'algoliasearch/lite'

  const searchClient = algoliasearch(
    'latency',
    '6783af546198179de4b49d578020a2a6'
  )

  const { instantsearch, rootMixin } = createInstantSearch({
    searchClient,
    indexName: 'products'
  })

  export default {
    data () {
      return {
        searchClient
      }
    },
    components: {
      AisInstantSearch,
      AisRefinementList,
      AisHits,
      AisHighlight,
      AisSearchBox,
      AisStats,
      AisPagination
    },
    asyncData() {
      return instantsearch
        .findResultsState({
          // find out which parameters to use here using ais-state-results
          query: 'iphone',
          hitsPerPage: 5,
          disjunctiveFacets: ['brand'],
          disjunctiveFacetsRefinements: { brand: ['Apple'] }
        })
        .then(() => ({
          instantSearchState: instantsearch.getState()
        }));
    },
    beforeMount() {
      instantsearch.hydrate(this.instantSearchState);
    },
    mixins: [rootMixin],
    head() {
      return {
        link: [
          {
            rel: 'stylesheet',
            href:
              'https://cdn.jsdelivr.net/npm/instantsearch.css@7.3.1/themes/algolia-min.css'
          }
        ]
      };
    }
  }
</script>
```

Error "TypeError: this.instantSearchInstance.__forceRender is not a function" is thrown when using Nuxt.

No errors?

"nuxt": "^2.0.0",
"vue-instantsearch": "^2.2.2"

Haroenv:
fixes #697

We are patching the InstantSearch instance in SSR, but you could be using the code for SSR without using the patched InstantSearch instance.

The error being thrown then is confusing, since it just complains about a missing property.

We have no differential error management with short & long error messages in prod/dev here yet, so I tried to be between both (not too long), this still has a .03kb difference in the bundle, so I updated the limits

### DIFF
--- a/src/mixins/widget.js
+++ b/src/mixins/widget.js
@@ -25,6 +25,11 @@ export const createWidgetMixin = ({ connector } = {}) => ({
 
       const { hydrated, started } = this.instantSearchInstance;
       if ((!started && hydrated) || this.$isServer) {
+        if (typeof this.instantSearchInstance.__forceRender !== 'function') {
+          throw new Error(
+            'You are using server side rendering with <ais-instant-search> instead of <ais-instant-search-server>.'
+          );
+        }
         this.instantSearchInstance.__forceRender(this.widget);
       }
     } else if (connector !== true) {


Haroenv:
This makes the size .03 kb bigger, but I see no way around that

src/mixins/widget.js
```diff
@@ -25,6 +25,11 @@ export const createWidgetMixin = ({ connector } = {}) => ({
 
       const { hydrated, started } = this.instantSearchInstance;
       if ((!started && hydrated) || this.$isServer) {
+        if (typeof this.instantSearchInstance.__forceRender !== 'function') {
+          throw new Error(
+            'You are using server side rendering with <ais-instant-search> instead of <ais-instant-search-server>.'
```
samouss: The name of the widget is incorrect, it should be `ais-instant-search-ssr`.

https://github.com/algolia/vue-instantsearch/blob/20c2153983f45cf69834bde5f7ab474ac1db932b/src/components/InstantSearchSsr.js#L4

Haroenv: fixed in 44f2d9c

PR #711, opened 2019-08-28
feat: Adds events to hook on to focus, blur, and reset

jschatz1:
At Netlify we needed these events to make changes to the rest of the site.


Haroenv:
I have published this as 2.4.0 :)

PR #721, opened 2019-09-17
feat: Expose `createSuitMixin` for easy class reuse in custom components

kevintruby:
While using the `createWidgetMixin`, there is no easy way to reproduce all the DOM classes typically generated for widget components; the mixin used internally is not exposed for consumption in production environments.

Would like to be able to leverage the same pattern when wrapping additional template around existing components.

### DIFF
--- a/src/instantsearch.js
+++ b/src/instantsearch.js
@@ -1,3 +1,4 @@
+export { createSuitMixin } from './mixins/suit';
 export { createWidgetMixin } from './mixins/widget';
 export * from './widgets';
 export { plugin as default } from './plugin';


Haroenv:
we have previously not exposed this to users since it's fairly trivial in implementation; as well as assuming people wouldn't want to use it

Haroenv:
Released in 2.5.0

kevintruby:
Thank you for the review, and for integrating/releasing this change!

I agree, the implementation can be fairly trivial (_I just defined a `baseClass` and then used computed properties to extend it for `-list`, `-item`, etc_), but exposing the suit helper seemed useful and harmless; figured it couldn't hurt to see if the Algolia team would approve or reject it 😄 

I'm really looking forward to updating our version of the library to use this 🙏 🙌

Issue #725, opened 2019-09-25
tree shaking is not working correctly


Haroenv:
**Bug 🐞**

### What is the current behavior?

Importing any component with the "recommended" way will end up with a bundle which includes the other connectors & components

### Make a sandbox with the current behavior

https://codesandbox.io/s/vue-instantsearch-app-vkq46

### What is the expected behavior?

bundle gets tree shaked

### Does this happen only in specific situations?

seems always 

### What is the proposed solution?

I tried:

1. sideEffects: false -> doesn't seem to do anything
2. changing `files` to include `src`, and importing directly -> this works

### What is the version you are using?

2.3.0

> Always try the latest one before opening an issue.

Haroenv:
see also this discourse post: https://discourse.algolia.com/t/webpack-tree-shaking-not-working-w-vue-instantsearch/8579/2

Haroenv:
I believe this has been fixed by #866

PR #726, opened 2019-09-26
feat(core): expose source files

Haroenv:
This should allow people who are using Webpack, which doesn't correctly work with "non-fully-used" single files to still import only what's needed.

The size of tree-shaking vue-instantsearch isn't too different from the whole bundle, because the biggest parts are the InstantSearch class & the algoliasearch-helper. Luckily this will be better in the next major version.

see #725 (fixes part of it as it gives a workaround, but ideally it wouldn't be necessary)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
   "license": "MIT",
   "version": "2.5.0",
   "files": [
-    "dist"
+    "dist",
+    "src"
   ],
   "main": "dist/vue-instantsearch.common.js",
   "module": "dist/vue-instantsearch.esm.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -138,5 +139,6 @@
       "jest-watch-typeahead/filename",
       "jest-watch-typeahead/testname"
     ]
-  }
+  },
+  "sideEffects": false
```
francoischalifour: Perhaps we can move that line above the scripts so that it's not buried here.

Haroenv: This way it’s alphabetical, and it’s usually last. Not sure what’s best

francoischalifour: Ah, didn't realize it was alphabetical. All good then!

PR #730, opened 2019-10-04
refactor: rely on InstantSearch highlighting helpers

samouss:
This PR updates the implementation of `ais-hightlight` & `ais-snippet` to leverage the helpers provided by InstantSearch.js. The implementation of those helpers is exactly what we have inside Vue InstantSearch. It doesn't work with `escape-HTML="false"` but it never did. It's not a regression in that sense.

We should update how the highlighting is implemented inside InstantSearch.js to have something more flexible (maybe like [React InstantSearch](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/core/highlight.js) does). At least once we update the InstantSearch.js implementation Vue will benefit from it.


src/components/Highlight.vue
```diff
@@ -6,9 +6,8 @@
 </template>
 
 <script>
+import instantsearch from 'instantsearch.js/es';
 import { createSuitMixin } from '../mixins/suit';
-import { getPropertyByPath } from '../util/object';
```
Haroenv: this helper can be removed now i think

src/components/Highlight.vue
```diff
@@ -29,27 +28,11 @@ export default {
...
   computed: {
     innerHTML() {
-      const attributeValue =
-        getPropertyByPath(
-          this.hit,
-          `_highlightResult.${this.attribute}.value`
-        ) || '';
-
-      if (process.env.NODE_ENV !== 'production' && attributeValue === '') {
-        warn(
```
samouss: We don't have the warning anymore, but it shouldn't be the responsibility of Vue InstantSearch. If we want to warn it should be done inside InstantSearch.js. I don't think it has a huge impact to remove it from Vue at the moment.

src/components/Highlight.vue
samouss: https://github.com/algolia/vue-instantsearch/pull/730/commits/b37ff57422dc9b091a56cf96f67f11455a53fcb6

PR #732, opened 2019-10-08
chore(examples): update examples to match the ones on instantsearch.js

yannickcr:
**Summary**

This PR updates the examples to match the ones from [`instantsearch.js`](https://github.com/algolia/instantsearch.js) to allow us to run the same [End-2-End test suite](https://github.com/algolia/instantsearch-e2e-tests) on them.

The changes are:

* Moving e-commerce example from `examples/ecommerce` to `examples/e-commerce`.
* Examples and Storybook that were compiled in`docs/dist` are now compiled in `website`.
* Fix e-commerce demo for IE11 (polyfills).
* Update e-commerce routing configuration to match the one from `instantsearch.js` (most of it is just copy/pasted from https://github.com/algolia/instantsearch.js/blob/17c70e9520349e9a2f5985dbe1320b14dd6109ef/examples/e-commerce/src/routing.ts).
* Update e-commerce example to match the Vanilla and Reac ones: missing aria attributes, different labels, `hitsPerPage` always set to `16`.
* Update e-commerce example to use `vue-instantsearch` build from the current branch.


examples/e-commerce/yarn.lock
```diff
@@ -1234,6 +1234,15 @@ algoliasearch-helper@^2.26.0, algoliasearch-helper@^2.26.1:
     qs "^6.5.1"
     util "^0.10.3"
 
+algoliasearch-helper@^2.26.1:
```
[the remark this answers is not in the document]
yannickcr: Should be an effect of my multiple tests.
I regenerate the `yarn.lock` and it should be correct now: https://github.com/algolia/vue-instantsearch/pull/732/commits/a2aaef39263ebb087a8dc415be23c362115ec6be

examples/e-commerce/yarn.lock
```diff
@@ -4412,6 +4412,21 @@ instantsearch.js@3.5.3, instantsearch.js@^3.5.1:
     prop-types "^15.5.10"
     qs "^6.5.1"
 
+instantsearch.js@^3.6.0:
```
Haroenv: same issue as with the helper earlier, only one version should be here

examples/e-commerce/src/App.vue
```diff
@@ -517,6 +519,10 @@ export default {
         value.max !== null ? value.max : range.max,
       ];
     },
+    isSelectedHitsPerPageValue(itemValue, force = false) {
+      const [, hitsPerPage] = document.location.search.match(/hitsPerPage=([0-9]+)/) || [];
```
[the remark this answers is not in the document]
yannickcr: Prettier is not set in the ESLint configuration on `vue-instantsearch`, we should propably fix that.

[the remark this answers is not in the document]
Haroenv: this was not done originally  because the eslint tooling for vue (vetur) kept reverting prettier's changes in templates. We can do this another time.

examples/e-commerce/yarn.lock
yannickcr: I wpdated instantsearch.js version in the example to avoid this https://github.com/algolia/vue-instantsearch/pull/732/commits/5805fa536fb59017b2ae88fb6e1d11c160f2b922

examples/e-commerce/src/App.vue
```diff
@@ -135,6 +135,9 @@
                         :duration="0"
                         @change="refine({ min: $event[0], max: $event[1] })"
                       >
+                        <template slot="dot" slot-scope="{ index, value }">
+                          <div :aria-valuemin="range.min" :aria-valuemax="range.max" :aria-valuenow="value" :data-handle-key="index" class="vue-slider-dot-handle" role="slider" tabindex="0" />
```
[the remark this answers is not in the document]
yannickcr: not sure to understand what you mean :thinking:

examples/e-commerce/package.json
```diff
@@ -10,9 +10,9 @@
   "dependencies": {
     "algoliasearch": "3.33.0",
     "classnames": "2.2.6",
-    "instantsearch.js": "3.5.3",
+    "instantsearch.js": "^3.6.0",
```
francoischalifour: We can pin this to 3.7.0 now (likely the last v3 version).

examples/e-commerce/src/App.vue
```diff
@@ -517,6 +519,10 @@ export default {
         value.max !== null ? value.max : range.max,
       ];
     },
+    isSelectedHitsPerPageValue(itemValue, force = false) {
```
francoischalifour: Is the `force` argument making the value as default? If so, it seems like this function is addressing several purposes.

examples/e-commerce/package.json
yannickcr: I put the exact same value as `vue-instantsearch` to avoid any duplicate in the lockfile.

examples/e-commerce/src/App.vue
```diff
@@ -517,6 +519,10 @@ export default {
         value.max !== null ? value.max : range.max,
       ];
     },
+    getSelectedHitsPerPageValue() {
+      const [, hitsPerPage] = document.location.search.match(/hitsPerPage=([0-9]+)/) || [];
```
Haroenv: are we sure we want to solve this like this? It seems quite messy to me to rely on the URL here

yannickcr: Why is it messy? Do you have a better solution in mind?

Haroenv: Widgets themselves shouldn't be taking the URL into account, since that's handled by routing. How is this solved in the other flavors?

yannickcr: The issue is that the `HitsPerPage` widget [only uses the `default` property on `items` to find the selected one](https://github.com/algolia/vue-instantsearch/blob/ede58524399dce43f6107d4e8d6be98417f29bbd/src/components/HitsPerPage.vue#L57). It does not take the current state into account ( like [`react-instantsearch` for example](https://github.com/algolia/react-instantsearch/blob/4906170382232d9c4eed3a5723f544e63eff9e0a/packages/react-instantsearch-core/src/connectors/connectHitsPerPage.js#L13-L27) ). Which forces us to retrieve this value by ourselves.

Haroenv: I think in this case the test should be failing instead of passing with this workaround. The behaviour is actually wrong.

Haroenv: we can skip the test for vue here?

yannickcr: I'd rather keep this small workaround for the demo than removing a test for Vue then working on a fix for `HitsPerPage` (should be something quick to fix).

Haroenv: if that's what you prefer

PR #745, opened 2019-11-07
feat(index): add multi-index widget

Haroenv:
This new widget `<ais-index>` allows you to search through multiple indices at once, without manually needing to keep parameters in sync. 

BREAKING CHANGE: this introduces the dependency on InstantSearch v4, which has its own breaking changes, you might see if you're using methods on `helper`. See more here: https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#upgrade-from-v3-to-v4


src/components/Index.js
```diff
@@ -0,0 +1,51 @@
+import { createSuitMixin } from '../mixins/suit';
+import { createWidgetMixin } from '../mixins/widget';
+// TODO: should InstantSearch expose a fake connector instead? Would be more consistent.
+import indexWidget from 'instantsearch.js/es/widgets/index/index';
+
+// wrapped in a dummy function, since indexWidget doesn't render
+const connectIndex = () => indexWidget;
```
Haroenv: This is fine, but maybe it should live in InstantSearch instead for consistency?

src/mixins/widget.js
```diff
@@ -3,14 +3,22 @@ import { warn } from '../util/warn';
...
+      // from: '$ais_instantSearchInstance',
       default() {
         const tag = this.$options._componentTag;
         throw new TypeError(
           `It looks like you forgot to wrap your Algolia search component "<${tag}>" inside of an "<ais-instant-search>" component.`
         );
       },
     },
+    indexWidget: {
+      // from: '$ais_indexWidget',
```
Haroenv: can we rename it? I prefer to use `indexWidget` inside the mixin, but `$ais_indexWidget` as real provided value, but the from didn't seem to work in my testing. cc @posva is that expected?

src/mixins/widget.js
```diff
@@ -21,7 +29,7 @@ export const createWidgetMixin = ({ connector } = {}) => ({
     if (typeof connector === 'function') {
       this.factory = connector(this.updateState, () => {});
       this.widget = this.factory(this.widgetParams);
-      this.instantSearchInstance.addWidget(this.widget);
+      this.root.addWidgets([this.widget]);
```
Haroenv: addWidget -> addWidgets to prevent deprecation warning showing in console

stories/Autocomplete.stories.js
```diff
@@ -59,18 +59,9 @@ storiesOf('ais-autocomplete', module)
   .add('using vue-autosuggest (multi-index)', () => ({
     template: `
       <div>
-        <ais-autocomplete
-          :indices="[
```
Haroenv: autocomplete widget needs to remove this prop

src/mixins/widget.js
posva: it should work, yeah. If you reproduce it on a jsfiddle, open an issue on vue. It could also be a bug on the test utils not working (not sure if that's what you meant by testing)

Haroenv: I can no longer reproduce this, don't know why it initially didn't work

package.json
```diff
@@ -50,8 +50,8 @@
     "release:publish": "scripts/release-publish.sh"
   },
   "dependencies": {
-    "algoliasearch-helper": "^2.26.1",
-    "instantsearch.js": "^3.6.0"
+    "algoliasearch-helper": "0.0.0-5a0352a",
```
francoischalifour: This can be using v3 now.

src/components/Index.js
francoischalifour: Yes – this will become more obvious once we decouple InstantSearch.js from the connectors.

src/components/Index.js
```diff
@@ -0,0 +1,50 @@
...
+      required: true,
+    },
+    indexId: {
+      type: String,
+      required: false,
+    },
+  },
+  render(createElement) {
+    return createElement(
+      'div',
```
francoischalifour: One notable difference with InstantSearch.js and React InstantSearch is that the `Index` widget is not render-less in Vue InstantSearch.

Is there a way to not render anything in Vue?

src/components/Index.js
```diff
@@ -0,0 +1,50 @@
...
+      {
+        class: {
```
francoischalifour: An index doesn't have class names in other flavors. I don't think we should move away from the other flavor implementations by adding more rendering differences.

package.json
Haroenv: InstantSearch hasn't yet been released, so was waiting for that

src/components/Index.js
Haroenv: yes, you can simply return the children. I can change that

Haroenv: I've been looking, and it's not possible, because Vue only allows one child. What we can do instead is not rendering anything if there's no children.

src/components/Index.js
Haroenv: In React InstantSearch it used to have a class. Since we are forced to have something here I don't think it's a big deal to add the class

src/components/Index.js
francoischalifour: Worst case, having a plain `div` (without class names) is fine for now.

src/components/Index.js
francoischalifour: It broadens the API surface in the end. Users mights rely on this class while they don't need it. If we add it now, we'll have trouble removing it later.

We shouldn't add a class just because it's technically not possible to have a render-less widget in Vue. I see having a plain `div` as an implementation detail (since you said it was not possible to not render anything), adding a CSS class makes it turn into an API.

Haroenv:
I don't understand why the e2e tests fail still :( @yannickcr

yannickcr:
@Haroenv Seems that there is an issue with the router.
In https://app.saucelabs.com/tests/79198096cecc4c0087e8cc09d942705f the parameters from the URLs are not applied to the widgets.
And in https://app.saucelabs.com/tests/7ba577349b0d462eb1366cd5bf11ace9 the URL is not updated when navigating in categories/pages.

Haroenv:
hmm, that's weird since Vue InstantSearch isn't doing anything special with the routing there that InstantSearch wouldn't also have

src/mixins/widget.js
```diff
@@ -50,19 +57,28 @@ Read more on using connectors: https://alg.li/vue-custom`
       this.widget.dispose &&
       this.instantSearchInstance.started // a widget can't be removed if IS is not started
     ) {
-      this.instantSearchInstance.removeWidget(this.widget);
+      this.root.removeWidgets([this.widget]);
     }
   },
+  computed: {
+    root() {
```
francoischalifour: This is not the root but rather the parent index, right?

Haroenv: yes, can call it like that as well, I meant root as in nearest node, but that's indeed wrong

package.json
```diff
@@ -114,7 +114,7 @@
   "bundlesize": [
     {
       "path": "./dist/vue-instantsearch.js",
-      "maxSize": "60.50 kB"
+      "maxSize": "46.25 kB"
```
eunjae-lee: Is this the effect of removing lodash? 🎉

src/components/__tests__/Index.js
```diff
@@ -0,0 +1,90 @@
...
+  <div>
+    hi there!
+  </div>
+</div>
+
+`);
+});
+
+it('provides the index widget', done => {
+  Vue.config.errorHandler = done;
```
eunjae-lee: We don't need to clear this at some point?

package.json
Haroenv: yep!

src/components/InstantSearchSsr.js
```diff
@@ -5,7 +5,6 @@ export default createInstantSearchComponent({
   inject: {
     // should be possible to configure this with {camelcase: ['error', {allow: ['^\\$_']}]}
```
eunjae-lee:
```suggestion
```

src/components/InstantSearchSsr.js
```diff
@@ -5,7 +5,6 @@ export default createInstantSearchComponent({
...
     // but that didn't work
```
eunjae-lee:
```suggestion
```

eunjae-lee:
This seems good to me. Is there any part of the code you want us to look deeply? Or anything to test out?

src/mixins/widget.js
yannickcr: I also find `root` misleading. `parent` or `parentIndex` would be clearer?

stories/Autocomplete.stories.js
```diff
@@ -59,18 +59,9 @@ storiesOf('ais-autocomplete', module)
...
-              label: 'ATIS',
-            },
-            {
-              value: 'instantsearch_query_suggestions',
-              label: 'query-suggestions',
-            }
-          ]"
-        >
+        <ais-index index-name="airbnb" />
+        <ais-index index-name="instantsearch_query_suggestions" />
```
yannickcr: I don't really understand this. Can you explain what adding these 2 `ais-index` without children does and what it the influence on autocomplete?

src/components/__tests__/Index.js
```diff
@@ -0,0 +1,90 @@
...
+
+  const indexWidget = { $$type: 'ais.index' };
+  __setWidget(indexWidget);
+
+  const ChildComponent = {
+    inject: ['$_ais_indexWidget'],
+    mounted() {
+      this.$nextTick(() => {
```
Haroenv: check if this is necessary

src/components/__tests__/InstantSearch-integration.js
```diff
@@ -0,0 +1,31 @@
+import { mount } from '@vue/test-utils';
+import InstantSearch from '../InstantSearch';
+import { createWidgetMixin } from '../../mixins/widget';
+
+it('provides a value which works with widgetMixin', () => {
```
Haroenv: rename to be same as index integration test

stories/Autocomplete.stories.js
francoischalifour: See "scoped results": algolia/instantsearch.js#3964.

src/components/__tests__/Index.js
Haroenv: It's necessary, without this nextTick the nothing seems to be called inside mounted, I assume it's a Jest limitation

src/components/__tests__/Index-integration.js
```diff
@@ -0,0 +1,103 @@
+jest.unmock('instantsearch.js/es');
+import { mount } from '@vue/test-utils';
+import Index from '../Index';
+import instantsearch from 'instantsearch.js/es';
+import { SearchParameters } from 'algoliasearch-helper';
+import { createWidgetMixin } from '../../mixins/widget';
+import { createFakeClient } from '../../util/testutils/client';
+
+it('child widgets get added to its parent index', () => {
```
francoischalifour:
```suggestion
it('child widgets get added to their parent index', () => {
```

stories/Index.stories.js
```diff
@@ -0,0 +1,76 @@
+import { storiesOf } from '@storybook/vue';
+import algoliasearch from 'algoliasearch';
+
+storiesOf('ais-index', module)
+  .add('one search box, two indices', () => ({
```
francoischalifour: In other flavors, we usually call the first story "default" to link it more easily in the documentation.

```suggestion
  .add('default', () => ({
```

PR #750, opened 2019-12-12
chore(docs): remove docs 🔥

Haroenv:
the docs for v2 are only published on netlify, but we don't use them. Removing this step halves the build time to within the expected limit!


package.json
```diff
@@ -30,10 +30,7 @@
     "storybook": "start-storybook -p 9001 -c .storybook",
     "storybook:build": "build-storybook -c .storybook -o website/stories",
     "examples:build": "yarn build && ./examples/build.sh",
-    "docs:build": "cd docs && yarn && yarn build",
-    "docs:build:dev": "cd docs && yarn && yarn build:dev",
-    "docs:watch": "cd docs && yarn && yarn dev",
-    "docs:deploy": "cd docs && yarn && yarn deploy",
+    "docs:build": "yarn storybook:build && yarn examples:build",
```
francoischalifour: We use this command in IS.js:

https://github.com/algolia/instantsearch.js/blob/bb5b9ee1f348bb45ad778b524bce4bc2ec6dd622/package.json#L32

```suggestion
    "website:build": "yarn storybook:build && yarn examples:build",
```

netlify.toml
```diff
@@ -1,5 +1,5 @@
 [build]
-    command = "yarn docs:build:dev"
+    command = "yarn docs:build"
```
francoischalifour:
```suggestion
    command = "yarn website:build"
```

PR #756, opened 2020-01-13
feat(algoliasearch): add support for algoliasearch v4

yannickcr:
This change add support for algoliasearch v4 to Vue-InstantSearch.

No changes were necessary in the code. Also since the client is mocked in the unit tests we did not add an extra job on CIrcleCI to test with the 2 versions of the client.

The examples were updated to use the latest v4 beta.

**Important note**: we also upgraded the algoliasearch-helper dependency to 3.0.0 which is the version that support algoliasearch v4.

**Update**: after upgrading algoliasearch-helper from 2.26.1 ot 3.0.0 the bundle size jumped from 59KB to 68KB. We need to investigate on it:

<details>
<summary>Bundle size</summary>
Before:

```shell
│   Bundle Size:  241.07 KB                  │
│   Minified Size:  238.69 KB                │
│   Gzipped Size:  59.32 KB                  │

```
After:

```shell
│   Bundle Size:  281.83 KB                  │
│   Minified Size:  279.41 KB                │
│   Gzipped Size:  67.93 KB                  │
└────────────────────────────────────────────
```
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "release:publish": "scripts/release-publish.sh"
   },
   "dependencies": {
-    "algoliasearch-helper": "^3.0.0",
+    "algoliasearch-helper": "^2.28.1",
     "instantsearch.js": "^3.6.0"
   },
   "peerDependencies": {
@@ -67,7 +67,7 @@
     "@wdio/selenium-standalone-service": "5.12.1",
     "@wdio/spec-reporter": "5.12.1",
     "@wdio/static-server-service": "5.12.1",
-    "algoliasearch": ">= 3.32.0 < 5",
+    "algoliasearch": "4.0.0-beta.14",
     "babel-eslint": "10.0.1",
     "babel-jest": "23.6.0",
     "babel-preset-es2015": "6.24.1",
--- a/package.json
+++ b/package.json
@@ -47,11 +47,11 @@
     "release:publish": "scripts/release-publish.sh"
   },
   "dependencies": {
-    "algoliasearch-helper": "^2.26.1",
+    "algoliasearch-helper": "^3.0.0",
     "instantsearch.js": "^3.6.0"
   },
   "peerDependencies": {
-    "algoliasearch": "^3.30.0",
+    "algoliasearch": ">= 3.32.0 < 5",
     "vue": "^2.4.0"
   },
   "devDependencies": {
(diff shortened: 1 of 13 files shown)


francoischalifour:
The bundle size increased because:

- Vue InstantSearch v2 uses InstantSearch.js v3
- InstantSearch.js v3 uses JavaScript Helper v2
- This PR introduced JavaScript Helper v3
- → There's now a duplication of the JavaScript Helper 2 and 3 in this repo

We cannot use the JavaScript Helper v3 in Vue InstantSearch v2 because it introduced breaking changes. Why isn't the Client v4 compatible with the Helper v2? (if that's only because of the `package.json` range, we can release a version that updates it)

Haroenv:
Yes, either we make the client v4 workable with v2 of the helper, or we do this change only in the next branch on Vue InstantSearch

package.json
```diff
@@ -47,11 +47,11 @@
     "release:publish": "scripts/release-publish.sh"
   },
   "dependencies": {
-    "algoliasearch-helper": "^2.26.1",
+    "algoliasearch-helper": "^3.0.0",
```
Haroenv:
```suggestion
    "algoliasearch-helper": "^2.28.1",
```

InstantSearch.js v3 needs an update with this too (however it should also be deduplicated if it doesn't)

nunomaduro: Confirmed, we still get `Gzipped Size:  68.82 KB` with this update. We need to update instantsearch first right? 

```

/Users/nunomaduro/Algolia/vue-instantsearch/node_modules/algoliasearch-helper/package.json:
    1  {
    2:   "name": "algoliasearch-helper",
    3    "version": "2.28.1",
    4    "description": "Helper for implementing advanced search features with algolia",

/Users/nunomaduro/Algolia/vue-instantsearch/node_modules/instantsearch.js/node_modules/algoliasearch-helper/package.json:
    1  {
    2:   "name": "algoliasearch-helper",
    3    "version": "2.26.1",
    4    "description": "Helper for implementing advanced search features with algolia",
```

yannickcr: Seems that yarn did not correctly deduplicated the dependencies in the lockfile, I will fix this.

Haroenv:
Just updating the lockfile isn't enough, we also need to update InstantSearch, since we can't be sure people will update the lockfile

yannickcr:
If I'm not mistaken `instantsearch.js` has `algoliasearch-helper@^2.26.0` as dependency which already covers the new `2.28.1` so the helper gets correctly deduplicated between `vue-instantsearch` and `instantsearch.js`.

PR #764, opened 2020-03-03
feat(server side rendering): support ais-index

Haroenv:
## What?

- Updated the API for server side rendering 
- Update internal APIs (forceRender) to take index widgets in account
- Allow InstantSearch to receive previously searched results
- Updated the examples as well with the local version of Vue InstantSearch.
- small changes to the Netlify & e2e test configuration to optimise them

## How?

Rendering once in a provided method (findResultsState) which gets used in `serverPretch` (a Vue feature) to add all widgets due to the regular "created" hook. Then we search, like InstantSearch does as well (using `searchOnlyDerivedHelpers`) and use those results to do the actual rendering of the app. Frontend-wise this uses those results too, so there's an API called `hydrate` (like before) which takes those results and sets them on a place InstantSearch expects in forceRender. There's also `getState` to allow someone to get access to the results outside of the `serverPrefetch` hook.

Fixes made to InstantSearch.js itself https://github.com/algolia/instantsearch.js/pull/4359 (keeping ui state in the right order even with different widget mounting order)

## Notes

BREAKING CHANGE: `createInstantSearch` is now `createServerRootMixin`, which should be used at the highest up level. See the examples for more info.

## How to test this?

1. check out this branch
2. `yarn build` (can comment out non-esm parts in the rollup config)
3. open the server side rendering or nuxt example


src/util/createInstantSearch.js
```diff
@@ -6,50 +6,109 @@ import { warn } from './warn';
...
+      // init main index here, so force render later works as expected
+      // because main index isn't a vue widget and doesn't get __forceRender
+      search.mainIndex.init({
+        instantSearchInstance: search,
+        parent: null,
+        uiState: initialUiState,
+      });
+      // force set last results on derived helper, since we search before init
+      // TODO: PR for hydrate on index widget
+      search.mainIndex.hydrate(derivedHelpers[indexName].lastResults);
```
Haroenv: PR to InstantSearch needed for this:

```ts
    /**
     * assign last results to the derived helper, supposed to be ran after init
     */
    hydrate(results) {
      derivedHelper!.lastResults = results;
    },
```

(inside the Index widget's API)

Haroenv: https://github.com/algolia/instantsearch.js/pull/4342

src/util/createInstantSearch.js
```diff
@@ -75,19 +143,40 @@ export const createInstantSearch = instantSearchOptions => {
...
-    const { lastResults } = instantSearchState;
-    search.searchParameters = lastResults._state;
+    const { lastResults, derived } = instantSearchState;
     search.helper = algoliaHelper(searchClient, indexName, lastResults._state);
     search.helper.lastResults = new SearchResults(
       new SearchParameters(lastResults._state),
       lastResults._rawResults
     );
 
+    // TODO: this is still causing _two_ close to identical requests on the frontend
```
Haroenv: the requests could come from somewhere else, but a main difference is that the second request contains the ais-highlight tags everywhere, while the first doesn't.

src/util/createInstantSearch.js
```diff
@@ -75,19 +143,40 @@ export const createInstantSearch = instantSearchOptions => {
...
+    derivedHelpers = Object.keys(derived).reduce((acc, indexId) => {
+      if (indexId === indexName) {
+        return acc;
+      }
+
+      // eslint-disable-next-line no-param-reassign
+      acc[indexId] = {
+        // __forceRender uses lastResults only, we don't want to duplicate
+        // the requests by doing a full derive here.
+        lastResults: new SearchResults(
```
Haroenv: instead of passing it this way and using index.hydrate later, InstantSearch could have an API which accepts indexId / SearchResults map. We can use that both for hydrate & for findResultsState

src/util/createInstantSearch.js
```diff
@@ -75,19 +143,40 @@ export const createInstantSearch = instantSearchOptions => {
       warn('The result of `getState()` needs to be passed to `hydrate()`.');
       return;
     }
-    const { lastResults } = instantSearchState;
-    search.searchParameters = lastResults._state;
+    const { lastResults, derived } = instantSearchState;
```
Haroenv: lastResults is actually duplicate with derived, so we probably should get rid of it?

src/util/createInstantSearch.js
```diff
@@ -6,50 +6,109 @@ import { warn } from './warn';
...
     }
 
+    const parentIndex = componentInstance.getParentIndex();
+    const rootIndex = componentInstance.instantSearchInstance;
+
+    const parent = parentIndex !== rootIndex ? parentIndex : null;
+    const uiState = initialUiState;
+
+    // API to get helper & state is different for InstantSearch as Index widget
+    // so we need to call those conditionally only if parentIndex is the widget
```
Haroenv: clean up this comment further

src/util/createInstantSearch.js
```diff
@@ -6,50 +6,109 @@ import { warn } from './warn';
...
+    const helper = parent ? parent.getHelper() : search.helper;
+    const results = parent ? parent.getResults() : search.helper.lastResults;
```
Haroenv: could InstantSearch have these to be synchronous with an index widget? Or could we say `parent || search.mainIndex`?

Haroenv: mainIndex as the fallback is the better solution

package.json
```diff
@@ -48,7 +48,8 @@
   },
   "dependencies": {
     "algoliasearch-helper": "^3.1.0",
-    "instantsearch.js": "^4.2.0"
+    "instantsearch.js": "^4.4.0",
```
Haroenv, line 51: this needs to be updated once https://github.com/algolia/instantsearch.js/pull/4359 is merged & released

patches/instantsearch.js+4.2.0.patch
```diff
@@ -0,0 +1,90 @@
+diff --git a/node_modules/instantsearch.js/es/lib/InstantSearch.js b/node_modules/instantsearch.js/es/lib/InstantSearch.js
+index 6e7c7ee..469c7af 100644
+--- a/node_modules/instantsearch.js/es/lib/InstantSearch.js
```
Haroenv, line 3: this file needs to be removed before merging

src/instantsearch.js
```diff
@@ -2,4 +2,4 @@ export { createSuitMixin } from './mixins/suit';
 export { createWidgetMixin } from './mixins/widget';
 export * from './widgets';
 export { plugin as default } from './plugin';
-export { createInstantSearch } from './util/createInstantSearch';
+export { createServerRootMixin } from './util/createServerRootMixin';
```
Haroenv, line 5: could use a second opinion on this API, we can also go with like before: 

```
const { instantsearch, rootMixin } = createInstantSearch()
```

(now it's)

```
const rootMixin = createInstantSearch()
```

eunjae-lee: You mean, now it's `const rootMixin = createServerRootMixin()`, right?

Since the new function returns only a mixin and we don't need an instace of InstantSearch there, so `createServerRootMixin` sounds proper to me.

Haroenv: Yes, the difference is that if we would change the signature to be just like before again, you could access the InstantSearch variables without them being put inside data. I'm not sure there's an advantage for either point though

Haroenv:
I've published the patches in the example files so you can test easier, and also deployed the nuxt example to now: https://nuxt-example.fingebimus.now.sh/search

src/util/createServerRootMixin.js
```diff
@@ -0,0 +1,255 @@
...
+function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
+  /* eslint-disable no-param-reassign */
+
+  const helper = algoliaHelper(searchClient, indexName);
+  const search = instantsearch(instantSearchOptions);
+
+  let resultsState;
+
+  /**
+   * main API for SSR, called in asyncData of a root component which contains instantsearch
```
eunjae-lee, line 53:
```suggestion
   * main API for SSR, called in serverPrefetch of a root component which contains instantsearch
```

src/util/createServerRootMixin.js
```diff
@@ -0,0 +1,255 @@
...
+        reject(e);
+      })
+    );
+  });
+}
+
+function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
+  /* eslint-disable no-param-reassign */
+
+  const helper = algoliaHelper(searchClient, indexName);
```
francoischalifour, line 47: Can you explain why we need to create a new helper instance (and cannot rely on `search.mainIndex.getHelper()`)?

src/util/createServerRootMixin.js
```diff
@@ -0,0 +1,255 @@
...
+
+    const results = search.__initialSearchResults[parent.getIndexId()];
+
+    const state = results._state;
+
+    // helper gets created in init, but that means it doesn't get the injected
+    // parameters, because those are from the lastResults
+    localHelper.state = state;
+
+    // TODO: copied from index widget, should be given to all widgets IMO
```
francoischalifour, line 148: > should be given to all widgets IMO

Wdym?

scripts/clean-node-modules.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# we have something weird going on, where installing via `file:../..` doesn't take "files" in account, and thus copies recursively. This causes cache to take many minutes and us to time-out. The long-term solution is to use prevent the recursive copying too, which will make the install faster.
+
+mkdir /tmp/empty
```
eunjae-lee, line 5: I'm not sure if this works in other environment like Windows, but let's fix it in another PR in another time.

src/util/createServerRootMixin.js
```diff
@@ -0,0 +1,255 @@
...
+  );
+}
+
+function searchOnlyWithDerivedHelpers(helper) {
+  return new Promise((resolve, reject) => {
+    helper.searchOnlyWithDerivedHelpers();
+
+    // we assume all derived helpers resolve at least in the same tick
+    helper.derivedHelpers[0].on('result', e => {
+      resolve(e);
```
eunjae-lee, line 33:
```suggestion
      resolve();
```

`e` doesn't seem to be used.

examples/nuxt/pages/search.vue
```diff
@@ -29,48 +53,64 @@
...
+  serverPrefetch() {
+    return this.instantsearch.findResultsState(this).then(algoliaState => {
+      this.$ssrContext.nuxt.algoliaState = algoliaState;
+    });
   },
   beforeMount() {
-    // Nuxt will merge `asyncData` and `data` on the client
-    instantsearch.hydrate(this.algoliaState);
+    // const results = this.$nuxt.context.nuxtState.algoliaState;
+    // TODO: read from a real Nuxt API, but this.$nuxt.context is undefined
```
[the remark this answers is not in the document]
Haroenv: I'll leave it as the `window`. In principle the other code should work too, but 1/2 times it threw an error for me

examples/ssr/src/main.js
```diff
@@ -14,30 +15,104 @@ Vue.config.productionTip = false;
...
   });
 
   const app = new Vue({
-    mixins: [rootMixin],
+    mixins: [
+      createServerRootMixin({
+        searchClient,
+        indexName: 'instant_search',
+        routing: {
+          router: {
```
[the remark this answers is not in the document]
Haroenv: Yes, I'd like to add this as part of the library soon after. The hard part is a universal way to get the current URL. That seems to be different in nuxt vs. vanilla vue server side rendering

src/util/createServerRootMixin.js
Haroenv: this code is written in the index widget, but not passed as an argument, so I need to recreate it here as far as I can tell

scripts/clean-node-modules.sh
Haroenv: netlify is only macOS so that's no big deal IMO

eunjae-lee: I meant @yannickcr ... 😆

src/util/createServerRootMixin.js
```diff
@@ -0,0 +1,255 @@
...
+
+        search.hydrate(results);
+
+        resultsState = Object.keys(results)
+          .map(indexId => {
+            const { _state, _rawResults } = results[indexId];
+            return [
+              indexId,
+              {
+                _state: JSON.parse(JSON.stringify(_state)),
```
eunjae-lee, line 101: Can you explain why we stringify and parse it again?

Haroenv: _state is a class, so we want to get all the keys, and only those which aren't functions. toJson is the simplest way to trigger that. In principle we could change this to Object.keys.reduce instead

scripts/clean-node-modules.sh
yannickcr: I'm on Linux, I'm not that barbaric :rofl:

eunjae-lee: ohhh my bad 🤪

src/util/createServerRootMixin.js
eunjae-lee: Ah got it. Can you add this comment in the code?

src/util/createServerRootMixin.js
Haroenv: helper is still `null` before init is called, right?

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -0,0 +1,467 @@
...
+      } = mount(app);
+
+      expect(instantsearch.mainIndex.getHelper()).toBe(null);
+
+      instantsearch.hydrate({
+        __identifier: 'stringified',
+        hello: serialized,
+      });
+
+      expect(instantsearch.mainIndex.getHelper().constructor.name).toBe(
```
francoischalifour, line 337: Testing `constructor.name` is an implementation detail and can break once we swap the helper implementation. What about asserting with `expect.any(AlgoliaSearchHelper)` as below?

src/util/createServerRootMixin.js
francoischalifour: Yeah, that's right!

src/util/createServerRootMixin.js
francoischalifour: Yes it makes sense. A comment in the code for "should be given to all widgets IMO" might not be the best place though.

src/util/__tests__/createServerRootMixin.test.js
Haroenv: It didn’t work for some reason, one test it worked and the next it failed. I think maybe Babel might have been involved in it, but this was the simplest way I found to have a passing test which asserts something

francoischalifour: Ok, then asserting for not `null` might be enough here?

PR #774, opened 2020-05-07
upgrading version of algoliasearch-helper to improve bundle size

mfissehaye:
The bundle size of algoliasearch-helper has been improved explained in [this](https://github.com/algolia/algoliasearch-helper-js/issues/617#issuecomment-510527506) comment. This pull request upgrades the version in vue-instantsearch for a smaller bundle size.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ coverage
 /website/stories
 /website/examples/*
 !/website/examples/index.html
+
+.idea/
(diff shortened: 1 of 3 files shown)


Haroenv:
Sorry, the helper has breaking changes that need to be taken in account upstream. You can use version 3 beta of Vue InstantSearch which already uses helper@3

Thanks for your pull request though!

PR #777, opened 2020-05-14
fix(range-input): refine correctly when start is given

eunjae-lee:


src/components/RangeInput.vue
```diff
@@ -11,7 +11,7 @@
     >
       <form
         :class="suit('form')"
-        @submit.prevent="refine({ min: minInput, max: maxInput })"
+        @submit.prevent="refine({ min: minInput || values.min, max: maxInput || values.max })"
```
dotboris, line 14: I believe that this doesn't fully fix the bug. I think that it'll fix everything when a refresh is done, but the input will still behave strangely when you hit the browser's back button.

Let's say that you a range input:

1.  You put in min: 10, max: 100
1.  You hit Go
1.  You change min to 20
1.  You hit Go again
1.  You hit the back button on the browser (with the history router the page will not reload but the state of the search changes)
1.  You hit Go again

In this scenario, the min value will be reset to 60.

I believe that this will still happen with your fix. In that scenario `minValue` would have been set to `60` and `values.min` would have been set to `50`.

dotboris: What if you added a watcher on `value.min` & `value.max`? In that watcher you could update `minValue` & `maxValue` every time `value.min` & `value.max` get updated by any source.

Haroenv:
minInput is already a computed variable, not sure if that needs to be computed in the render again. Could you maybe start with writing a test of the behaviour which is failing now?

eunjae-lee:
> minInput is already a computed variable, not sure if that needs to be computed in the render again. Could you maybe start with writing a test of the behaviour which is failing now?

I don't think `minInput` is a computed variable.

```js
@change="minInput = $event.currentTarget.value"
```

```js
  data() {
    return {
      minInput: undefined,
      maxInput: undefined,
    };
  },
```

It just stores the value from @change event handler.

And the test case I added fails without my change.

https://github.com/algolia/vue-instantsearch/pull/777/files#diff-17a43ab3d900702112db6482b0d0fc51R311-R333

src/components/RangeInput.vue
eunjae-lee: > In this scenario, the min value will be reset to 60.
> I believe that this will still happen with your fix. In that scenario minValue would have been set to 60 and values.min would have been set to 50.

Do you mean it will be reset to 20, while it should've stayed as 10, right?
You're right. I didn't think of that path. Thanks!

eunjae-lee:
Kapture 2020-05-15 at 11 29 13

src/components/RangeInput.vue
dotboris, line 14: I've just noticed that this is going to break when `minInput` or `maxInput` are set to `0`. Since `0` is falsy in JS, it'll end up taking the right value of the `||`

eunjae-lee: Thanks @dotboris for the review and finding it out !
https://github.com/algolia/vue-instantsearch/pull/777/commits/5189f33bd60d6279d08a194074423ddd891df8b7

Haroenv:
In a similar range input I made yesterday I used a watcher instead of `updated` to update the min & max values: https://codesandbox.io/s/thirsty-taussig-y61nc?file=/src/UnitRangeInput.vue

eunjae-lee:
Netlify fails due to time out, and e2e fails with connection errors. (I retried a couple of times). So I'm just merging this.

PR #783, opened 2020-05-29
fix(Highlight, Snippet): prevent XSS via routing

eunjae-lee:
related to: https://github.com/algolia/instantsearch.js/pull/4344

In Vue InstantSearch, `v-html` is used in `ais-highlight` and `ais-snippet`. This can allow reflected Cross-Site Scripting (XSS) attack with a malicious URL.

## How to fix

The search API response includes `_highlightResult` or `_snippetResult`. And we've been using those values at `v-html` directly.

On the other hand, React InstantSearch parses `_highlightResult` or `_snippetResult` and manually render dom with it. So I've copied the logic from React InstantSearch.

`Highlighter.vue` is where the manual rendering happens. It's used by both `ais-highlight` and `ais-snippet`.

## What happens to users

It is only internal change. So it doesn't affect users.

## Clean up

- `src/util/highlight.js`
- `src/util/__tests__/highlight.test.js`

The files above are copied from React InstantSearch.
Once this PR is merged (or right before merged?), we can move these files back to InstantSearch.js, so that Vue InstantSearch and Angular InstantSearch can share the same logic without copy&paste.


src/components/Highlighter.vue
```diff
@@ -0,0 +1,62 @@
...
+        preTag: this.preTag,
+        postTag: this.postTag,
+      });
+    },
+  },
+};
+</script>
+
+<style>
+</style>
```
Haroenv, line 62: No need for style tag

src/util/highlight.js
```diff
@@ -0,0 +1,113 @@
+// copied from React InstantSearch
+const HIGHLIGHT_TAGS = {
+  highlightPreTag: `<ais-highlight-0000000000>`,
```
Haroenv, line 3: Let’s use the one from InstantSearch.js: `__ais-highlight__` & `__/ais-highlight`. That way it’s not a breaking change & you can choose to escape the whole string without issue still.

src/util/highlight.js
```diff
@@ -0,0 +1,113 @@
...
+  highlightPostTag: `</ais-highlight-0000000000>`,
+};
+
+const getPropertyByPath = (object, path) =>
```
Haroenv, line 7: We have this helper already in lib/object or something like that

src/util/highlight.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  return parseHighlightedAttribute({
+    preTag,
+    postTag,
+    highlightedValue: highlightObject.value,
+  });
+}
+
+export const highlight = ({
```
Haroenv, line 100: This level of function doesn’t really seem needed, since we only use parseAlgoliaHit, right?

src/components/Highlight.vue
```diff
@@ -1,17 +1,23 @@
...
-  <span
-    :class="suit()"
-    v-html="innerHTML"
+  <highlighter
+    :hit="hit"
+    :attribute="attribute"
+    :highlighted-tag-name="highlightedTagName"
+    :suit="suit"
+    highlight-property="_highlightResult"
+    pre-tag="<mark>"
```
Haroenv, line 8: Would this not more be “highLightComponent" instead of strings?

Haroenv: Actually that’s highlightedTagName, what’s this used for then?

src/components/Highlight.vue
```diff
@@ -1,17 +1,23 @@
 <template>
-  <span
-    :class="suit()"
-    v-html="innerHTML"
+  <highlighter
```
Haroenv, line 2: A lowercase component without split isn’t valid. We can call this AisHighlighter

src/components/Highlighter.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <span
+    :class="suit()"
+  >
+    <component
+      v-for="({ value, isHighlighted }, index) in parsedHighlightedValue"
+      :class="[isHighlighted && suit('highlighted')]"
+      :key="index"
+      :is="isHighlighted ? highlightedTagName : textNode"
```
Haroenv, line 9: Why not span as non highlighted?

src/components/Snippet.vue
```diff
@@ -1,17 +1,23 @@
...
-  <span
-    :class="suit()"
-    v-html="innerHTML"
+  <highlighter
+    :hit="hit"
+    :attribute="attribute"
+    :highlighted-tag-name="highlightedTagName"
+    :suit="suit"
+    highlight-property="_snippetResult"
+    pre-tag="<mark>"
```
Haroenv, line 8: Thinking more, the pre tag here should be what the user or we set as highlightPreTag, no? So that would be the `__ais-highlight__`

src/components/Highlighter.vue
eunjae-lee: François is right. In React InstantSearch, span is used for non highlighted, but not in the others. Just text node.

src/components/Snippet.vue
eunjae-lee: oh okay, I think I am missing some stuff.

image

I'm seeing `__ais-highlight__` in the network panel.

but when I log `hits` in the `Hits` component,

image

I see `<mark>`. Do you when it's replaced by which?

Haroenv: it's replaced by `escapeHighlight` in InstantSearch.js, you need to set the replaced tags there forcibly somehow to the original tags, or disable `escapeHighlight`

eunjae-lee: @Haroenv Okay, I think I can create a pull-request in InstantSearch.js.
In the PR, there is [parseAlgoliaHit](https://github.com/algolia/vue-instantsearch/pull/783/files#diff-26a063579221d23a8e71d6870cc3c702R100) function (or whatever the name is). It behaves according to `hits.__escaped`.

If it's `true`, it will parse with

```js
export const TAG_REPLACEMENT = {
  highlightPreTag: '<mark>',
  highlightPostTag: '</mark>',
};
```

If it's `false`, it will parse with

```js
export const TAG_PLACEHOLDER = {
  highlightPreTag: '__ais-highlight__',
  highlightPostTag: '__/ais-highlight__',
};
```

Then Vue-InstantSearch uses this function.
What do you think?

Haroenv: that should work, good idea. Ideally in InstantSearch.js we should change the behaviour to not just replace, but in this case it should work.

eunjae-lee: Yes, in this way, we don't break InstantSearch.js and provide extra function for flavors. I'll continue on this path!

src/components/Highlighter.vue
Haroenv, line 62:
```suggestion
```

eunjae-lee:
I'll re-work this pr, once I finish the core part in InstantSearch.js.

Haroenv:
I think the best course of action is shipping this as part of Vue InstantSearch in v3 & then once InstantSearch has the methods too, we migrate to those (InstantSearch.js is pinned as a dependency)

eunjae-lee:
I thought we agreed on working on InstantSearch.js first.

I can continue to work on this PR but why not in Vue-IS@v2?

Haroenv:
Because this needs to be released (as part of v2 & v3) before v3 can be released, otherwise the "vulnerability" would be in effect for 3.0.0

src/util/highlight.js
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/783/commits/caba208088709f649e81ab4a6c4dd9a06e9e7be5

src/util/highlight.js
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/783/commits/8a2ca6e62fe3f5049fec87240b40495410bcd4de

src/components/Highlight.vue
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/783/commits/395b9856f3c3f9927957e02d98a63f3d7cf04fdd

src/components/Snippet.vue
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/783/commits/711134ce8e047b3ac3d0c39547a37751b2d5fece

src/components/Highlight.vue
```diff
@@ -1,17 +1,23 @@
...
-    :class="suit()"
-    v-html="innerHTML"
+  <ais-highlighter
+    :hit="hit"
+    :attribute="attribute"
+    :highlighted-tag-name="highlightedTagName"
+    :suit="suit"
+    highlight-property="_highlightResult"
+    pre-tag="<mark>"
+    post-tag="</mark>"
```
Haroenv, line 9: We need to be able to fall back to `__ais-highlight__` / `__/ais-highlight__` if escape-highlight is false of the parent hits or infinite hits

eunjae-lee: yeah, I said [that](https://github.com/algolia/vue-instantsearch/pull/783#discussion_r433163166), but I realized that we have only `hit` inside ais-highlight. We don't have `hits`, so cannot tell `hits.__escaped` is true or not.

eunjae-lee: Previously, Vue InstantSearch just set innerHTML with

```js
computed: {
    innerHTML() {
      return instantsearch.highlight({
        attribute: this.attribute,
        hit: this.hit,
        highlightedTagName: this.highlightedTagName,
      });
    },
  },
```

and I see `highlight` function in InstantSearch.js already assumes the hit is escaped.
What do you think?

eunjae-lee:
Yes, I agree. We could've introduced an option here for non-escaped values, but seems out of the scope.
And one thing I've discovered that I think we should address later is, the current version in Vue-InstantSearch cannot handle array properties, like

```vue
<ais-highlight attribute="categories" :hit="item"></ais-highlight>
```

where `categories` is an array. It's supported in React InstantSearch with [`separator`](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-dom/src/components/Highlighter.js#L91:L91).

Haroenv:
I think it would be able to do if you loop outside and render `categories[I]`

eunjae-lee:
@Haroenv I've found an issue here.
In `ais-highlight` and `ais-snippet`, normally we expect `hits` to be already escaped from [here](https://github.com/algolia/instantsearch.js/blob/master/src/connectors/hits/connectHits.js#L76-L78).

So values like `TV & Home Theater` are displayed as `TV &amp; Home Theater`.

I'm thinking of writing `unescape` function in Vue-InstantSearch (to be aligned with [`escape`](https://github.com/algolia/instantsearch.js/blob/master/src/lib/utils/escape.ts)).

What do you think?

Haroenv:
good point @eunjae-lee, can you also add a new test for that? It is going to be safe, since we no longer use v-html

PR #794, opened 2020-06-10
chore: add shipjs

eunjae-lee:
This PR adds Ship.js as a release tool.
I've added the environment variables to CircleCI.

Once this is merged, running `yarn release` will prepare the next beta version automatically which is `v3.0.0-beta.2` (based on the current version `v3.0.0-beta.1`).

When we decide to ship it as a stable version, after running `yarn release`, when Ship.js asks if the next version is x.y.z, then we type "n" and then type "3.0.0", then it will continue as a stable release.


.circleci/config.yml
```diff
@@ -68,10 +68,30 @@ jobs:
...
       - test_build
       - test_unit
       - test_e2e
+  release_if_needed:
+    jobs:
+      - shipjs_trigger:
+          filters:
+            branches:
+              only:
+                - next
```
Haroenv, line 96:
```suggestion
```

package.json
```diff
@@ -44,7 +44,8 @@
     "changelog": "conventional-changelog --preset angular --infile CHANGELOG.md --same-file",
     "changelog:unreleased": "conventional-changelog --preset angular --output-unreleased",
     "release:branch": "scripts/release-branch.sh",
-    "release:publish": "scripts/release-publish.sh"
+    "release:publish": "scripts/release-publish.sh",
```
Haroenv, line 47: both these scripts should be removed from package.json & the file system :)

eunjae-lee: Yes, but I was thinking of removing them in a separate PR once you use Ship.js in this repo and it works as you expect. I can remove them here if you want :)

Haroenv: If shipjs doesn't work we can always revert this commit, otherwise we will forget to remove it I think

eunjae-lee: okay good point

PR #801, opened 2020-06-11
fix(server): make vue-server-renderer not a dependency

Haroenv:
this is done by making this dependency-injected to prevent the dependency being required for everyone

this is a breaking change, but I think I'll release it as 3.1.0 and will deprecate 3.0.0 & 3.0.1 as soon as this is merged and released Monday

### DIFF
--- a/examples/nuxt/package.json
+++ b/examples/nuxt/package.json
@@ -14,7 +14,8 @@
     "algoliasearch": "^4.0.1",
     "cross-env": "^5.2.0",
     "nuxt": "^2.4.5",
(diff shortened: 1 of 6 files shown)


Haroenv:
`npm deprecate`

yannickcr:
It may be possible to put it in `optionalDependencies` and load it conditionally. This way we keep the current API and can enforce a specific version without requiring it explicitly?

Haroenv:
that only  works in commonjs, but the problem is for the esm build

yannickcr:
What is not working in the esm build ?

PR #804, opened 2020-06-16
feat(InfiniteHits): support cache

eunjae-lee:
This PR adds support for cache.
related to: https://github.com/algolia/instantsearch.js/pull/4431

It basically accepts a new prop `cache` and pass it to the connector.
I've added a story to test sessionStorage cache in action.


stories/InfiniteHits.stories.js
```diff
@@ -1,5 +1,6 @@
 import { storiesOf } from '@storybook/vue';
 import { action } from '@storybook/addon-actions';
+import { createInfiniteHitsSessionStorageCache } from 'instantsearch.js/es/lib/infiniteHitsCache';
```
francoischalifour, line 3: Should we reexport `createInfiniteHitsSessionStorageCache` from `vue-instantsearch`?

src/components/InfiniteHits.vue
```diff
@@ -85,13 +85,18 @@ export default {
...
+      default: undefined,
+    },
   },
   computed: {
     widgetParams() {
       return {
         showPrevious: this.showPrevious,
         escapeHTML: this.escapeHTML,
         transformItems: this.transformItems,
+        cache: this.cache,
```
francoischalifour, line 99: Do we want to test that it forwards the `cache` option to the connector? (I thought we did for all widgets.)

stories/InfiniteHits.stories.js
Haroenv: It's an interesting idea, but we don't do that for now for other things like router, I'm not sure if that can have a bundle size impact?

src/components/InfiniteHits.vue
Haroenv: I don't think we test every forward, since usually the test & implementation is written at the same time, so it will have the same mistake (if any) right?

francoischalifour: That's mainly to test future unintended changes and regressions. If we somehow edit/remove this line and that the `cache` feature doesn't work, no tests will tell us so.

(I'm pointing this out because I thought we did for all widgets — apparently we don't.)

stories/InfiniteHits.stories.js
francoischalifour: I don't think it'll have a bundle size impact but it's worth trying.

On a UX perspective, it doesn't make sense to import something from a library's dependency (you're not even aware that you use `instantsearch.js` when you use `vue-instantsearch`).

Haroenv: let's do it in a stacked PR? There's argument for exposing the connectors etc. too, so let's join that?

eunjae-lee: I guess not in a stacked PR but a separate PR as in something like `chore: reexport things from IS.js...`

If we're going to reexport stuff, then we'd better do them altogether with doc changes.
What do you think?

src/components/InfiniteHits.vue
eunjae-lee: If so, what if we test if this component exposes a correct `widgetParams`? I think it's `createWidgetMixin`'s job to forward widget params to the connector and "is it forwarded correctly?" needs to be tested there.

If we want to keep the contract of what needs to be passed, then we could just test snapshots of the widgetParams.

PR #809, opened 2020-07-02
fix(routing): do not warn when `router` or `stateMapping` is missing

francoischalifour:
In the `routing` doc, we state:

> The object is accepted if it has either of these keys [...]
>
> — [Reference](https://www.algolia.com/doc/api-reference/widgets/instantsearch/vue/#widget-param-routing)

But we actually do warn when `router` or `stateMapping` is missing. This removes the warning in this case.

### DIFF
--- a/src/components/__tests__/InstantSearch.js
+++ b/src/components/__tests__/InstantSearch.js
@@ -3,8 +3,15 @@ import { mount } from '@vue/test-utils';
 import instantsearch from 'instantsearch.js/es';
 import InstantSearch from '../InstantSearch';
 import { version } from '../../../package.json';
+import { warn } from '../../util/warn';
 
-beforeEach(() => jest.clearAllMocks());
+jest.mock('../../util/warn');
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  warn.mockReset();
+  warn.mockRestore();
+});
 
 it('passes props to InstantSearch.js', () => {
   const searchClient = {};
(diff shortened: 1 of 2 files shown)


src/components/__tests__/InstantSearch.js
```diff
@@ -3,8 +3,15 @@ import { mount } from '@vue/test-utils';
...
 import { version } from '../../../package.json';
+import { warn } from '../../util/warn';
 
-beforeEach(() => jest.clearAllMocks());
+jest.mock('../../util/warn');
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  warn.mockReset();
+  warn.mockRestore();
```
yannickcr, line 13: I don't think `warn.mockReset()` and `warn.mockRestore()` are useful here.

* The value of the mock's usage data are already reset by `jest.clearAllMocks()`.
* `warn.mockReset()` also removes mocked return values or implementations but there is none here.
* `warn.mockRestore()` only works for mocks created with `jest.spyOn()` which is not the case here.

src/components/__tests__/InstantSearch.js
```diff
@@ -213,9 +217,44 @@ found in
...
+  });
+  mount(InstantSearch, {
+    propsData: {
+      searchClient: {},
+      indexName: 'indexName',
+      routing: { stateMapping: {} },
+    },
+  });
+
+  expect(warn).toHaveBeenCalledTimes(0);
```
yannickcr, line 257: I find it more clear to use `not.toHaveBeenCalled` here. WDYT?

```suggestion
  expect(warn).not.toHaveBeenCalled();
```

francoischalifour: We use both in this codebase and mainly `toHaveBeenCalledTimes(0)` in InstantSearch.js because it provides better error messages.

PR #827, opened 2020-07-31
fix(highlight): don't remove whitespace-only nodes

### LINKED ISSUE #826: ais-highlight removes spaces
**Bug 🐞**

When typing in the search-box, ais-highlight removes spaces between consecutive words.

2020-07-31 08 14 12 vue-instantsearch netlify app 5327eff91408

https://codesandbox.io/s/github/algolia/create-instantsearch-app/tree/templates/vue-instantsearch

Spaces between highlighted words should remain.

3.1.0

Haroenv:
Vue by default removes nodes which are just whitespace: https://github.com/vuejs/vue/blob/ec78fc8b6d03e59da669be1adf4b4b5abf670a34/dist/vue.common.dev.js#L2527

We need to prevent this, since sometimes between highlights, there is just a space, we need to make the value not "wrapped". We can't make the value always "block" (i don't know the correct name for this), since that would add spaces between "inline" highlights.

fixes #826

### DIFF
--- a/src/components/Highlighter.vue
+++ b/src/components/Highlighter.vue
@@ -40,7 +40,7 @@ export default {
         functional: true,
         render(createElement, context) {
           const slots = context.slots();
-          return slots.default || ' ';
+          return slots.default;
         },
       },
     };
--- a/src/components/Highlighter.vue
+++ b/src/components/Highlighter.vue
@@ -5,7 +5,7 @@
       :class="[isHighlighted && suit('highlighted')]"
       :key="index"
       :is="isHighlighted ? highlightedTagName : textNode"
-    ><template v-if="value === ' '"> {{ value }} </template><template v-else>{{ value }}</template></component>
+    >{{ value }}</component>
   </span>
 </template>
 
--- a/src/components/Highlighter.vue
+++ b/src/components/Highlighter.vue
@@ -5,9 +5,7 @@
       :class="[isHighlighted && suit('highlighted')]"
       :key="index"
       :is="isHighlighted ? highlightedTagName : textNode"
-    ><template v-if="value === ' '">
-      {{ value }}
-    </template><template v-else>{{ value }}</template></component>
+    ><template v-if="value === ' '"> {{ value }} </template><template v-else>{{ value }}</template></component>
   </span>
 </template>
 
--- a/src/components/Highlighter.vue
+++ b/src/components/Highlighter.vue
@@ -1,13 +1,13 @@
 <template>
-  <span
-    :class="suit()"
-  >
+  <span :class="suit()">
     <component
       v-for="({ value, isHighlighted }, index) in parsedHighlights"
       :class="[isHighlighted && suit('highlighted')]"
       :key="index"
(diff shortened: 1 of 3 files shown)


src/components/Highlighter.vue
```diff
@@ -40,7 +40,7 @@ export default {
         functional: true,
         render(createElement, context) {
           const slots = context.slots();
-          return slots.default;
+          return slots.default || ' ';
```
Haroenv, line 43:
```suggestion
          return slots.default;
```

this doesn't work (cry cry)

src/components/Highlighter.vue
```diff
@@ -1,13 +1,13 @@
...
+  <span :class="suit()">
     <component
       v-for="({ value, isHighlighted }, index) in parsedHighlights"
       :class="[isHighlighted && suit('highlighted')]"
       :key="index"
       :is="isHighlighted ? highlightedTagName : textNode"
-    >{{ value }}</component>
+    ><template v-if="value === ' '">
+      {{ value }}
+    </template><template v-else>{{ value }}</template></component>
```
eunjae-lee, line 10: I'm having a hard time understanding this even after reading your summary. Can you help me a little bit? How is `v-if` different from `v-else`?

Kocal: It is different due to how white spaces are handled in the first block and the second block, 
image

Haroenv: Thanks @Kocal, indeed the difference is in the form. If text or moustache is directly without space to a tag, it will be rendered without wrapping spaces, however if it's on a newline, whitespace-only blocks will be preserved.

see https://github.com/vuejs/vue/issues/9208

Haroenv: I've just changed it slightly  to make the difference more clear

leading & trailing whitespace | before this PR
---|---
Screenshot 2020-07-31 at 13 41 18 | Screenshot 2020-07-31 at 13 41 01

eunjae-lee: Thanks for the explanation @Kocal @Haroenv 
This is tricky 😅

francoischalifour: @Haroenv Since this fix isn't straightforward, can add a comment in the code that explains what you described in https://github.com/algolia/vue-instantsearch/pull/827#discussion_r463560724 including the Vue issue?

Haroenv: Good call. Since you can't leave a comment in that place (you'd change the rendering), I found an alternative solution, where you can leave a comment

src/util/parseAlgoliaHit.js
```diff
@@ -39,7 +39,10 @@ function parseHighlightedAttribute({ preTag, postTag, highlightedValue = '' }) {
 
       if (splitByPostTag[1] !== '') {
         elements.push({
-          value: splitByPostTag[1],
+          // Vue removes nodes which are just a single space (vuejs/vue#9208),
```
Haroenv, line 42: is this what you expected @francoischalifour?

francoischalifour: Yep, thank you!

PR #830, opened 2020-08-05
fix(ssr): fix missing scopedResults during render

### LINKED ISSUE #829: AisCurrentRefinements app crash - cannot read property `reduce` of undefined
**Bug 🐞**

Making use of `AisCurrentRefinements` causes the app to crash with the error:

> TypeError: Cannot read property 'reduce' of undefined

[Sandbox](https://github.com/jackwkinsey/uvue-sandbox)

Some configuration is required:

You'll need to copy `.env.example` to `.env.local` and provide the following in that file:

```
VUE_APP_ALGOLIA_API_KEY
VUE_APP_ALGOLIA_APP_ID
VUE_APP_ALGOLIA_INDEX_NAME
```

To run, execute `npm run ssr:serve` then navigate to the search page and select a refinement to filter by (which should change your url; i.e. `/search?brand=somebrand`). If the app hasn't crashed yet then refresh and you should see the error since the page should crash.

This sandbox mimics our project's setup and uses the `AisCurrentRefinements` component (though this issue also exists with the `AisClearRefinements` component). Using this component causes the app to crash. We're gracefully handling this crash in our actual project so the component crashing doesn't crash the whole app and allows everything to work as expected, but we do get an error logged in our system.

The `AisCurrentRefinements` component should work as expected without causing a crash.

It seems to be caused by an initial render of the component.

From the error log it appears that `./node_modules/instantsearch.js/es/connectors/current-refinements/connectCurrentRefinements.js` calls `reduce` on an array that can be undefined in the `render` method. (Note: a similar issue exists in `connectClearRefinements.js`)

Specifically, the render function has these lines:

```
var scopedResults = _ref3.scopedResults;
...
var items = scopedResults.reduce(...)
```

For some reason, `scopedResults` can be undefined (seems to be on initial render; we are using uvue and SSR which may be a cause for this) which causes the `reduce` call to blow up. If a check is added then the app nor the component crashes and everything works as expected (i.e. the `AisCurrentRefinements` component works as intended).

vue-instantsearch: 3.1.0

yannickcr:
The `scopedResults` object was missing when rendering the widget on server-side, crashing the widgets that were relying on it (like 
`AisCurrentRefinements` for example).

This PR fixes this by recalculating the scoped results and adding them to the render options (for this we're using the same method as in InstantSearch.js https://github.com/algolia/instantsearch.js/blob/68b20f487fcd54fd7dec11b4c494b6aa94a18516/src/widgets/index/index.ts#L154-L167)

Fixes #829

### DIFF
--- a/src/util/__tests__/createServerRootMixin.test.js
+++ b/src/util/__tests__/createServerRootMixin.test.js
@@ -414,6 +414,13 @@ Array [
         {
           helper: expect.any(Object),
           results: expect.any(Object),
+          scopedResults: expect.arrayContaining([
+            expect.objectContaining({
+              helper: expect.any(Object),
+              indexId: expect.any(String),
+              results: expect.any(Object),
+            }),
+          ]),
           state: expect.any(Object),
           instantSearchInstance: expect.any(Object),
         },
@@ -423,6 +430,13 @@ Object {
   "helper": Any<Object>,
   "instantSearchInstance": Any<Object>,
   "results": Any<Object>,
+  "scopedResults": ArrayContaining [
+    ObjectContaining {
+      "helper": Any<Object>,
+      "indexId": Any<String>,
+      "results": Any<Object>,
+    },
+  ],
   "searchMetadata": Object {
     "isSearchStalled": false,
   },
--- a/src/util/createServerRootMixin.js
+++ b/src/util/createServerRootMixin.js
@@ -178,9 +178,27 @@ function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
           }, {}),
       });
 
+    function resolveScopedResultsFromWidgets(widgets) {
+      const indexWidgets = widgets.filter(w => w.$$type === 'ais.index');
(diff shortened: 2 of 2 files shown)


src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -414,6 +414,13 @@ Array [
         {
           helper: expect.any(Object),
           results: expect.any(Object),
+          scopedResults: expect.arrayContaining([
+            expect.objectContaining({
+              helper: expect.any(Object),
+              indexId: expect.any(String),
```
francoischalifour, line 420: Why don't we assert actual values?

yannickcr: Since most of the values are coming from the test itself or from mocks then we are just checking the shape of the `RenderOptions` object, not its actual values.

yannickcr:
@Haroenv should be the same.

PR #856, opened 2020-09-08
chore(ci): fix e2e tests

Haroenv:
- changed examples script to allow building a single example
- actually build & move folders for e-commerce demo
- fix a small error in slider in the example

### DIFF
--- a/examples/e-commerce/src/App.vue
+++ b/examples/e-commerce/src/App.vue
@@ -123,7 +123,6 @@
                 <template slot="default">
                   <ais-range-input attribute="price">
                     <div slot-scope="{ currentRefinement, range, refine, canRefine }">
-                      {{ { range, currentRefinement, canRefine } }}
                       <vue-slider
                         :min="range.min"
                         :max="range.max"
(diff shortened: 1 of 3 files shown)


examples/e-commerce/src/App.vue
```diff
@@ -302,7 +302,6 @@
               <div
                 slot-scope="{
                   currentRefinement,
-                  nbPages,
```
yannickcr, line 305: Why removing this? Is it unused?

examples/e-commerce/src/App.vue
```diff
@@ -515,8 +514,8 @@ export default {
     formatNumber,
     toValue(value, range) {
       return [
-        value.min !== null ? value.min : range.min,
-        value.max !== null ? value.max : range.max,
+        value.min != null ? value.min : range.min,
+        value.max != null ? value.max : range.max,
```
yannickcr, line 518: Would not be better to test for `value.min !== null && value.min !== undefined` or `typeof value.min === 'number'` (if this is the correct condition) here?

examples/e-commerce/src/App.vue
Haroenv: yep, unused

eunjae-lee (approved):
The change looks good to me as long as the CI passes :)

PR #863, opened 2020-09-22
feat(ssr): forward router instance to findResultsState clone

Haroenv:
in our guide for routing with vue router, you need to access the Vue Router instance. The most straightforward way for that is to move the construction of the server mixin into `data()`. `router` is useful for creating an instantsearch router compatible with vueRouter

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "vue-jest": "2.6.0",
     "vue-json-tree": "0.3.3",
     "vue-loader": "14.2.2",
-    "vue-router": "^3.4.3",
+    "vue-router": "3.4.3",
     "vue-server-renderer": "^2.6.11",
     "vue-slider-component": "3.0.15",
     "vue-template-compiler": "2.5.18",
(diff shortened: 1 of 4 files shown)


eunjae-lee (approved):
makes sense to me

package.json
```diff
@@ -103,6 +103,7 @@
     "vue-jest": "2.6.0",
     "vue-json-tree": "0.3.3",
     "vue-loader": "14.2.2",
+    "vue-router": "^3.4.3",
```
francoischalifour, line 106: Do we want to pin that dep?

Haroenv:
```suggestion
    "vue-router": "3.4.3",
```

yarn.lock
```diff
@@ -15612,6 +15612,11 @@ vue-property-decorator@^8.0.0:
   dependencies:
     vue-class-component "^7.0.1"
 
+vue-router@^3.4.3:
```
Haroenv, line 15615:
```suggestion
vue-router@3.4.3:
```

francoischalifour (approved):

PR #864, opened 2020-09-22
feat(ssr): forward vuex instance to findResultsState clone

Haroenv:
Regardless of whether you _need_ to access it, it's commonly done for otherwise prerendered state. Since Vuex is a core plugin, we must support it like this

stacked PR on #863

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,8 @@
     "vue-server-renderer": "^2.6.11",
     "vue-slider-component": "3.0.15",
     "vue-template-compiler": "2.5.18",
-    "vuetify": "1.5.3"
+    "vuetify": "1.5.3",
+    "vuex": "3.5.1"
   },
   "bundlesize": [
     {
--- a/src/util/__tests__/createServerRootMixin.test.js
+++ b/src/util/__tests__/createServerRootMixin.test.js
@@ -2,6 +2,7 @@ import Vue from 'vue';
 import { mount } from '@vue/test-utils';
 import _renderToString from 'vue-server-renderer/basic';
 import Router from 'vue-router';
+import Vuex from 'vuex';
 import { createServerRootMixin } from '../createServerRootMixin';
 import InstantSearchSsr from '../../components/InstantSearchSsr';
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -107,7 +107,8 @@
     "vue-server-renderer": "^2.6.11",
     "vue-slider-component": "3.0.15",
     "vue-template-compiler": "2.5.18",
-    "vuetify": "1.5.3"
+    "vuetify": "1.5.3",
+    "vuex": "^3.5.1"
```
Haroenv, line 111:
```suggestion
    "vuex": "3.5.1"
```

yarn.lock
```diff
@@ -15676,6 +15676,11 @@ vuetify@1.5.3:
   resolved "https://registry.yarnpkg.com/vuetify/-/vuetify-1.5.3.tgz#4bdc0a72ade1cd3521dc0c08a363ea40b914443e"
   integrity sha512-4V3i4HwKdt//1vAEiex3DpHsLwX3oHCIx/gpukSYvBGJUgHILUZ5JOEOAQJ1yvMV1w3P+dsmvlmRdthjci5VRg==
 
+vuex@^3.5.1:
```
Haroenv, line 15679:
```suggestion
vuex@3.5.1:
```

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -262,6 +263,64 @@ Array [
...
+      const searchClient = createFakeClient();
+
+      Vue.use(Vuex);
+
+      const store = new Vuex.Store({
+        state: {
+          count: 0,
+        },
+        mutations: {
+          increment(state) {
```
francoischalifour, line 277: We don't use this in the test. Do we need it?

Haroenv: ah yes, good point! I copied from my experiments earlier, but we don't need a full fledged store

francoischalifour (approved):

eunjae-lee (approved):

PR #865, opened 2020-09-23
feat(ssr): forward propsData to recreated component

Haroenv:
propsData isn't valid as argument to `.extend`, but you don't need to call `new Vue(comp)`, it's the same as calling `new comp({propsData...})`


eunjae-lee (approved):
Looks good to me

PR #866, opened 2020-09-24
 fix(esm): fix tree shaking on ESM build 

yannickcr:
Unlike what is described in the documentation ( https://www.algolia.com/doc/guides/building-search-ui/going-further/improve-performance/vue/#optimize-build-size ) tree shaking does not works on Vue InstantSearch.

A simple way to test tree shaking is to try compiling the following file using `webpack`:

```js
// src/index.js
import Vue from 'vue';
import { AisInstantSearch, AisSearchBox } from 'vue-instantsearch';

Vue.component(AisInstantSearch.name, AisInstantSearch);
Vue.component(AisSearchBox.name, AisSearchBox);

new Vue({
  el: '#app',
  render: h => h()
})
```
```bash
$ npm i vue vue-instantsearch vue-server-renderer webpack webpack-cli
$ webpack         
Hash: 51b8ee76731f678ea519
Version: webpack 4.44.2
Time: 4815ms
Built at: 2020-09-23 15:20:27
  Asset     Size  Chunks                    Chunk Names
main.js  449 KiB       0  [emitted]  [big]  main
Entrypoint main [big] = main.js
 [3] (webpack)/buildin/global.js 472 bytes {0} [built]
[36] ./src/index.js + 70 modules 364 KiB {0} [built]
     | ./src/index.js 243 bytes [built]
     |     + 70 hidden modules
    + 55 hidden modules
```

As you can see the resulting bundle has a total size of 449 KiB, indicating that all vue-instantsearch was added to the bundle.

This is because, unlike the other flavors, we are using Rollup here to compile our ESM build and it will bundle all our files in a single one, making tree shaking impossible.

The solution here is to use the [`output.preserveModules`](https://rollupjs.org/guide/en/#outputpreservemodules) option to keep our existing file structure. I also had to upgrade Rollup to v1.32.1 to avoid an issue on filenames ( https://github.com/rollup/rollup/issues/2847 ).

One particularity of our project is that we are reading the `version` of `vue-instantsearch` from the `package.json` (for the agent), which makes Rollup recreate the complete directory structure in the `dist` directory. Not great but it has no impact on the usage of the library from a user POV (one solution to avoid this would be to avoid reading from the `package.json` entirely and use an environment variable instead).

Result of a compilation using `webpack` with this change:
```
$ webpack
Hash: b2199f20ebfd93466b0b
Version: webpack 4.44.2
Time: 725ms
Built at: 09/23/2020 3:51:55 PM
  Asset     Size  Chunks             Chunk Names
main.js  161 KiB       0  [emitted]  main
Entrypoint main = main.js
 [5] (webpack)/buildin/global.js 472 bytes {0} [built]
[34] ./test-treeshake.js + 38 modules 94.5 KiB {0} [built]
     | ./test-treeshake.js 246 bytes [built]
     | ./es/mixins/suit.js 292 bytes [built]
     | ./es/mixins/widget.js 1.82 KiB [built]
     | ./es/util/warn.js 117 bytes [built]
     | ./es/util/suit.js 206 bytes [built]
     | ./es/util/polyfills.js 506 bytes [built]
     | ./es/util/createInstantSearchComponent.js 1.34 KiB [built]
     |     + 32 hidden modules
[35] ./es/instantsearch.js 2.4 KiB [built]
[71] ./es/connectors/connectStateResults.js 512 bytes [built]
[72] ./es/mixins/panel.js 1.02 KiB [built]
[73] ./es/plugin.js 181 bytes [built]
[74] ./es/util/createServerRootMixin.js 3.23 KiB [built]
[75] ./es/util/parseAlgoliaHit.js 1.04 KiB [built]
[76] ./es/util/unescape.js 250 bytes [built]
[77] ./es/widgets.js 3.55 KiB [built]
    + 143 hidden modules
```

With the fix the bundle size is reduced to 161 KiB, tree shaking is now working correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
   "version": "3.2.0",
   "files": [
     "dist",
-    "src"
+    "src",
+    "es"
   ],
   "main": "dist/vue-instantsearch.common.js",
   "module": "es/src/instantsearch.js",
--- a/package.json
+++ b/package.json
@@ -103,10 +103,12 @@
     "vue-jest": "2.6.0",
     "vue-json-tree": "0.3.3",
     "vue-loader": "14.2.2",
+    "vue-router": "3.4.3",
     "vue-server-renderer": "^2.6.11",
     "vue-slider-component": "3.0.15",
     "vue-template-compiler": "2.5.18",
-    "vuetify": "1.5.3"
+    "vuetify": "1.5.3",
+    "vuex": "3.5.1"
   },
   "bundlesize": [
     {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "sideEffects": false,
   "repository": "https://github.com/algolia/vue-instantsearch",
   "scripts": {
-    "build": "rollup -c",
+    "build": "rm -rf dist es && rollup -c",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "storybook:build": "build-storybook -c .storybook -o website/stories",
     "examples:build": "yarn build && ./examples/build.sh",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "src"
   ],
   "main": "dist/vue-instantsearch.common.js",
-  "module": "es/instantsearch.js",
+  "module": "es/src/instantsearch.js",
   "sideEffects": false,
   "repository": "https://github.com/algolia/vue-instantsearch",
   "scripts": {
@@ -87,7 +87,7 @@
     "jest-watch-typeahead": "0.2.1",
     "lodash": "4.17.19",
     "prettier": "1.14.3",
-    "rollup": "1.21.0",
+    "rollup": "1.32.1",
     "rollup-plugin-buble": "0.19.6",
     "rollup-plugin-commonjs": "10.0.1",
     "rollup-plugin-filesize": "9.0.0",
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "jest-watch-typeahead": "0.2.1",
     "lodash": "4.17.19",
     "prettier": "1.14.3",
-    "rollup": "1.16.7",
+    "rollup": "1.21.0",
     "rollup-plugin-buble": "0.19.6",
     "rollup-plugin-commonjs": "10.0.1",
     "rollup-plugin-filesize": "9.0.0",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "src"
   ],
   "main": "dist/vue-instantsearch.common.js",
-  "module": "dist/vue-instantsearch.esm.js",
+  "module": "es/instantsearch.js",
   "sideEffects": false,
   "repository": "https://github.com/algolia/vue-instantsearch",
   "scripts": {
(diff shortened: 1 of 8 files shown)


Haroenv:
We need to make sure the script either still creates `dist/vue-instantsearch.esm.js` or removes it from the folder, it might be a breaking change / accidentally publishing old versions

yannickcr:
`dist/vue-instantsearch.esm.js` no longer exists since this file was not working for tree shaking (the Rollup configuration does not create it anymore).

I'm afraid I did not understand what you mean by removing it from the folder, and the risk of publishing an old version :thinking:

Haroenv:
the dist folder doesn't seem to be cleared, if you still have a `dist/vue-instantsearch.esm.js` file, it will still exist

Kocal:
I think @Haroenv means that if on your machine you still have a `dist` folder containing old versions, and you publish the package from your machine, then the `dist` folder will be published aswell but it don't have too.

yannickcr:
I see, I was thinking the release process was handled by ship.js on this project so this issue would not occur. I'll add a cleaning step.

Haroenv (changes requested):
if someone imported as `vue-instantsearch/dist/vue-instantsearch.esm.js`, that will now be broken, but I don't think it's a real problem. In the future I'd love an e2e test for tree shaking.

I've tried this out in a vue application, and it doesn't seem to make a difference, the whole Vue InstantSearch code was in there, from unused imports too

yannickcr:
@Haroenv  Can you share your test application ?

package.json
```diff
@@ -22,11 +22,11 @@
     "src"
```
Haroenv, line 22:
```suggestion
    "src",
    "es"
```

Haroenv: this was the issue that opted out of the tree shaking 👍

Haroenv (approved):
with these changes it works for me now :)

Haroenv:
my app was a fork of https://codesandbox.io/s/multiple-algolia-apps-issue-forked-brm17?file=/src/views/AppleProducts.vue with `@haroenv/not-vue-instant-search` (published from here)

rollup.config.js
```diff
@@ -28,26 +44,26 @@ export default [
...
       },
+    ],
+    plugins: [...plugins],
+  },
+  {
+    input: 'src/instantsearch.js',
+    external: [
+      'algoliasearch-helper',
+      'instantsearch.js/es',
+      'instantsearch.js/es/connectors',
```
eunjae-lee, line 55: Do we need this line?

Haroenv: we import directly from es/connectors, so yes as far as I can tell. It's the same as before in any case

yannickcr: Yes, I did not touch this. It is the same configuration as before.

eunjae-lee: Yeah I saw it was the same before, but got curious if `/connectors` has any effect while we already have `instantsearch.js/es`. And we import things like `instantsearch.js/es/lib/xxx` but they're not here, either.

I removed `/connectors` and built. And the size of cjs and esm seem to be the same.
But I'm not sure about anything since I haven't dug in depth. Not a blocker at all here.

eunjae-lee (approved):
Looks good to me.

PR #871, opened 2020-09-29
fix(panel): react to changes on initial render

Haroenv:
you can see this when there's a query which is applied from the URL (sorry, not relevant in the storybook) which makes the `canRefine` (`ais-Panel--noRefinement`) different than the default (`true`)


Haroenv:
relevant vue thread: https://forum.vuejs.org/t/watchers-not-triggered-on-initialization/12475/2

eunjae-lee (approved):
Looks good to me

PR #875, opened 2020-10-01
fix: fix crash in mapStateToCanRefine when state is empty

### LINKED ISSUE #874: Cannot read property 'count' of undefined on refinement components
**Bug 🐞**

Console error is shown "Cannot read property 'count' of undefined" for every refinement component since last version. Component works as expected though.

Error in callback for watcher "state": "TypeError: Cannot read property 'count' of undefined"
 
https://codesandbox.io/s/nice-ives-16unm?file=/src/App.vue

No error should show up in the console.

No.

-

v3.4.1

yannickcr:
Since #871 `mapStateToCanRefine` can be called with an empty object. It was triggering some crashes if we where trying to access some deep properties from `state` in it.

This fix address two things on this:
* Add some checks for the existence of each level in the `state` object.
* Cast the return value so are sure to return a `Boolean`.

TODO: Tests are missing, I'm not sure on how to test this, any idea @Haroenv ?

fixes #874

### DIFF
--- a/src/components/Breadcrumb.vue
+++ b/src/components/Breadcrumb.vue
@@ -58,7 +58,7 @@ export default {
   mixins: [
     createWidgetMixin({ connector: connectBreadcrumb }),
     createPanelConsumerMixin({
-      mapStateToCanRefine: state => state.canRefine,
+      mapStateToCanRefine: state => Boolean(state.canRefine),
     }),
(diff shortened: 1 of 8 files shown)


Haroenv:
I have this too, but was still figuring out the tests...

Haroenv:
#876 is more complete, so let's continue there

PR #876, opened 2020-10-01
fix(panel): safely access state in mapStateToCanRefine

### LINKED ISSUE #874: Cannot read property 'count' of undefined on refinement components

Haroenv:
Most of the widgets were assuming the shape of state is always "right", but since #871 it is also called with an empty object. This PR does the following things:

- changes all widgets to work correctly with empty object
- makes sure always a boolean is returned from mapStateToCanRefine to make tests easier
- make tests of mapStateToCanRefine consistent, as well as adding a unit test to each of the methods using `{}` as state

closes #875 
fixes #874

### DIFF
--- a/src/components/__tests__/Menu.js
+++ b/src/components/__tests__/Menu.js
@@ -332,7 +332,8 @@ it('calls the Panel mixin with `canRefine`', () => {
       canRefine: false,
     },
   });
-     expect(mapStateToCanRefine()).toBe(false); 
+  expect(mapStateToCanRefine()).toBe(false);
+
   expect(wrapper.vm.mapStateToCanRefine({})).toBe(false);
 });
 


src/components/__tests__/__snapshots__/RefinementList.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`allows search bar classes override when it's searchable 1`] = `
 
-<div class="ais-RefinementList ais-RefinementList--noRefinement">
+<div class="ais-RefinementList">
```
Haroenv, line 5: this is because the test used to be wrong (didn't pass `canRefine` in defaultState)

src/components/__tests__/Menu.js
```diff
@@ -313,27 +313,27 @@ describe('default render', () => {
...
-    });
+  expect(mapStateToCanRefine()).toBe(true);
 
-    expect(mapStateToCanRefine()).toBe(false);
+  wrapper.setData({
+    state: {
+      canRefine: false,
+    },
   });
+
```
eunjae-lee, line 335:
```suggestion
     expect(mapStateToCanRefine()).toBe(false); 
```

I guess this was deleted by accident?

eunjae-lee (approved):
Looks good to me 🙌

src/components/__tests__/Menu.js
Haroenv: yep, thanks!

yannickcr (approved):

src/components/__tests__/Menu.js
```diff
@@ -313,27 +313,27 @@ describe('default render', () => {
...
+  expect(mapStateToCanRefine()).toBe(true);
 
-    expect(mapStateToCanRefine()).toBe(false);
+  wrapper.setData({
+    state: {
+      canRefine: false,
+    },
   });
+     expect(mapStateToCanRefine()).toBe(false); 
+  expect(wrapper.vm.mapStateToCanRefine({})).toBe(false);
```
Haroenv, line 336:
```suggestion
  expect(mapStateToCanRefine()).toBe(false);

  expect(wrapper.vm.mapStateToCanRefine({})).toBe(false);
```

PR #898, opened 2020-12-16
fix(ssr): forward slots of the AisInstantSearchSsr component

Haroenv:
The clone created in findResultsState also manually needs to forward slots, or the data won't be available.

### DIFF
--- a/src/util/createServerRootMixin.js
+++ b/src/util/createServerRootMixin.js
@@ -90,6 +90,7 @@ function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
 
         // https://stackoverflow.com/a/48195006/3185307
         app.$slots = componentInstance.$slots;
+        app.$scopedSlots = componentInstance.$scopedSlots;
 
         app.$options.serverPrefetch = [];
 
--- a/src/util/createServerRootMixin.js
+++ b/src/util/createServerRootMixin.js
@@ -88,6 +88,9 @@ function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
           propsData: componentInstance.$options.propsData,
         });
 
+        // https://stackoverflow.com/a/48195006/3185307
+        app.$slots = componentInstance.$slots;
+
         app.$options.serverPrefetch = [];
 
         app.instantsearch.helper = helper;
--- a/src/util/createServerRootMixin.js
+++ b/src/util/createServerRootMixin.js
@@ -90,7 +90,7 @@ function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
 
         // https://stackoverflow.com/a/48195006/3185307
         app.$slots = componentInstance.$slots;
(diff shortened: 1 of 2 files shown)


Haroenv:
A demo is available here: https://github.com/algolia/vue-instantsearch/tree/demo/slot-component/examples/nuxt

eunjae-lee:
```
yarn build && (cd examples/nuxt && yarn && yarn dev)
```

image

I guess it happens [here](https://github.com/algolia/vue-instantsearch/blob/demo/slot-component/src/util/createServerRootMixin.js#L166:L166).

Did I do it wrong?

Haroenv:
you possibly still have an old version of the package installed somehow (maybe nuxt cache or something like that), I had it as well, but not sure what causes the old version to stay

src/util/createServerRootMixin.js
```diff
@@ -88,6 +88,9 @@ function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
           propsData: componentInstance.$options.propsData,
         });
 
+        // https://stackoverflow.com/a/48195006/3185307
+        app.$slots = componentInstance.$slots;
```
Haroenv, line 92:
```suggestion
        app.$slots = componentInstance.$slots;
        app.$scopedSlots = componentInstance.$scopedSlots;
```

eunjae-lee (commented):
I think it makes sense but it didn't work when I visited `/` first and clicked `/search-2` with the following error:
image

And it'd be nice to have a test case to cover this.

ntraykov:
I am not allowed to send more messages in the forum for today. That's why I am continuing here.

This is how I've implemented it on my side: https://codesandbox.io/s/competent-darkness-93vc4?file=/components/InstantSearch.vue

Unfortunately, it gives Bad Gateway error, which also happens in the original sandbox from the examples: https://codesandbox.io/s/github/algolia/vue-instantsearch/tree/master/examples/nuxt

ntraykov:
Ok, If ais-search-box is a direct child of ais-instant-search-ssr, there is no problem. Like this:

```
<ais-instant-search-ssr>
    <ais-search-box />
</ais-instant-search-ssr>
```

Haroenv:
sandbox for this version @eunjae-lee https://codesandbox.io/s/kind-dream-x7zk1?file=/pages/search.vue

eunjae-lee:
https://codesandbox.io/s/frosty-shadow-lm50t?file=/pages/search-2.vue

It seems to work except for one thing that might be related to [this](https://github.com/algolia/vue-instantsearch/pull/898/files#diff-ac44dbb445b26102c11a145082de41832504ef96bb746a1a05387f759c3d8909R170).

- `/` -> `/search`: works
- `/` -> `/search-2`: works
- between `/search` and `/search-2`: doesn't work

eunjae-lee (approved):

PR #922, opened 2021-03-04
 feat(ais-dynamic-widgets): add implementation

Haroenv:
**Summary**

Adds a new widget _ais-experimental-dynamic-widgets_ that can be used to conditionally render other widgets. This condition is based on the search results, by default results.facetOrdering.facet.values, but can be overridden using transformItems

**Result**

```vue
<template>
  <ais-experimental-dynamic-widgets :transform-items="transformItems">
    <ais-refinement-list attribute="test1" />
    <ais-menu attribute="test2" />
    <ais-panel>
      <ais-hierarchical-menu :attributes="hierarchicalAttributes" />
    </ais-panel>
  </ais-experimental-dynamic-widgets>
</template>

<script>
export default {
  data() {
    return {
      hierarchicalAttributes: ["test3", "unused"],
    };
  },
  methods(_attributes, { results }) {
    // modified if you want to use userData instead of facet ordering
    return results.userData[0].ordering;
  },
};
</script>
```

See also:

- https://github.com/algolia/instantsearch.js/pull/4687
- DX-1665

### DIFF
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -37,7 +37,7 @@ export default {
   props: {
     transformItems: {
       type: Function,
-      required: true,
+      default: undefined,
     },
   },
   render(createElement) {
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -59,7 +59,8 @@ export default {
 
     // by default, render everything, but hidden so that the routing doesn't disappear
     if (!this.state) {
-      const allComponents = Array.from(components.values());
+      const allComponents = [];
+      components.forEach(component => allComponents.push(component));
 
       return createElement(
         'div',
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -7,10 +7,7 @@ function getVueAttribute(vnode) {
     if (vnode.componentOptions.propsData.attribute) {
       return vnode.componentOptions.propsData.attribute;
     }
-    if (
-      vnode.componentOptions &&
-      Array.isArray(vnode.componentOptions.propsData.attributes)
-    ) {
+    if (Array.isArray(vnode.componentOptions.propsData.attributes)) {
       return vnode.componentOptions.propsData.attributes[0];
     }
   }
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -58,8 +58,7 @@ export default {
 
     // by default, render everything, but hidden so that the routing doesn't disappear
     if (!this.state) {
-      const allComponents = [];
-      components.forEach(vnode => allComponents.push(vnode));
+      const allComponents = Array.from(components.values());
 
       return createElement(
         'div',
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -2,13 +2,14 @@ import { createWidgetMixin } from '../mixins/widget';
 import { EXPERIMENTAL_connectDynamicWidgets } from 'instantsearch.js/es/connectors';
 import { createSuitMixin } from '../mixins/suit';
 
-function getVueAttribute(vnode) {
-  if (vnode.componentOptions && vnode.componentOptions.propsData) {
-    if (vnode.componentOptions.propsData.attribute) {
-      return vnode.componentOptions.propsData.attribute;
+function getWidgetAttribute(vnode) {
+  const props = vnode.componentOptions && vnode.componentOptions.propsData;
+  if (props) {
+    if (props.attribute) {
+      return props.attribute;
     }
-    if (Array.isArray(vnode.componentOptions.propsData.attributes)) {
-      return vnode.componentOptions.propsData.attributes[0];
+    if (Array.isArray(props.attributes)) {
+      return props.attributes[0];
     }
   }
 
@@ -20,7 +21,7 @@ function getVueAttribute(vnode) {
   if (Array.isArray(children)) {
     // return first child with a truthy attribute
     return children.reduce(
-      (acc, curr) => acc || getVueAttribute(curr),
+      (acc, curr) => acc || getWidgetAttribute(curr),
       undefined
     );
   }
@@ -43,7 +44,7 @@ export default {
   render(createElement) {
     const components = new Map();
     (this.$slots.default || []).forEach(vnode => {
-      const attribute = getVueAttribute(vnode);
+      const attribute = getWidgetAttribute(vnode);
       if (attribute) {
         components.set(
           attribute,
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -1,0 +1,89 @@
+import { createWidgetMixin } from '../mixins/widget';
+import { EXPERIMENTAL_connectDynamicWidgets } from 'instantsearch.js/es/connectors';
+import { createSuitMixin } from '../mixins/suit';
+
+function getVueAttribute(vnode) {
+  if (vnode.componentOptions && vnode.componentOptions.propsData) {
+    if (vnode.componentOptions.propsData.attribute) {
+      return vnode.componentOptions.propsData.attribute;
+    }
+    if (
+      vnode.componentOptions &&
+      Array.isArray(vnode.componentOptions.propsData.attributes)
+    ) {
+      return vnode.componentOptions.propsData.attributes[0];
+    }
+  }
+
+  const children =
+    vnode.componentOptions && vnode.componentOptions.children
+      ? vnode.componentOptions.children
+      : vnode.children;
+
+  if (Array.isArray(children)) {
+    // return first child with a truthy attribute
+    return children.reduce(
+      (acc, curr) => acc || getVueAttribute(curr),
+      undefined
+    );
+  }
+
+  return undefined;
+}
+
+export default {
+  name: 'AisExperimentalDynamicWidgets',
+  mixins: [
+    createWidgetMixin({ connector: EXPERIMENTAL_connectDynamicWidgets }),
+    createSuitMixin({ name: 'DynamicWidgets' }),
+  ],
+  props: {
+    transformItems: {
+      type: Function,
+      required: true,
+    },
+  },
+  render(createElement) {
+    const components = new Map();
+    (this.$slots.default || []).forEach(vnode => {
+      const attribute = getVueAttribute(vnode);
+      if (attribute) {
+        components.set(
+          attribute,
+          createElement(
+            'div',
+            { key: attribute, class: [this.suit('widget')] },
+            [vnode]
+          )
+        );
+      }
+    });
+
+    // by default, render everything, but hidden so that the routing doesn't disappear
+    if (!this.state) {
+      const allComponents = [];
+      components.forEach(vnode => allComponents.push(vnode));
+
+      return createElement(
+        'div',
+        { attrs: { hidden: true }, class: [this.suit()] },
+        allComponents
+      );
+    }
+
+    return createElement(
+      'div',
+      { class: [this.suit()] },
+      this.state.attributesToRender.map(attribute => components.get(attribute))
+    );
+  },
+  computed: {
+    widgetParams() {
+      return {
+        transformItems: this.transformItems,
+        // we do not pass "widgets" to the connector, since Vue is in charge of rendering
+        widgets: [],
+      };
+    },
+  },
+};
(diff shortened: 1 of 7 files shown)


src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,89 @@
...
+  if (vnode.componentOptions && vnode.componentOptions.propsData) {
+    if (vnode.componentOptions.propsData.attribute) {
+      return vnode.componentOptions.propsData.attribute;
+    }
+    if (
+      vnode.componentOptions &&
+      Array.isArray(vnode.componentOptions.propsData.attributes)
+    ) {
+      return vnode.componentOptions.propsData.attributes[0];
+    }
```
shortcuts, line 15:
```suggestion
    if (Array.isArray(vnode.componentOptions.propsData.attributes)) {
      return vnode.componentOptions.propsData.attributes[0];
    }
```

The first condition doesn't seem necessary

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,89 @@
...
+    // by default, render everything, but hidden so that the routing doesn't disappear
+    if (!this.state) {
+      const allComponents = [];
+      components.forEach(vnode => allComponents.push(vnode));
+
+      return createElement(
+        'div',
+        { attrs: { hidden: true }, class: [this.suit()] },
+        allComponents
+      );
```
shortcuts, line 71: Is there a reason not to spread `components`, or directly pass a `components.map(...)`?

Haroenv: components is a Map, so it can't be mapped directly

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,86 @@
...
+            [vnode]
+          )
+        );
+      }
+    });
+
+    // by default, render everything, but hidden so that the routing doesn't disappear
+    if (!this.state) {
+      const allComponents = [];
+      components.forEach(vnode => allComponents.push(vnode));
```
francoischalifour, line 62: Any reasons to not use a single expression?

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,178 @@
...
+
+  expect(wrapper.html()).toMatchInlineSnapshot(`
+
+<div class="ais-DynamicWidgets">
+</div>
+
+`);
+});
+
+it('renders attributesToRender 1', () => {
```
francoischalifour, line 97: What does this test mean? Can its name reflect the actual scenario?

francoischalifour: A suggestion:

```suggestion
it('renders only the components with attributes that belong to `attributesToRender`', () => {
```

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,178 @@
...
+  });
+  expect(wrapper.html()).toMatchInlineSnapshot(`
+
+<div class="ais-DynamicWidgets">
+</div>
+
+`);
+});
+
+it('renders nothing with empty items', () => {
```
francoischalifour, line 69: Perhaps we can add a test with a panel header and empty items?

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,86 @@
...
+        { attrs: { hidden: true }, class: [this.suit()] },
+        allComponents
+      );
+    }
+
+    return createElement(
+      'div',
+      { class: [this.suit()] },
+      this.state.attributesToRender.map(attribute => components.get(attribute))
+    );
```
francoischalifour, line 75: We don't seem to test the dynamic nature of this widget, represented by these lines.

src/components/DynamicWidgets.js
Haroenv, line 62:
```suggestion
      const allComponents = [...components.values()];
```

to check whether this works as an alternative that's shorter

Haroenv: not possible with this babel config

src/components/DynamicWidgets.js
Haroenv: was really hard to get Vue to correctly recognise a change in data internally, but it works now: 2002f4b

src/components/DynamicWidgets.js
Haroenv: d9779b5

src/components/__tests__/DynamicWidgets.js
Haroenv: 2002f4b

src/components/__tests__/DynamicWidgets.js
Haroenv: 2002f4b

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,85 @@
...
+            { key: attribute, class: [this.suit('widget')] },
+            [vnode]
+          )
+        );
+      }
+    });
+
+    // by default, render everything, but hidden so that the routing doesn't disappear
+    if (!this.state) {
+      const allComponents = Array.from(components.values());
```
francoischalifour, line 61: Is using `Map.prototype.values()` OK given our browser support for VIS? ([see browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values#browser_compatibility))

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,85 @@
...
+
+function getVueAttribute(vnode) {
+  if (vnode.componentOptions && vnode.componentOptions.propsData) {
+    if (vnode.componentOptions.propsData.attribute) {
+      return vnode.componentOptions.propsData.attribute;
+    }
+    if (Array.isArray(vnode.componentOptions.propsData.attributes)) {
+      return vnode.componentOptions.propsData.attributes[0];
+    }
+  }
```
francoischalifour, line 13: That's easier to read/understand. Wdyt?

```suggestion
  const props = vnode.componentOptions && vnode.componentOptions.propsData;

  if (props) {
    if (props.attribute) {
      return props.attribute;
    }
    if (Array.isArray(props.attributes)) {
      return props.attributes[0];
    }
  }
```

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,85 @@
+import { createWidgetMixin } from '../mixins/widget';
+import { EXPERIMENTAL_connectDynamicWidgets } from 'instantsearch.js/es/connectors';
+import { createSuitMixin } from '../mixins/suit';
+
+function getVueAttribute(vnode) {
```
francoischalifour, line 5:
```suggestion
function getWidgetAttribute(vnode) {
```

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,85 @@
...
+          createElement(
+            'div',
+            { key: attribute, class: [this.suit('widget')] },
+            [vnode]
+          )
+        );
+      }
+    });
+
+    // by default, render everything, but hidden so that the routing doesn't disappear
```
francoischalifour, line 59: Won't this behavior result in an unexpected UI flash were some widget will appear and then get removed because computed based on `state.attributesToRender`?

What if we don't render anything when we don't have a state?

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,409 @@
...
+});
+
+it('renders attributesToRender (panel)', () => {
+  const localVue = createLocalVue();
+
+  localVue.use(plugin);
+
+  __setState({
+    attributesToRender: ['test2'],
+  });
```
francoischalifour, line 236: Can we created a similar test with this? (to ensure that the panel component is not added to the DOM)

```js
  __setState({
    attributesToRender: [],
  });
```

src/components/DynamicWidgets.js
Haroenv: note that the widgets we render initially are hidden, so this won't cause a flash, but possibly slightly delayed visuals compared to server side rendering (which solves this, as the widget information is available on that server + initial frontend render)

francoischalifour: Right, I think I rather meant: won't this results in an unnecessary search parameters computation because it mounts widgets that will likely get removed?

Why don't we render nothing?

src/components/DynamicWidgets.js
Haroenv: right, only Map is IE11 compatible, what did you mean as alternative in https://github.com/algolia/vue-instantsearch/pull/922#discussion_r638579981 then?

src/components/DynamicWidgets.js
Haroenv: if we render nothing, the parameters of the URL can't be applied unfortunately, this was the best workaround I found

src/components/DynamicWidgets.js
```diff
@@ -0,0 +1,87 @@
...
+    vnode.componentOptions && vnode.componentOptions.children
+      ? vnode.componentOptions.children
+      : vnode.children;
+
+  if (Array.isArray(children)) {
+    // return first child with a truthy attribute
+    return children.reduce(
+      (acc, curr) => acc || getWidgetAttribute(curr),
+      undefined
+    );
```
eunjae-lee, line 26:
```suggestion
    for (let child of children) {
      const attribute = getWidgetAttribute(child);
      if (attribute) {
        return attribute;
      }
    }
```

kinda old school but easier to read along with that comment, IMO.

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,420 @@
...
+  </div>
+</div>
+
+`);
+});
+
+it('renders nothing without children', () => {
+  __setState({
+    attributesToRender: [],
+  });
```
eunjae-lee, line 103: needed in this case?

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,420 @@
...
+  expect(wrapper.html()).toMatchInlineSnapshot(`
+
+<div class="ais-DynamicWidgets">
+</div>
+
+`);
+});
+
+it('renders nothing with empty items', () => {
```
eunjae-lee, line 118:
```suggestion
it('renders nothing with empty attributesToRender', () => {
```

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,420 @@
...
+    localVue,
+    propsData: {
+      transformItems: items => items,
+    },
+    slots: {
+      default: `<ais-refinement-list attribute="test1"/>`,
+    },
+    stubs: {
+      'ais-refinement-list': MockRefinementList,
+      'ais-menu': MockMenu,
```
eunjae-lee, line 137:
```suggestion
```

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,420 @@
...
+    >
+    </div>
+  </div>
+</div>
+
+`);
+
+  attributesToRender = ['test3'];
+
+  wrapper.vm.$forceUpdate();
```
eunjae-lee, line 368: can you try 

```js
await wrapper.setData({ attributesToRender: ['test3'] });
```

not sure if it works in this VTU version, though.

src/components/DynamicWidgets.js
Haroenv: for of needs transpilation, what's wrong with the reduce?

eunjae-lee: didn't think of transpilation. nothing wrong with the reduce, just it took me several seconds to follow where it's going.

src/components/__tests__/DynamicWidgets.js
Haroenv: Yes, that's what this test does, check if there's nothing in the return from attributesToRender, it doesn't render the initial widgets hidden anymore

src/components/__tests__/DynamicWidgets.js
Haroenv: doesn't seem to work, i think because of the mock

Haroenv: doesn't seem to work, i think because of the mock

src/components/__tests__/DynamicWidgets.js
```diff
@@ -0,0 +1,420 @@
...
+it('renders nothing without children', () => {
+  __setState({
+    attributesToRender: [],
+  });
+
+  const wrapper = mount(DynamicWidgets, {
+    propsData: {
+      transformItems: items => items,
+    },
+  });
```
eunjae-lee, line 109: If I understand correctly, this test is about not providing any children `slots: { default: ... }` and seeing if it renders an empty container. So setting attributesToRender to an empty array isn't relevant. With this, we don't know the reason why the empty container is there is because of attributesToRender: [] or no children.

Haroenv: solved in c0d1823

eunjae-lee (approved):
let's go!

PR #931, opened 2021-03-15
fix(examples): fix examples to use vue-instantsearch from npm

yannickcr:
The examples are currently using `vue-instantsearch` from the repository.

While it is great to run the examples with the latest changes it also prevent the examples from working in CodeSandbox. This is a big issue since they are linked in the documentation (https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/vue/).

This PR changes this to use the `vue-instantsearch` package from npm.

### DIFF
--- a/examples/build.sh
+++ b/examples/build.sh
@@ -14,7 +14,7 @@ function build_example {
     name=$(basename "$dir")
     echo "building example: $name"
     cd $name
-    if [[ "$name" != "nuxt" && "$name" != "ssr" ]]; then
+    if [[ "$name" != "ssr" ]]; then
       yarn
       yarn build
       mkdir -p ../../website/examples/$name
--- a/examples/build.sh
+++ b/examples/build.sh
@@ -16,7 +16,7 @@ function build_example {
     cd $name
     if [[ "$name" != "ssr" ]]; then
       yarn
-      yarn build
+      [[ "$name" == "nuxt" ]] && yarn generate || yarn build
       mkdir -p ../../website/examples/$name
       cp -R dist/* ../../website/examples/$name
     else
--- a/examples/default-theme/package.json
+++ b/examples/default-theme/package.json
@@ -12,7 +12,7 @@
     "instantsearch.css": "^7.1.1",
     "instantsearch.js": "^4.2.0",
     "vue": "^2.6.7",
-    "vue-instantsearch": "^3.0.0-beta.0"
+    "vue-instantsearch": "3.5.0"
   },
   "devDependencies": {
     "@vue/cli-plugin-babel": "^3.4.1",
--- a/examples/default-theme/package.json
+++ b/examples/default-theme/package.json
@@ -12,7 +12,7 @@
     "instantsearch.css": "^7.1.1",
     "instantsearch.js": "^4.2.0",
     "vue": "^2.6.7",
-    "vue-instantsearch": "3.5.0"
+    "vue-instantsearch": "^3.0.0-beta.0"
   },
   "devDependencies": {
     "@vue/cli-plugin-babel": "^3.4.1",
--- a/examples/e-commerce/package.json
+++ b/examples/e-commerce/package.json
@@ -12,7 +12,7 @@
     "classnames": "2.2.6",
     "instantsearch.js": "^4.5.0",
     "vue": "2.6.10",
-    "vue-instantsearch": "3.5.0",
+    "vue-instantsearch": "file:../../",
     "vue-slider-component": "3.0.32"
   },
   "devDependencies": {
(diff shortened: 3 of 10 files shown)


yannickcr:
@Haroenv I was not able to make the `ssr` example to work, it seems that this change https://github.com/algolia/vue-instantsearch/pull/865 broke something in the example. (`Vue.component(Object.assign({}, componentInstance.$options, options));` returns `undefined`).

Any idea?

examples/e-commerce/package.json
```diff
@@ -12,7 +12,7 @@
     "classnames": "2.2.6",
     "instantsearch.js": "^4.5.0",
     "vue": "2.6.10",
-    "vue-instantsearch": "file:../../",
+    "vue-instantsearch": "3.5.0",
```
Haroenv, line 15: changing this means that it doesn't run the local files anymore for e2e tests

Haroenv:
> @Haroenv I was not able to make the `ssr` example to work, it seems that this change #865 broke something in the example. (`Vue.component(Object.assign({}, componentInstance.$options, options));` returns `undefined`).
> 
> Any idea?

How do I reproduce that @yannickcr ?

examples/e-commerce/package.json
yannickcr: Right. But without this change examples does not works on codesandbox (and appears broken in the documentation).

We could make a separate e-commerce demo specifically for e2e tests, wdyt?

yannickcr:
@Haroenv in `examples/ssr/package.json`, upgrade `vue-instantsearch` to `3.5.0` and `vue`, `vue-server-renderer` and `vue-template-compiler` to `2.6.11`.

then:

```bash
yarn
yarn serve
```

When displaying the page you will get this error:

```js
TypeError: i is not a constructor
    at app.js:14792:832
    at async Promise.all (index 0)
```
The errors comes from https://github.com/algolia/vue-instantsearch/blob/218238bebfa5330b054c484a0697fcb01bee3f0d/src/util/createServerRootMixin.js#L81-L89

In our case `Extended` is `undefined`

examples/e-commerce/package.json
Haroenv: we definitely could, or could use the netlify instead of codesandbox on the documentation, as used to work

examples/default-theme/package.json
```diff
@@ -12,7 +12,7 @@
     "instantsearch.css": "^7.1.1",
     "instantsearch.js": "^4.2.0",
     "vue": "^2.6.7",
-    "vue-instantsearch": "3.5.0"
+    "vue-instantsearch": "^3.0.0-beta.0"
```
Haroenv: should this not be the latest?

examples/build.sh
```diff
@@ -14,9 +14,9 @@ function build_example {
     name=$(basename "$dir")
     echo "building example: $name"
     cd $name
-    if [[ "$name" != "nuxt" && "$name" != "ssr" ]]; then
+    if [[ "$name" != "ssr" ]]; then
       yarn
-      yarn build
+      [[ "$name" == "nuxt" ]] && yarn generate || yarn build
```
Haroenv, line 19: we could alias generate to build in the package.json so this isn't needed?

yannickcr: There is already a yarn build in the `package.json` for nuxt and I'm afraid it is used for something else.

yannickcr:
Tried to generate a static version of the example for deployment on Netlify (using the `target: static` setting) but for an unknown reason there is still an error on Netlify: "Uncaught (in promise) TypeError: "exports" is read-only"

https://nuxtjs.org/docs/2.x/deployment/netlify-deployment

examples/e-commerce/package.json
eunjae-lee: @Haroenv  I prefer having a codesandbox on the doc over the netlify because it enables people to modify it by themselves.
What if we modify the e2e script so that it adds `"vue-instantsearch": "file:../../"` on the fly?

Haroenv: that seems like a reasonable solution, but how come we don't have this issue in InstantSearch.js for example? There we also use file:../../ right? 

Alternatively setting up a yarn workspace here would also solve it I think. Do you want to take over this PR @eunjae-lee ?

eunjae-lee: Yeah, `yarn workspace` will definitely fix the issue. Haven't looked into why it doesn't happen in InstantSearch.js, though.
I'll take it over and will continue in the near future.

Haroenv: looks like https://github.com/algolia/instantsearch.js/blob/master/examples/e-commerce/package.json#L24-L26 is used in InstantSearch.js, but it also doesn't work in codesandbox: https://codesandbox.io/s/github/algolia/instantsearch.js/tree/master/examples/e-commerce

Issue #936, opened 2021-03-23
AisInstantSearchSsr does not work well with nuxt i18n translation functions


podlebar:
**Bug 🐞**

### What is the current behavior?
I use nuxt.js with <ais-instant-search-ssr> and for the translation of the whole website i use the official nuxt i18n module (https://i18n.nuxtjs.org/). 
The common function to translate strings with nuxt i18n is is `$t()`. When i place this somewhere inside the <ais-instant-search-ssr> component i get the error `"TypeError: Cannot read property '_t' of undefined"`. Also i use the <ais-refinement-list> component with `transform-items` where i translate the labels into the desired language but this leads to the same error.

### Make a sandbox with the current behavior

Not possible as true SSR with Nuxt is needed.. 

### What is the expected behavior?
The `$t()` function should be usable inside the <ais-instant-search-ssr> component without any ugly workaround.
My workaround is currently to re-render all <ais-refinement-list> components as soo the `mounted()`hook is called. 

### Does this happen only in specific situations?
No.. in al situations where i use a translation function inside a <ais-instant-search-ssr> component.

### What is the proposed solution?
Make `$t()` available in <ais-instant-search-ssr> or find a way to inject `$t()` and `$tc()` functions into <ais-instant-search-ssr>. (`$tc()` is used for pluralization.)

### What is the version you are using?
"nuxt": "^2.14.3"
"vue-instantsearch": "^3.5.0",
"algoliasearch": "^4.8.6",

Haroenv:
I can't find the source code easily of `$t`, but it might be https://github.com/nuxt-community/i18n-module/blob/master/src/index.js#L105 or related. Basically the code in Vue InstantSearch related to this is:

https://github.com/algolia/vue-instantsearch/blob/218238bebfa5330b054c484a0697fcb01bee3f0d/src/util/createServerRootMixin.js#L71-L96

It makes a copy of the current component to render intermediary, to get access to the widgets that are its children in `serverPrefetch`. We should open that up to be overridable so you can retrieve the $t function yourself without us necessarily having to make a patch in the library.

Similar issues we've fixed before: https://github.com/algolia/vue-instantsearch/pull/898 https://github.com/algolia/vue-instantsearch/pull/865 https://github.com/algolia/vue-instantsearch/pull/864 https://github.com/algolia/vue-instantsearch/pull/863

podlebar:
[...]

PR #937, opened 2021-03-24
feat(ssr): expose clone component

Haroenv:
related to #936, this would allow people to link items to the cloned component themselves

### DIFF
--- a/src/util/createServerRootMixin.js
+++ b/src/util/createServerRootMixin.js
@@ -40,7 +40,39 @@ function searchOnlyWithDerivedHelpers(helper) {
   });
 }
 
-function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
+function defaultCloneComponent(componentInstance) {
+  const options = {
+    serverPrefetch: undefined,
+    fetch: undefined,
+    _base: undefined,
+    name: 'ais-ssr-root-component',
+    // copy over global Vue APIs
+    router: componentInstance.$router,
+    store: componentInstance.$store,
+  };
+
+  const Extended = componentInstance.$vnode
+    ? componentInstance.$vnode.componentOptions.Ctor.extend(options)
+    : Vue.component(Object.assign({}, componentInstance.$options, options));
+
+  const app = new Extended({
+    propsData: componentInstance.$options.propsData,
+  });
+
+  // https://stackoverflow.com/a/48195006/3185307
+  app.$slots = componentInstance.$slots;
+  app.$root = componentInstance.$root;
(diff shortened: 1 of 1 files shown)


src/util/createServerRootMixin.js
```diff
@@ -68,32 +100,7 @@ function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
...
-          propsData: componentInstance.$options.propsData,
-        });
-
-        // https://stackoverflow.com/a/48195006/3185307
-        app.$slots = componentInstance.$slots;
-
-        app.$root = componentInstance.$root;
-
-        app.$options.serverPrefetch = [];
+        app = cloneComponent(componentInstance);
```
eunjae-lee, line 103: what do you think of cloneComponent vs (disclaimer: temp name) augmentApp, where

augmentApp = (app, componentInstance) => app

so that user can benefit from our default implementation or just ignore and create a fresh app instance based on componentInstance if they want.

Haroenv: app is the name of the component though here, not an "app", just `app = new Vue` is a common naming. Cloning the component isn't optional, it's more that they might want to copy more items than what we did in the default one (like i18n in the linked issue)

src/util/createServerRootMixin.js
```diff
@@ -235,7 +242,11 @@ function augmentInstantSearch(instantSearchOptions, searchClient, indexName) {
 }
 
 export function createServerRootMixin(instantSearchOptions = {}) {
-  const { searchClient, indexName } = instantSearchOptions;
+  const {
+    searchClient,
+    indexName,
+    $cloneComponent = defaultCloneComponent,
```
tkrugg, line 248: when does $cloneComponent gets passed to `instantSearchOptions`

Haroenv: could be a separate argument for Vue InstantSearch-specific arguments too, but didn't see much use yet. Can be changed!

src/util/createServerRootMixin.js
eunjae-lee: makes sense!

eunjae-lee (approved):
looks good to me.

src/util/createServerRootMixin.js
eunjae-lee: yeah we could see it as `type Options = InstantSearchOptions & VueInstantSearchOptions`, but no strong opinion, so I'm also okay with having it as in the second parameter.

tkrugg (approved):

PR #939, opened 2021-03-26
feat(middlewares): add middlewares prop to ais-instant-search

eunjae-lee:
## Summary

This PR adds `middlewares` prop to `<ais-instant-search>`.

<details>
<summary><i>ignore this. we're using `unuse` method.</i></summary>
Writing Vue code, users will expect `middlewares` prop to be reactive. Currently this PR handles newly added middlewares.

However, removed ones in the prop are not being handled at the moment.

1. In which case will people want to remove middlewares in the middle of their app's lifecycle?
2. We don't have an API to remove middleware from InstantSearch. We can manually call `middlewareToRemove.unsubscribe()`, but `instantSearchInstance` [still loops over](https://github.com/algolia/instantsearch.js/blob/master/src/lib/InstantSearch.ts#L580-L584) `this.middleware` to notify state changes, etc. So we need to have a proper API to support this reactive behavior in Vue.js (or other flavors).

Until we have an answer or usecase for (1), why don't we ship the current version first (not supporting removed middlewares)?

Do you think it will cause confusion?
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "algoliasearch-helper": "^3.1.0",
-    "instantsearch.js": "^4.16.1"
+    "instantsearch.js": "https://pkg.csb.dev/algolia/instantsearch.js/commit/59ed228e/instantsearch.js"
   },
   "peerDependencies": {
     "algoliasearch": ">= 3.32.0 < 5",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "algoliasearch-helper": "^3.1.0",
-    "instantsearch.js": "https://pkg.csb.dev/algolia/instantsearch.js/commit/59ed228e/instantsearch.js"
+    "instantsearch.js": "^4.20.0"
   },
   "peerDependencies": {
     "algoliasearch": ">= 3.32.0 < 5",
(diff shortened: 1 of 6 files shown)


francoischalifour:
Introducing an `unuse` method in InstantSearch.js shouldn't be much work IMO.

src/util/createInstantSearchComponent.js
```diff
@@ -34,6 +34,16 @@ export const createInstantSearchComponent = component =>
           // private InstantSearch.js API:
           this.instantSearchInstance._searchFunction = searchFunction;
         },
+        middlewares: {
+          immediate: true,
+          handler(next, prev) {
+            (next || [])
```
Haroenv, line 40: the previous middleware that no longer are in next middleware (opposite filter) should call a method on InstantSearch that removes the middleware reference from the list and calls its dispose, very similar to removeWidgets

package.json
```diff
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "algoliasearch-helper": "^3.1.0",
-    "instantsearch.js": "^4.16.1"
+    "instantsearch.js": "https://pkg.csb.dev/algolia/instantsearch.js/commit/59ed228e/instantsearch.js"
```
eunjae-lee, line 49: this is temporary. let's not merge this until we release it in InstantSearch.js.

Haroenv (approved):
lgtm once InstantSearch is released

Haroenv (approved):
let's go!

PR #949, opened 2021-04-12
feat(vue 3): support createElement for vue 3

eunjae-lee:
Since vue 3, it doesn't expose `createElement` in the `render` function. Instead we must import it from `vue`.

related: https://v3.vuejs.org/guide/migration/render-function-api.html#overview

Also, `this.$slots.default` has been changed to `this.$slots.default()` in vue 3 (https://v3.vuejs.org/api/instance-properties.html#slots)

### DIFF
--- a/src/components/InstantSearch.js
+++ b/src/components/InstantSearch.js
@@ -1,4 +1,5 @@
 import instantsearch from 'instantsearch.js/es';
+import { h as createElementVue3 } from 'vue';
 import { createInstantSearchComponent } from '../util/createInstantSearchComponent';
 import { warn } from '../util/warn';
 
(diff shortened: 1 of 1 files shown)


src/components/InstantSearch.js
```diff
@@ -87,16 +88,29 @@ export default createInstantSearchComponent({
...
   },
-  render(createElement) {
-    return createElement(
-      'div',
-      {
-        class: {
-          [this.suit()]: true,
-          [this.suit('', 'ssr')]: false,
+  render(createElementVue2) {
+    if (createElementVue3) {
```
Haroenv, line 92: could we not do `createElement = createEleementv3 || createElementv2`? This can make the output significantly smaller :)

Haroenv:
I think this PR is the same as https://github.com/algolia/vue-instantsearch/pull/855

eunjae-lee:
> I think this PR is the same as #855

didn't see that!

PR #951, opened 2021-04-13
feat(vue 3): support inject without `this`

eunjae-lee:
Vue 3 doesn't bind `this` when calling `default()` of a injected variable.

https://github.com/vuejs/vue-next/blob/540e26f49c09edf09b6a60ac2a978fdec52686bf/packages/runtime-core/src/apiInject.ts#L62-L64

### DIFF
--- a/src/mixins/widget.js
+++ b/src/mixins/widget.js
@@ -11,12 +11,7 @@ export const createWidgetMixin = ({ connector } = {}) => ({
         );
       },
     },
-    getParentIndex: {
-      from: '$_ais_getParentIndex',
-      default() {
-        return () => this.instantSearchInstance.mainIndex;
-      },
-    },
+    getProvidedParentIndex: '$_ais_getParentIndex',
   },
   data() {
     return {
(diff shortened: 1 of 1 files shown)


src/mixins/widget.js
```diff
@@ -72,6 +67,11 @@ Read more on using connectors: https://alg.li/vue-custom`
     },
   },
   methods: {
+    getParentIndex() {
```
eunjae-lee, line 70: not sure if it's the best approach, though

src/mixins/widget.js
```diff
@@ -11,12 +11,7 @@ export const createWidgetMixin = ({ connector } = {}) => ({
         );
```
eunjae-lee, line 11: this needs to be fixed too

src/mixins/widget.js
Haroenv: this makes sense, but would need to see it together with the InstantSearch change

eunjae-lee: @Haroenv can you explain what you mean here? ↑

PR #952, opened 2021-04-14
feat(vue 3): update rollup-plugin-vue

eunjae-lee:
This PR updates the build process into two, one for vue@2 and another for vue@3.

Although we do not use any specific Vue3-only APIs to render components, we'd better split the bundle outputs now for the future benefit.

Also, it's recommended to have different bundle outputs.

The order of the plugin has changed because of [this issue](https://github.com/vuejs/rollup-plugin-vue/issues/231).

## References

- https://github.com/vuejs/rollup-plugin-vue/issues/426#issuecomment-753841934
- https://github.com/vuejs/rollup-plugin-vue/issues/427#issuecomment-753867261

--- 

## How `vue-demi` works

It's a runtime dependency.

### in Vue 2

```js
import Vue from 'vue'

var isVue2 = true
var isVue3 = false
var Vue2 = Vue
var version = Vue.version

export {
  Vue,
  Vue2,
  isVue2,
  isVue3,
  version,
}

export const computed = undefined;
export const createApp = undefined;
export const createRef = undefined;
export const customRef = undefined;
...	// ↑ all the vue3-only APIs
```

### in Vue 3

```js
import * as Vue from 'vue'

var isVue2 = false
var isVue3 = true
var Vue2 = undefined

export * from 'vue'
export {
  Vue,
  Vue2,
  isVue2,
  isVue3,
}
```

So basically, the way we use this library is going to be

```js
import { somethingVue3OnlyAPI, isVue3, Vue2 } from 'vue-demi';

if (isVue3) {
  somethingVue3OnlyAPI();
} else {
  Vue2.somethingElse();
}
```


eunjae-lee:
Okay, at this point, a very simple example works fine on both vue2 and vue3.

[vue2-test.zip](https://github.com/algolia/vue-instantsearch/files/6585755/vue2-test.zip)
[vue3-test.zip](https://github.com/algolia/vue-instantsearch/files/6585756/vue3-test.zip)

Unfortunately they don't work on CodeSandbox because in this PR of Vue InstantSearch, we're aliasing vue-demi like this:

```
"vue-demi": "npm:@eunjae-lee/vue-demi@0.10.2"
```

I don't think CodeSandbox is handling `npm:` prefix. I've sent an inquiry.

One caveat of the way this branch exports the bundles is,

```js
import ... from 'vue-instantsearch'; // ← for vue2

import ... from 'vue-instantsearch/dist/vue3/es'; // ← for vue3
```

not sure how to shorten the import path for vue3, but we can do it in another PR.

src/components/InstantSearchSsr.js
```diff
@@ -1,6 +1,5 @@
 import { createInstantSearchComponent } from '../util/createInstantSearchComponent';
-import { isVue3 } from 'vue-demi';
-import { h } from 'vue';
+import { isVue3, h } from 'vue-demi';
```
Haroenv, line 2: can this h be used in vue 2? if so the conditional on isVue3 can go away

rollup.config.js
```diff
@@ -7,14 +8,24 @@ import { terser } from 'rollup-plugin-terser';
...
+}
+
 const processEnv = conf => ({
   // parenthesis to avoid syntax errors in places where {} is interpreted as a block
   'process.env': `(${JSON.stringify(conf)})`,
 });
 
+const vuePlugin = process.env.VUE_VERSION === 'vue3' ? vueV3 : vueV2;
+const outputDir =
+  process.env.VUE_VERSION === 'vue3' ? 'dist/vue3' : 'dist/vue2';
```
Haroenv, line 24: should the output dir for vue 2 stay the root?

package.json
```diff
@@ -115,11 +119,11 @@
   },
   "bundlesize": [
     {
-      "path": "./dist/vue-instantsearch.js",
+      "path": "./dist/vue2/umd/index.js",
```
Haroenv, line 122: we should track the vue 3 bundles too

package.json
```diff
@@ -98,7 +101,8 @@
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-replace": "2.1.0",
     "rollup-plugin-terser": "4.0.4",
-    "rollup-plugin-vue": "3.0.0",
+    "rollup-plugin-vue2": "npm:rollup-plugin-vue@3.0.0",
+    "rollup-plugin-vue3": "npm:rollup-plugin-vue@6.0.0",
```
Haroenv, line 105: smooth!

package.json
```diff
@@ -47,8 +49,8 @@
   "dependencies": {
     "algoliasearch-helper": "^3.1.0",
     "instantsearch.js": "^4.20.0",
-    "vue-demi": "npm:@eunjae-lee/vue-demi@0.9.3",
-    "mitt": "^2.1.0"
+    "mitt": "^2.1.0",
+    "vue-demi": "npm:@eunjae-lee/vue-demi@0.10.2"
```
Haroenv, line 53: is vue-demi a runtime dependency?

eunjae-lee: Yes. I updated the description of this PR to have more info on vue-demi.

src/components/InstantSearchSsr.js
eunjae-lee: No, in case of Vue 2, that `h` is undefined and we need to use the one passed to `render` function.

rollup.config.js
eunjae-lee: Yeah, I wanted to discuss with you about the output dirs.

With

```js
  "main": "dist/vue2/cjs",
  "module": "dist/vue2/es",
```

that config, for the vue2 users, it will be the same. So I think it's okay to have it under `dist/vue`.

Where should we have vue3 bundles?

For now, I'm importing like this:

```js
import ... from 'vue-instantsearch/dist/vue3/es';
```

package.json
eunjae-lee: added! https://github.com/algolia/vue-instantsearch/pull/952/commits/7be3418e89e5cea0a9beabecb73588cbb0988034

Haroenv:
can you remove the parts of vue-demi that we don't use in Vue InstantSearch, or do you think that isn't logical?

eunjae-lee:
> can you remove the parts of vue-demi that we don't use in Vue InstantSearch, or do you think that isn't logical?

Yeah I already did it in my [fork](https://github.com/eunjae-lee/vue-demi), including the removal of @vue/composition-api compatibility.

Do you see something specific that I missed?

Haroenv:
https://github.com/eunjae-lee/vue-demi/blob/master/lib/v2/index.esm.js#L17 is still there no? 

deprecated usage also should go away: https://github.com/eunjae-lee/vue-demi/blob/master/lib/v2/index.d.ts#L6-L10

eunjae-lee:
> [https://github.com/eunjae-lee/vue-demi/blob/master/lib/v2/index.esm.js#L17](https://github.com/eunjae-lee/vue-demi/blob/master/lib/v2/index.esm.js?rgh-link-date=2021-06-08T15%3A10%3A31Z#L17) is still there no?
> 
> deprecated usage also should go away: [https://github.com/eunjae-lee/vue-demi/blob/master/lib/v2/index.d.ts#L6-L10](https://github.com/eunjae-lee/vue-demi/blob/master/lib/v2/index.d.ts?rgh-link-date=2021-06-08T15%3A10%3A31Z#L6-L10)

You're right. The second one should go away. However the first one should stay.

In Vue 2 mode, 

```js
/**VCA-EXPORTS**/
export * from '@vue/composition-api'
/**VCA-EXPORTS**/
```

becomes

```js
/**VCA-EXPORTS**/
export const computed = undefined;
export const createApp = undefined;
export const createRef = undefined;
export const customRef = undefined;
export const defineAsyncComponent = undefined;
export const defineComponent = undefined;
export const del = undefined;
export const getCurrentInstance = undefined;
export const h = undefined;
export const inject = undefined;
export const isRaw = undefined;
export const isReactive = undefined;
export const isReadonly = undefined;
export const isRef = undefined;
export const markRaw = undefined;
export const nextTick = undefined;
export const onActivated = undefined;
export const onBeforeMount = undefined;
export const onBeforeUnmount = undefined;
export const onBeforeUpdate = undefined;
export const onDeactivated = undefined;
export const onErrorCaptured = undefined;
export const onMounted = undefined;
export const onServerPrefetch = undefined;
export const onUnmounted = undefined;
export const onUpdated = undefined;
export const provide = undefined;
export const proxyRefs = undefined;
export const reactive = undefined;
export const readonly = undefined;
export const ref = undefined;
export const set = undefined;
export const shallowReactive = undefined;
export const shallowReadonly = undefined;
export const shallowRef = undefined;
export const toRaw = undefined;
export const toRef = undefined;
export const toRefs = undefined;
export const triggerRef = undefined;
export const unref = undefined;
export const useCSSModule = undefined;
export const useCssModule = undefined;
export const warn = undefined;
export const watch = undefined;
export const watchEffect = undefined;
/**VCA-EXPORTS**/
```

They're necessary because when we do

```js
import { isVue3, h } from 'vue-demi';


...

render(createElement) {
  return (isVue3 ? h : createElement)(...)
}
```

In vue 2, we still need `h` to be exported even if it's undefined.

Haroenv:
I wonder if for our use case, we can manually do `export const h = undefined`

package.json
```diff
@@ -115,12 +119,20 @@
   },
   "bundlesize": [
     {
-      "path": "./dist/vue-instantsearch.js",
+      "path": "./dist/vue2/umd/index.js",
```
Haroenv, line 122: changing existing file names could be a breaking change for when someone is forcing UMD at the moment, could we keep the existing names the same and for example only add a new folder for vue 3?

Haroenv: Actually no, let' have a clean dist and release this as a major version

Haroenv (approved):
this looks great!

package.json
eunjae-lee: Oh I didn't realize we don't have jsdelivr or unpkg in the package.json at Vue InstantSearch. Yeah, the url of UMD will change. Let's go as a breaking change.

eunjae-lee:
> I wonder if for our use case, we can manually do `export const h = undefined`

Definitely. I'll update vue-demi and apply it in another PR.

PR #953, opened 2021-04-14
feat(insights): expose `send-event` in widgets

eunjae-lee:
## Summary

This PR exposes `send-event` in the following widgets:

- HierarchicalMenu
- Hits
- InfiniteHits
- Menu
- MenuSelect
- NumericMenu
- RangeInput
- RatingMenu
- RefinementList
- ToggleRefinement

By supporting the `middlewares` API in the recent change, [the `insights` middleware works](https://codesandbox.io/s/throbbing-darkness-g7zpg?file=/src/App.vue) with Vue InstantSearch but with this PR, we can let users write custom components and still utilize `sendEvent` method.

### TODO

- [x] Add tests

### DIFF
--- a/src/components/RefinementList.vue
+++ b/src/components/RefinementList.vue
@@ -17,6 +17,7 @@
       :createURL="state.createURL"
       :is-from-search="state.isFromSearch"
       :can-refine="state.canRefine"
+      :send-event="state.sendEvent"
     >
       <div
         :class="suit('searchBox')"
--- a/src/components/__tests__/HierarchicalMenu.js
+++ b/src/components/__tests__/HierarchicalMenu.js
@@ -501,7 +501,7 @@ it('exposes send-event method for insights middleware', () => {
     scopedSlots: {
       default: `
       <div slot-scope="{ sendEvent }">
-        <button @click="sendEvent()">Click</button>
+        <button @click="sendEvent()">Send Event</button>
       </div>
       `,
     },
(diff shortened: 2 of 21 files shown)


src/components/RefinementList.vue
```diff
@@ -17,6 +17,7 @@
       :createURL="state.createURL"
       :is-from-search="state.isFromSearch"
       :can-refine="state.canRefine"
+      :send-event="state.sendEvent"
```
Haroenv, line 20: do we still need to attach the default sendEvent, or is that done in the InstantSearch connector?

eunjae-lee: @Haroenv You mean calling the function when refining? If that's what you mean, it's done in the InstantSearch.js connector's refine and we're just exposing them as to the slot so that users can run it for custom purpose.

src/components/__tests__/HierarchicalMenu.js
```diff
@@ -488,6 +489,28 @@ it('calls the Panel mixin with `items.length`', () => {
...
+        <button @click="sendEvent()">Click</button>
+      </div>
+      `,
+    },
+  });
+
+  wrapper.find('button').trigger('click');
+  expect(sendEvent).toHaveBeenCalledTimes(1);
+});
+
```
eunjae-lee, line 513: @Haroenv do you see any better way to test if `sendEvent` exists?
By the way, I think we can only check if it exists. The logic of sending events is already tested in InstantSearch.js.

Haroenv: this is a good e2e style test, anything else will depend on internals too much IMO

Haroenv (approved):
nothing to remark!

PR #998, opened 2021-06-09
feat(vue 3): rename beforeDestroy to beforeUnmount

eunjae-lee:
## Summary

This PR renames `beforeDestroy` to `beforeUnmount` in case of Vue3 because it's been renamed.

https://v3.vuejs.org/guide/migration/introduction.html#other-minor-changes


Haroenv (approved):
I can't see another way doing this :)

PR #1000, opened 2021-06-10
fix(vue 3): fix unit tests for panel

eunjae-lee:
## Summary

This PR fixes the failing unit tests for the panel widget.
It was testing with the internal (the event emitter), and got broken after we decided to use `mitt`. We should change the tests not to test with the internal, but for now, I'm just making it work.

(I'll deal with e2e tests later, too.)

### DIFF
--- a/src/components/SearchInput.vue
+++ b/src/components/SearchInput.vue
@@ -155,11 +155,13 @@ export default {
     value: {
       type: String,
       required: false,
+      default: undefined,
     },
     modelValue: {
       type: String,
       required: false,
-    }
+      default: undefined,
+    },
   },
   emits: ['input', 'update:modelValue', 'blur', 'focus', 'reset'],
   data() {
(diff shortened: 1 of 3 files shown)


src/components/SearchInput.vue
```diff
@@ -155,11 +155,13 @@ export default {
     value: {
       type: String,
       required: false,
+      default: undefined,
```
Haroenv, line 158: what is this needed for?

eunjae-lee: eslint gave me this

```
/home/circleci/vue-instantsearch/src/components/SearchInput.vue
  155:5  error  Prop 'value' requires default value to be set       vue/require-default-prop
  159:5  error  Prop 'modelValue' requires default value to be set  vue/require-default-prop
```

I'm not sure why it didn't alert earlier.

Haroenv (approved):
looks good

PR #1002, opened 2021-06-11
fix(vue 3): fix Panel

eunjae-lee:
## Summary

This PR fixes an issue to make Panel widget work in Vue 3.

(Even though it's in Vue 3, it tried to access $scopedSlots which doesn't exist any more.)

### DIFF
--- a/src/components/Panel.vue
+++ b/src/components/Panel.vue
@@ -1,7 +1,7 @@
 <template>
   <div :class="[suit(), !canRefine && suit('', 'noRefinement')]">
     <div
-      v-if="$slots.header || $scopedSlots.header"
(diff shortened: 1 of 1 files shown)


Haroenv:
how come the tests fail again @eunjae-lee ?

Haroenv (changes requested):
this should probably get tests added

eunjae-lee:
> how come the tests fail again @eunjae-lee ?

This branch was outdated. Now it's up-to-date.

> this should probably get tests added

I think once we run unit tests for both vue 2 and vue 3, this will be covered.

Haroenv (approved):
good point, the code is already correct

PR #1003, opened 2021-06-14
chore(vue 3): enable unit tests for Vue 3

eunjae-lee:
## Summary

This PR makes CircleCI run linting and unit tests for Vue 3.

This PR is intended to enable CircleCI job and the environmental tasks. The failing unit tests will be taken care of in separate PRs.

* ref: https://github.com/lmiller1990/vtu-next-demo

### TODOs

These are the most recurring examples of errors:

1. widget mixin (seems to be broken in many cases)
2. different attribute order (not sure how to fix it yet ↓)
```
    -     <option value
    -             class="ais-MenuSelect-option"
    +     <option class="ais-MenuSelect-option"
    +             value
```
3. different dom outputs (cause not identified)
```
    -       <span>
    -         Apple
    -       </span>
    +       Apple (50)
```
4. using `h` from Vue2
```
  ● createServerRootMixin › hydrate › sets helper & mainHelper

    TypeError: h is not a function

      658 |         render(h) {
      659 |           return h(InstantSearchSsr, {}, [
    > 660 |             h(Configure, {
```
5. using `Vue` constructor

### DIFF
--- a/test/utils/index.js
+++ b/test/utils/index.js
@@ -2,7 +2,7 @@ import { isVue3 } from 'vue-demi';
 
 export const mount = isVue3
   ? (component, options = {}) => {
-      const { propsData, mixins, ...restOptions } = options;
+      const { propsData, mixins, provide, ...restOptions } = options;
       // If we `import` this, it will try to import Vue3-only APIs like `defineComponent`,
       // and jest will fail. So we need to `require` it.
       const wrapper = require('@vue/test-utils2').mount(component, {
@@ -10,6 +10,7 @@ export const mount = isVue3
         props: propsData,
         global: {
           mixins,
+          provide,
         },
       });
       wrapper.destroy = () => wrapper.unmount();
--- a/test/utils/index.js
+++ b/test/utils/index.js
@@ -13,7 +13,7 @@ export const mount = isVue3
           provide,
         },
       });
-      wrapper.destroy = () => wrapper.unmount();
+      wrapper.destroy = wrapper.unmount;
       return wrapper;
     }
   : require('@vue/test-utils').mount;
--- a/test/utils/index.js
+++ b/test/utils/index.js
@@ -1,13 +1,14 @@
 import { isVue3 } from 'vue-demi';
-import { mount as mount1 } from '@vue/test-utils';
-import { mount as mount2 } from '@vue/test-utils2';
+import { mount as mountVue2 } from '@vue/test-utils';
 
 export const mount = isVue3
   ? (component, options = {}) => {
       const { propsData, ...restOptions } = options;
-      return mount2(component, {
+      // If we `import` this, it will try to import Vue3-only APIs like `defineComponent`,
+      // and jest will fail. So we need to `require` it.
+      return require('@vue/test-utils2').mount(component, {
         ...restOptions,
         props: propsData,
       });
     }
-  : mount1;
+  : mountVue2;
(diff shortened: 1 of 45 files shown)


Haroenv:
looks like the tests on vue 3 still get the vue test utils vue 2:

```
    > 2 | import { mount as mountVue2 } from '@vue/test-utils';
```

Haroenv:
something is wrong with the babel/jest setup, I think it didn't get applied? 

```
FAIL  src/components/__tests__/ConfigureRelatedItems.js
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/circleci/vue-instantsearch/jest.setup.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import { isVue2, Vue2 } from 'vue-demi';
                                                                                             ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
```

eunjae-lee:
> something is wrong with the babel/jest setup, I think it didn't get applied?

The generated `babel.config.js` on the server was different from what I had in my mac. Fixed in https://github.com/algolia/vue-instantsearch/pull/1003/commits/0e8ab9d8d4443f4e1ad26330c599332fa2c7d82a and now I see the same errors on CircleCI.

Haroenv (changes requested):
I think it makes more sense to make the CI pass or at least having a list of what's to do to pass CI after this, wdyt?

eunjae-lee:
> I think it makes more sense to make the CI pass or at least having a list of what's to do to pass CI after this, wdyt?

Yeah it make sense. I'll write a list of todo here by today.

Haroenv:
Do you prefer having these tests merged & the feat/vue3-compat branch left failing, or waiting until there's a stack with the passing tests?

eunjae-lee:
> Do you prefer having these tests merged & the feat/vue3-compat branch left failing, or waiting until there's a stack with the passing tests?

Yeah, I wasn't clear. I'd like to get this merged and later create PRs to fix the issues until `feat/vue3-compat` branch becomes green.

test/utils/index.js
```diff
@@ -0,0 +1,19 @@
...
+      // and jest will fail. So we need to `require` it.
+      const wrapper = require('@vue/test-utils2').mount(component, {
+        ...restOptions,
+        props: propsData,
+        global: {
+          mixins,
+          provide,
+        },
+      });
+      wrapper.destroy = () => wrapper.unmount();
```
Haroenv, line 16:
```suggestion
      wrapper.destroy = wrapper.unmount;
```

does this work?

Haroenv (approved):
test setup seems correct

test/utils/index.js
eunjae-lee: I just tried and it works!

PR #1008, opened 2021-07-02
test(vue 3): replace slot-scope with v-slot

eunjae-lee:
## Summary

This PR replaces `slot-scope` with `v-slot` so that all the test cases can pass for both Vue 2 and Vue 3.

`slot-scope` is removed since Vue 3 and `v-slot` is supported since Vue 2.6.0.

---

Sorry for this massive PR.

image

Hiding whitespace changes make it easier to review this PR.

### Two major types of changes in this PR:

1.
```
slot-scope="..."
```

to

```
v-slot="..."
```

2. Mounting the test subject as a dependent component and use template to render them (and plus, using `data` to pass the props to the inner component.)

```js
    const wrapper = mount(Breadcrumb, {
      propsData: defaultProps,
      scopedSlots: {
        default: defaultScopedSlot,
      }
    });
```

to

```js
    const wrapper = mount({
      components: { Breadcrumb },
      data() {
        return { props: defaultProps };
      },
      template: `
        <Breadcrumb v-bind="props">
          ${defaultSlot}
        </Breadcrumb>
      `,
    });
```

### DIFF
--- a/src/components/__tests__/__snapshots__/StateResults.js.snap
+++ b/src/components/__tests__/__snapshots__/StateResults.js.snap
@@ -7,11 +7,7 @@ exports[`renders explanation if no slot is used 1`] = `
     Use this component to have a different layout based on a certain state.
   </p>
   <p>
-    Fill in the slot, and get access to the following things on the
-    <code>
-      v-slot
-    </code>
-    :
+    Fill in the slot, and get access to the following things:
   </p>
   <pre>
     results: [
--- a/src/components/__tests__/__snapshots__/StateResults.js.snap
+++ b/src/components/__tests__/__snapshots__/StateResults.js.snap
@@ -9,7 +9,7 @@ exports[`renders explanation if no slot is used 1`] = `
   <p>
     Fill in the slot, and get access to the following things on the
     <code>
-      slot-scope
+      v-slot
     </code>
     :
   </p>
--- a/src/components/StateResults.vue
+++ b/src/components/StateResults.vue
@@ -9,7 +9,7 @@
       </p>
       <p>
         Fill in the slot, and get access to the following things on the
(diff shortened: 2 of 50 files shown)


src/components/__tests__/__snapshots__/StateResults.js.snap
```diff
@@ -9,7 +9,7 @@ exports[`renders explanation if no slot is used 1`] = `
   <p>
     Fill in the slot, and get access to the following things on the
     <code>
-      slot-scope
+      v-slot
     </code>
```
Haroenv, line 13: maybe this should be "Fill in the slot and get access to the following things:"

that way it's not prescribing how to read it, and still straightforward for vue 2 usage

Haroenv (commented):
these tests are so much better! I always disliked how far away from the real use case they were

This is actually a refactor or a test though, not a fix, since the code itself didn't change :)

src/components/__tests__/__snapshots__/StateResults.js.snap
eunjae-lee: I just checked and we do not explicitly say anywhere "requires vue >= 2.6.0", but since we have it in our peerDependencies, can we assume it's being used on vue >= 2.6.0? If that's the case, `v-slot` might be enough.

Haroenv: I just don't think "v-slot" or "slot-scope" is the best wording

eunjae-lee: it makes sense https://github.com/algolia/vue-instantsearch/pull/1008/commits/d5d33642fdf1d868925488a21b0a8135cf610499

Haroenv (approved):

PR #1009, opened 2021-07-07
chore: remove vue-demi

eunjae-lee:
## Summary

This PR removes `vue-demi` and instead include a file `src/util/vue-utils.js` which is basically does the same thing as vue-demi.

In the build script `scripts/build-vue3.sh`, it temporarily copies the vue 3 version of `vue-utils.js` into `src/util`.

Now that we have the control over the conditional codes at the build time, dead code is eliminated by rollup, which was an unexpectedly good side-effect.

### Why

`vue-demi` relies on `postinstall` and it can be flaky and we've already seen a case that it failed to work. Since we have two different build outputs, we can just remove vue-demi and statically include what's supposed to be in the version.


scripts/vue-utils-v2.js
```diff
@@ -0,0 +1,55 @@
+import Vue from 'vue';
```
Haroenv, line 1: why aren't these files in src, but in `scripts`? 

I suggest:

```
/util/vue/index-3.js
/util/vue/index-2.js
/util/vue/index.js
```

and with index.js having

```
export * from './index-2.js'
```

and in the script that file is replaced with

```
export * from './index-3.js'
```

eunjae-lee: yeah that's cleaner. thanks! https://github.com/algolia/vue-instantsearch/pull/1009/commits/c3eef175996ab58ab52b3de554f0d803bf10092a

.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           command: |
             yarn remove vue vue-jest vue-template-compiler babel-preset-es2015
             yarn add vue@3.0.6 vue-jest@5.0.0-alpha.10 vue-loader@16.1.2 jest@26.6.3 babel-jest@25.2.6 @babel/preset-env@7.14.5 -D -E
-            cp ./scripts/vue-utils-v3.js src/util/vue-utils.js
+            echo "export * from './index-3';" > src/util/vue/index.js
```
Haroenv, line 69: why doesn't this use yarn build:vue-3 or the build-vue3.sh script?

eunjae-lee: When we run unit tests, we actually don't need `dist`.

Haroenv: I see, would still be interesting to be able to test vue 3 locally without swapping dependencies though

scripts/build-vue3.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/sh
 set -e
 
-vue-demi-switch 3
+# use v3 for src/util/vue
```
Haroenv, line 4: read existing content, and save that in a variable, then set it back (avoids a bug or accidentally having a diff if the original changes?)

scripts/build-vue2.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 set -e
 
-vue-demi-switch 2
 VUE_VERSION=vue2 rollup -c
+
```
Haroenv, line 5: also read the content and write just the export

src/util/vue/index.js
```diff
@@ -0,0 +1 @@
+export * from './index-2';
```
Haroenv, line 1:
```suggestion
// ... explain how this file gets replaced by the scripts
export * from './index-2';
```

Haroenv (commented):
👌

.circleci/config.yml
eunjae-lee: Yeah the local workflow to fix a failing unit test in vue 3 is quite a mess. So far I have no idea.

Haroenv (approved):

PR #1013, opened 2021-07-09
test: sort attributes asc in snapshot tests

eunjae-lee:
## Summary

This PR updates the jest configuration to sort attributes to ascending in snapshot tests.

### Why?

I don't know exactly in what order attributes get sorted in Vue 2 (probably as written), but in Vue 3, it's definitely ascending.

And it makes it impossible for us to maintain a single set of snapshots for both Vue 2 and 3.

### How it's done

We've been using `jest-serializer-html`, which uses `diffable-html` under the hood. `diffable-html` formats html but doesn't accept any option. So here we use `patch-package` to add an option there which is `sortAttributes: Function`, a sort comparator.

And `jest-serializer-html` is a simple wrapper around `diffable-html`, so we removed it and added a single file at `test/jest-serializer-html.js`. It calls the formatter from `diffable-html` with the sort function.


patches/diffable-html+4.0.0.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/node_modules/diffable-html/dist/diffable-html.js b/node_modules/diffable-html/dist/diffable-html.js
+index 0d7d767..98fbad8 100644
+--- a/node_modules/diffable-html/dist/diffable-html.js
++++ b/node_modules/diffable-html/dist/diffable-html.js
+@@ -21,7 +21,7 @@ var voidElements = [
+   'track',
+   'wbr' ];
+ 
+-var format = function(html) {
++var format = function(html, { sortAttributes = undefined } = {}) {
```
Haroenv, line 10: could you not open a PR to diffable html for this?

Haroenv (commented):
the actual snapshot files look totally ok, I've reviewed, although I'd rather rely on a deployed version of diffable-html. Is there any reason why you didn't make a PR?

patches/diffable-html+4.0.0.patch
tkrugg: It looks like he did but then went for the patch solution. https://github.com/rayrutjes/diffable-html/pull/39

tkrugg (approved):

Haroenv (changes requested):
we should rely on either a published fork, or reopen the PR on diffable-html and jest-serializer-html, while they are stable packages, if this patch doesn't get applied somehow (postinstall doesn't get executed on yarn 2 by default for example) the tests don't pass

eunjae-lee:
> we should rely on either a published fork, or reopen the PR on diffable-html and jest-serializer-html, while they are stable packages, if this patch doesn't get applied somehow (postinstall doesn't get executed on yarn 2 by default for example) the tests don't pass

btw, we're getting the ownership of the repos soon.

package.json
```diff
@@ -125,16 +127,11 @@
     "brotli-size": "4.0.0"
   },
   "jest": {
-    "setupFiles": [
-      "<rootDir>/jest.setup.js"
-    ],
+    "setupTestFrameworkScriptFile": "./jest.setup.js",
```
eunjae-lee, line 128: `expect` doesn't exist at the `setupFiles` phase.

Haroenv (approved):
let's go!

PR #1014, opened 2021-07-09
test: update tests compatible for both vue 2 and vue 3 [1/2]

eunjae-lee:
## Summary

This PR mainly updates all the incompatible test cases to work with both vue 2 and 3.

---

It's normal that the CI doesn't pass here. It's fixed in https://github.com/algolia/vue-instantsearch/pull/1021


src/util/createServerRootMixin.js
```diff
@@ -14,13 +15,33 @@ function walkIndex(indexWidget, visit) {
...
+  if (!_renderToString) {
+    if (isVue3) {
+      throw new Error('you need to install @vue/server-renderer');
+    } else {
+      throw new Error('you need to install vue-server-renderer');
+    }
+  }
+
+  if (isVue3) {
+    return _renderToString(app);
```
eunjae-lee, line 36: It returns a promise in vue 3.

src/util/createServerRootMixin.js
```diff
@@ -40,31 +61,56 @@ function searchOnlyWithDerivedHelpers(helper) {
...
+  if (isVue3) {
+    const appOptions = Object.assign({}, componentInstance.$options, options);
+    appOptions.mixins = [...appOptions.mixins, ...mixins];
+    // Unlike Vue 2, there is no componentInstance.$options.propsData in Vue 3.
+    // The only way to pass the propsData is to spread componentInstance
+    // in the second argument, hoping the rest wouldn't make any side effect.
+    // At this point, we don't even have the definition of the props.
+    // So we cannot pass exactly the propsData only.
+    // FIXME: Maybe we need to get the list of props in `createServerRootMixin`.
+    app = createSSRApp(appOptions, _objectSpread({}, componentInstance));
```
eunjae-lee, line 88: This may change the signature of `createServerRootMixin` for vue 3, but needs more tests.

src/util/createServerRootMixin.js
```diff
@@ -40,31 +61,56 @@ function searchOnlyWithDerivedHelpers(helper) {
...
+  }
 
   // https://stackoverflow.com/a/48195006/3185307
   app.$slots = componentInstance.$slots;
   app.$root = componentInstance.$root;
-  app.$options.serverPrefetch = [];
+  if (isVue2) {
+    app.$options.serverPrefetch = [];
+  } else {
+    // FIXME: ???
```
eunjae-lee, line 112: IIRC, this is to prevent infinite calling of `serverPrefetch`, but this instance variable doesn't exist in vue 3 and without it, it seems okay so far.

src/util/createServerRootMixin.js
```diff
@@ -88,36 +134,35 @@ function augmentInstantSearch(
...
-        app = cloneComponent(componentInstance);
-
-        app.instantsearch.helper = helper;
-        app.instantsearch.mainHelper = helper;
-
-        app.instantsearch.mainIndex.init({
-          instantSearchInstance: app.instantsearch,
-          parent: null,
-          uiState: app.instantsearch._initialUiState,
+        app = cloneComponent(componentInstance, {
```
eunjae-lee, line 142: Unlike vue 2, in vue 3, we cannot access its instance variables like this.
So in order to manipulate its data, I'm passing a mixin to `cloneComponent`, so that the manipulation can happen in `created` lifecycle.

eunjae-lee (commented):
`src/util/__tests__/createServerRootMixin.test.js` is collapsed by default. When you review the file, hiding whitespace changes might ease the diff.

.circleci/config.yml
```diff
@@ -65,13 +65,16 @@ jobs:
           name: Install Vue3
           command: |
             yarn remove vue vue-jest vue-template-compiler babel-preset-es2015
-            yarn add vue@3.0.6 vue-jest@5.0.0-alpha.10 vue-loader@16.1.2 jest@26.6.3 babel-jest@25.2.6 @babel/preset-env@7.14.5 -D -E
-            echo "export * from './index-3';" > src/util/vue/index.js
+            yarn add vue@3.1.2 @vue/server-renderer@3.1.2 vue-jest@5.0.0-alpha.10 vue-loader@16.1.2 jest@26.6.3 babel-jest@25.2.6 @babel/preset-env@7.14.5 @babel/plugin-transform-runtime@7.14.5 -D -E
```
Haroenv, line 68: this is starting to be a lot of code, shouldn't it be in a dedicated script instead of inline in yaml?

src/components/InstantSearch.js
```diff
@@ -97,7 +97,9 @@ export default createInstantSearchComponent({
           [this.suit('', 'ssr')]: false,
         },
       },
-      isVue3 ? this.$slots.default() : this.$slots.default
+      isVue3
+        ? this.$slots.default && this.$slots.default()
+        : this.$slots.default
```
Haroenv, line 102: maybe we should have a shared util for this?

src/components/__tests__/DynamicWidgets.js
```diff
@@ -58,10 +53,11 @@ it('renders all children without state', () => {
       </ais-panel>
       `,
     },
-    stubs: {
+    components: {
```
Haroenv, line 56: so much cleaner, thanks!

src/components/__tests__/Index.js
```diff
@@ -55,8 +54,6 @@ it('renders its children', () => {
 });
 
 it('provides the index widget', done => {
-  Vue.config.errorHandler = done;
```
Haroenv, line 58: this is needed, otherwise if there's an error, the test still passes

src/components/__tests__/InstantSearch.js
```diff
@@ -27,6 +28,9 @@ it('passes props to InstantSearch.js', () => {
       stalledSearchDelay: 250,
       searchFunction,
     },
+    slots: {
+      default: '',
+    },
```
Haroenv, line 33: not needed anymore after the fix with &&, no?

src/components/__tests__/InstantSearch.js
```diff
@@ -52,6 +57,9 @@ it('throws on usage of appId or apiKey', () => {
       appId: 'blabla',
       indexName: 'something',
     },
+    slots: {
+      default: '',
```
Haroenv, line 61: same question

src/components/__tests__/InstantSearch.js
```diff
@@ -327,16 +382,21 @@ it('disposes the instantsearch instance on unmount', async () => {
...
 
   const ChildComponent = {
     inject: ['$_ais_instantSearchInstance'],
     mounted() {
       this.$nextTick(() => {
         expect(typeof this.$_ais_instantSearchInstance).toBe('object');
-        expect(wrapper.vm.instantSearchInstance).toBe(
-          this.$_ais_instantSearchInstance
-        );
+        expect(this.$_ais_instantSearchInstance).toBe(instantSearchInstance);
```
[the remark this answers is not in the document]
Haroenv: you also fixed the order of the actual / expected

Haroenv (approved):
this is great! approve with the context that the default slots in the InstantSearch test should be removed

src/components/__tests__/InstantSearch.js
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/1014/commits/5df22838e95fe5890e420a6bc98edd0d3b9fb4db

src/components/InstantSearch.js
eunjae-lee: Makes sense. I'll see what I can do in a separate PR.

.circleci/config.yml
eunjae-lee: I'll do it in a separate PR (because there are other stacked PRs).

eunjae-lee:
please don't delete the branch yet.

PR #1016, opened 2021-07-12
feat(props): remove defaults

Haroenv:
for facetOrdering, we need to let InstantSearch.js decide the real default value of props. At the same time I removed all other defaults to avoid differences

### DIFF
--- a/src/components/QueryRuleCustomData.vue
+++ b/src/components/QueryRuleCustomData.vue
@@ -36,7 +36,7 @@ export default {
     transformItems: {
       type: Function,
       required: false,
-      default: items => items,
+      default: undefined,
     },
   },
   computed: {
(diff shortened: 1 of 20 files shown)


src/components/QueryRuleCustomData.vue
```diff
@@ -36,7 +36,7 @@ export default {
     transformItems: {
       type: Function,
       required: false,
```
shortcuts, line 38: do we need to state it if it's `false`?

shortcuts (approved):
Changes looks correct!

src/components/QueryRuleCustomData.vue
Haroenv: It doesn't really make much difference, so required defaults to true

PR #1020, opened 2021-07-16
fix(vue 3): update the implementation of createServerRootMixin

eunjae-lee:
## Summary

This PR updates the implementation of `createServerRootMixin`.
It excludes the test cases for vue 3 about forwarding $root and $slots. I haven't figured it out, and if it's really necessary, etc.

I will try it out separately and if so, I'll probably do it in a separate PR as a fix (because I don't think I'll do it immediately).

---

It's normal that the CI doesn't pass here. It's fixed in https://github.com/algolia/vue-instantsearch/pull/1021 because

```
feat/vue3-compat
  ㄴ test: update tests compatible for both vue 2 and vue 3 [1/2] #1014
    ㄴ this PR
       ㄴ  test: update tests compatible for both vue 2 and vue 3 [2/2] #1021
```

However once it's approved, I will rebase this and merge it as an individual commit to `feat/vue3-compat` (not mixing with the "update tests" PRs).

### DIFF
--- a/src/util/__tests__/createServerRootMixin.test.js
+++ b/src/util/__tests__/createServerRootMixin.test.js
@@ -1,15 +1,18 @@
-import Vue from 'vue';
-import { mount } from '../../../test/utils';
-import _renderToString from 'vue-server-renderer/basic';
+import { mount, createSSRApp, renderCompat } from '../../../test/utils';
 import Router from 'vue-router';
 import Vuex from 'vuex';
-import { createServerRootMixin } from '../createServerRootMixin';
+import { createStore } from 'vuex4';
+import {
+  createServerRootMixin,
+  renderToString,
+} from '../createServerRootMixin';
 import InstantSearchSsr from '../../components/InstantSearchSsr';
 import Configure from '../../components/Configure';
 import SearchBox from '../../components/SearchBox.vue';
 import { createWidgetMixin } from '../../mixins/widget';
 import { createFakeClient } from '../testutils/client';
 import { createSerializedState } from '../testutils/helper';
+import { isVue3, isVue2, Vue2 } from '../vue-compat';
 import {
   SearchResults,
   SearchParameters,
@@ -17,15 +20,6 @@ import {
 } from 'algoliasearch-helper';
 
 jest.unmock('instantsearch.js/es');
-
-function renderToString(app) {
-  return new Promise((resolve, reject) =>
-    _renderToString(app, {}, (err, res) => {
-      if (err) reject(err);
-      resolve(res);
-    })
-  );
-}
 
 const forceIsServerMixin = {
   beforeCreate() {
@@ -44,55 +38,55 @@ process.env.VUE_ENV = 'server';
 describe('createServerRootMixin', () => {
   describe('creation', () => {
     it('requires searchClient', () => {
-      expect(
-        () =>
-          new Vue({
-            mixins: [
-              createServerRootMixin({
-                searchClient: undefined,
-                indexName: 'lol',
-              }),
-            ],
-          })
+      expect(() =>
+        createSSRApp({
+          mixins: [
+            createServerRootMixin({
+              searchClient: undefined,
+              indexName: 'lol',
+            }),
+          ],
+        })
       ).toThrowErrorMatchingInlineSnapshot(
         `"createServerRootMixin requires \`searchClient\` and \`indexName\` in the first argument"`
       );
     });
 
     it('requires indexName', () => {
-      expect(
-        () =>
-          new Vue({
-            mixins: [
-              createServerRootMixin({
-                searchClient: createFakeClient(),
-                indexName: undefined,
-              }),
-            ],
-          })
+      expect(() =>
+        createSSRApp({
+          mixins: [
+            createServerRootMixin({
+              searchClient: createFakeClient(),
+              indexName: undefined,
+            }),
+          ],
+        })
       ).toThrowErrorMatchingInlineSnapshot(
         `"createServerRootMixin requires \`searchClient\` and \`indexName\` in the first argument"`
       );
     });
 
     it('creates an instantsearch instance on "data"', () => {
-      const app = new Vue({
+      const App = {
         mixins: [
           createServerRootMixin({
             searchClient: createFakeClient(),
             indexName: 'lol',
           }),
         ],
-      });
+        render: () => null,
+      };
 
-      expect(app.$data).toEqual({
+      const wrapper = mount(App);
+      expect(wrapper.vm.$data).toEqual({
         instantsearch: expect.objectContaining({
           start: expect.any(Function),
         }),
       });
     });
 
-    it('provides the instantsearch instance ', () => {
+    it('provides the instantsearch instance', done => {
       const App = {
         mixins: [
           createServerRootMixin({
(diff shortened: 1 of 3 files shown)


src/util/createServerRootMixin.js
```diff
@@ -40,31 +61,56 @@ function searchOnlyWithDerivedHelpers(helper) {
...
     fetch: undefined,
     _base: undefined,
     name: 'ais-ssr-root-component',
-    // copy over global Vue APIs
-    router: componentInstance.$router,
-    store: componentInstance.$store,
   };
+  if (isVue2) {
+    // copy over global Vue APIs
+    options.router = componentInstance.$router;
```
Haroenv, line 73: this can be put in the else of isVue3

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -18,15 +21,6 @@ import {
 
 jest.unmock('instantsearch.js/es');
 
-function renderToString(app) {
```
eunjae-lee, line 21: Now it's moved to `createServerRootMixin.js`.

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -336,187 +374,191 @@ Array [
...
 
-      const wrapper = new Vue({
+      const wrapper = createSSRApp({
         mixins: [forceIsServerMixin],
-        render(h) {
-          return h(App, { props: { someProp } });
-        },
+        render: renderCompat(h =>
+          h(App, isVue3 ? { someProp } : { props: { someProp } })
+        ),
```
eunjae-lee, line 401: Little change in the syntax. Because of things like this, in general, it's easier (and more readable) to write in template form rather than in render function.

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -336,187 +374,191 @@ Array [
...
       });
 
       await renderToString(wrapper);
     });
 
-    it('forwards slots', async done => {
-      const searchClient = createFakeClient();
+    // FIXME: make these work with Vue 3
+    if (isVue2) {
```
eunjae-lee, line 408: forwarding $slots and $root doesn't work in Vue 3 yet.

However the app (that is created by `createSSRApp()`) doesn't have `$slots` or `$root` as instance variables in the first place. To better understand this issue, I need to find out the real life examples in vue 2 where forwarding `$slots` or `$root` matter, and replicate it with vue 3 to know if it still matters or not.

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -576,33 +623,37 @@ Array [
...
           this.instantsearch.hydrate({
             movies: nonSerialized,
           });
-        },
-      };
 
-      const {
-        vm: { instantsearch },
-      } = mount(app);
+          expect(this.instantsearch.__initialSearchResults).toEqual(
```
eunjae-lee, line 646: Instead of accessing `app.vm.xxx`, we're asserting inside the lifecycle method.

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -576,33 +623,37 @@ Array [
...
+            expect.objectContaining({ movies: expect.any(SearchResults) })
+          );
 
-      expect(instantsearch.__initialSearchResults).toEqual(
-        expect.objectContaining({ movies: expect.any(SearchResults) })
-      );
+          expect(this.instantsearch.__initialSearchResults.movies).toEqual(
```
eunjae-lee, line 650: I'm not sure why the object reference is different, but it might have something to do with mocking or something around it. Not sure, though.

src/util/createServerRootMixin.js
```diff
@@ -14,13 +15,33 @@ function walkIndex(indexWidget, visit) {
...
+  if (!_renderToString) {
+    if (isVue3) {
+      throw new Error('you need to install @vue/server-renderer');
+    } else {
+      throw new Error('you need to install vue-server-renderer');
+    }
+  }
+
+  if (isVue3) {
+    return _renderToString(app);
```
eunjae-lee, line 36: It returns a promise in vue 3.

src/util/createServerRootMixin.js
```diff
@@ -40,31 +61,56 @@ function searchOnlyWithDerivedHelpers(helper) {
...
+  if (isVue3) {
+    const appOptions = Object.assign({}, componentInstance.$options, options);
+    appOptions.mixins = [...appOptions.mixins, ...mixins];
+    // Unlike Vue 2, there is no componentInstance.$options.propsData in Vue 3.
+    // The only way to pass the propsData is to spread componentInstance
+    // in the second argument, hoping the rest wouldn't make any side effect.
+    // At this point, we don't even have the definition of the props.
+    // So we cannot pass exactly the propsData only.
+    // FIXME: Maybe we need to get the list of props in `createServerRootMixin`.
+    app = createSSRApp(appOptions, _objectSpread({}, componentInstance));
```
eunjae-lee, line 88: This may change the signature of `createServerRootMixin` for vue 3, but needs more real life examples to verify.

src/util/createServerRootMixin.js
```diff
@@ -40,31 +61,56 @@ function searchOnlyWithDerivedHelpers(helper) {
...
+  }
 
   // https://stackoverflow.com/a/48195006/3185307
   app.$slots = componentInstance.$slots;
   app.$root = componentInstance.$root;
-  app.$options.serverPrefetch = [];
+  if (isVue2) {
+    app.$options.serverPrefetch = [];
+  } else {
+    // FIXME: ???
```
eunjae-lee, line 112: IIRC, this is to prevent infinite calling of `serverPrefetch`, but this instance variable doesn't exist in vue 3 and without it, it seems okay so far.

src/util/createServerRootMixin.js
```diff
@@ -88,36 +134,35 @@ function augmentInstantSearch(
...
-        app = cloneComponent(componentInstance);
-
-        app.instantsearch.helper = helper;
-        app.instantsearch.mainHelper = helper;
-
-        app.instantsearch.mainIndex.init({
-          instantSearchInstance: app.instantsearch,
-          parent: null,
-          uiState: app.instantsearch._initialUiState,
+        app = cloneComponent(componentInstance, {
```
eunjae-lee, line 142: Unlike vue 2, in vue 3, we cannot access its instance variables like this.
So in order to manipulate its data, I'm passing a mixin to `cloneComponent`, so that the manipulation can happen in `created` lifecycle.

src/util/createServerRootMixin.js
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/1020/commits/0a2a7784561b8dda4ed4ca4e165e1d652c598775

src/util/__tests__/createServerRootMixin.test.js
Haroenv: I like renderCompat, could we use it everywhere (InstantSearch/configure component too)? maybe with these conditionals already part of the item?

eunjae-lee: Good idea. I will do it in a separate PR. Just to clarify,

> with these conditionals already part of the item?

What do you mean by "item"? You mean the createElement, h, and these conditional props altogether covered by renderCompat function?

src/util/__tests__/createServerRootMixin.test.js
Haroenv: $slots  matter for nuxt where the InstantSearch component is in a layout, $root for similar case with i18n, see https://github.com/algolia/vue-instantsearch/pull/898 for example

src/util/__tests__/createServerRootMixin.test.js
Haroenv: is it with the serialising?

src/util/createServerRootMixin.js
Haroenv: it's thanks to https://github.com/algolia/vue-instantsearch/pull/1020/files#diff-ac44dbb445b26102c11a145082de41832504ef96bb746a1a05387f759c3d8909L45 in vue 3

src/util/__tests__/createServerRootMixin.test.js
eunjae-lee: ah right. I got confused with other test cases where it used to be `toBe` but I had to replace with `toEqual`. In this case it makes sense to use `toEqual`.

src/util/createServerRootMixin.js
eunjae-lee: Ah you're right!

src/util/createServerRootMixin.js
```diff
@@ -14,13 +15,33 @@ function walkIndex(indexWidget, visit) {
...
-    _renderToString(app, (err, res) => {
-      if (err) reject(err);
-      resolve(res);
-    })
-  );
+export function renderToString(app) {
+  let _renderToString;
+  try {
+    _renderToString = isVue3
+      ? require('@vue/server-renderer').renderToString
```
Haroenv, line 22: this needs to be added to peer dependencies, but with peerDependenciesMeta as optional (https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta) (and also same for vue-server-renderer)

src/util/createServerRootMixin.js
Haroenv: what do you expect to be solved in the fixme before merging?

src/util/createServerRootMixin.js
Haroenv: this seems like a better option anyway

src/util/createServerRootMixin.js
```diff
@@ -88,36 +131,35 @@ function augmentInstantSearch(
...
-        app.instantsearch.mainIndex.init({
-          instantSearchInstance: app.instantsearch,
-          parent: null,
-          uiState: app.instantsearch._initialUiState,
+        app = cloneComponent(componentInstance, {
+          mixins: [
+            {
+              created() {
+                // eslint-disable-next-line consistent-this
+                renderedApp = this;
```
Haroenv, line 144: is that not the same as `app` already?

Haroenv (commented):
overall these changes look like the wouldn't break vue 2

src/util/createServerRootMixin.js
eunjae-lee: I'm not too sure. We could merge this as-is and I can make some examples to see how they act in real life, and later come back with a solution.

src/util/createServerRootMixin.js
eunjae-lee: I don't remember why I did it but I just confirmed they're the same.

src/util/createServerRootMixin.js
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/1020/commits/3b264d2b1f950fea76d141d8eac49ae7c3b2f42d

Haroenv (approved):

PR #1021, opened 2021-07-17
 test: update tests compatible for both vue 2 and vue 3 [2/2]

eunjae-lee:
## Summary

This PR mainly updates all the incompatible test cases to work with both vue 2 and 3.

### About the scope of this PR

* We have no idea on when VTU is going to release the update. So I prefer replacing `vm.$set()` with `setData()` later in a separate PR.
* Another tasks I'm planning as separate PRs:
  * `renderCompat` for InstantSearch, Configure, ... (covers createElement, h, and conditional props & attrs)
  * shared util for `isVue3 ? this.$slots.default && this.$slots.default() : this.$slots.default`
  * extracting as a script to add dependencies and run tests for vue 3

This PR is a stacked PR on https://github.com/algolia/vue-instantsearch/pull/1020, so it's easier to do them separately. Otherwise, we will apply them both in this branch and in the one for #1020.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -13,10 +13,27 @@ expect.addSnapshotSerializer(
   })
 );
 
+const toHaveBooleanAttribute = attribute => wrapper => {
+  // vue2: https://codesandbox.io/s/optimistic-blackwell-enw30
+  // vue3: https://codesandbox.io/s/affectionate-vaughan-f6sf0
+  const value = wrapper.attributes(attribute);
+  if ((isVue2 && value === attribute) || (isVue3 && value === '')) {
+    return { pass: true };
+  } else {
+    return {
+      pass: false,
+      message: () => `expected ${wrapper} to have \`${attribute}\` attribute`,
+    };
+  }
+};
+
 expect.extend({
-  toHaveEmptyHTML: received => {
-    const html = received.html();
-    if ((isVue2 && html === '') || (isVue3 && html === '<!---->')) {
+  toHaveEmptyHTML: wrapper => {
+    const html = wrapper.html();
+    if (
+      (isVue2 && html === '') ||
+      (isVue3 && ['<!---->', '<!--v-if-->'].includes(html))
+    ) {
       return {
         pass: true,
       };
@@ -27,4 +44,7 @@ expect.extend({
       };
     }
   },
+  toBeDisabled: toHaveBooleanAttribute('disabled'),
+  toBeHidden: toHaveBooleanAttribute('hidden'),
+  toBeAutofocused: toHaveBooleanAttribute('autofocus'),
 });
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -14,8 +14,11 @@ expect.addSnapshotSerializer(
 );
 
 const toHaveBooleanAttribute = attribute => wrapper => {
-  // vue2: https://codesandbox.io/s/optimistic-blackwell-enw30
-  // vue3: https://codesandbox.io/s/affectionate-vaughan-f6sf0
+  // :hidden="true" becomes
+  // hidden="hidden" in Vue 2 and
+  // hidden="" in Vue 3.
+
+  // So we need this to write correct tests to match them in both versions.
   const value = wrapper.attributes(attribute);
   if ((isVue2 && value === attribute) || (isVue3 && value === '')) {
     return { pass: true };
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -13,6 +13,23 @@ expect.addSnapshotSerializer(
   })
 );
 
+const toHaveEmptyHTML = wrapper => {
+  const html = wrapper.html();
+  if (
+    (isVue2 && html === '') ||
+    (isVue3 && ['<!---->', '<!--v-if-->'].includes(html))
+  ) {
+    return {
+      pass: true,
+    };
+  } else {
+    return {
+      pass: false,
+      message: () => `expected ${html} to be an empty HTML string`,
+    };
+  }
+};
+
 const toHaveBooleanAttribute = attribute => wrapper => {
   // :hidden="true" becomes
   // hidden="hidden" in Vue 2 and
@@ -31,22 +48,7 @@ const toHaveBooleanAttribute = attribute => wrapper => {
 };
 
 expect.extend({
-  toHaveEmptyHTML: wrapper => {
-    const html = wrapper.html();
-    if (
-      (isVue2 && html === '') ||
-      (isVue3 && ['<!---->', '<!--v-if-->'].includes(html))
-    ) {
-      return {
-        pass: true,
-      };
-    } else {
-      return {
-        pass: false,
-        message: () => `expected ${html} to be an empty HTML string`,
-      };
-    }
-  },
+  toHaveEmptyHTML,
   toBeDisabled: toHaveBooleanAttribute('disabled'),
   toBeHidden: toHaveBooleanAttribute('hidden'),
   toBeAutofocused: toHaveBooleanAttribute('autofocus'),
--- a/src/components/SearchBox.vue
+++ b/src/components/SearchBox.vue
@@ -21,15 +21,44 @@
         :class-names="classNames"
         v-model="currentRefinement"
       >
+        <template
+          v-slot:loading-indicator
+          v-if="isVue3"
+        >
+          <slot
+            name="loading-indicator"
+          />
+        </template>
         <slot
+          v-if="isVue2"
           name="loading-indicator"
           slot="loading-indicator"
         />
+
+        <template
+          v-slot:submit-icon
+          v-if="isVue3"
+        >
+          <slot
+            name="submit-icon"
+          />
+        </template>
         <slot
+          v-if="isVue2"
           name="submit-icon"
           slot="submit-icon"
         />
+
+        <template
+          v-slot:reset-icon
+          v-if="isVue3"
+        >
+          <slot
+            name="reset-icon"
+          />
+        </template>
         <slot
+          v-if="isVue2"
           name="reset-icon"
           slot="reset-icon"
         />
@@ -42,6 +71,7 @@
 import { connectSearchBox } from 'instantsearch.js/es/connectors';
 import { createSuitMixin } from '../mixins/suit';
 import { createWidgetMixin } from '../mixins/widget';
+import { isVue3, isVue2 } from '../util/vue-compat';
 import SearchInput from './SearchInput.vue';
 
 export default {
@@ -86,6 +116,8 @@ export default {
   data() {
     return {
       localValue: '',
+      isVue2,
+      isVue3,
     };
   },
   computed: {
--- a/src/components/__tests__/CurrentRefinements.js
+++ b/src/components/__tests__/CurrentRefinements.js
@@ -1,4 +1,4 @@
-import { mount } from '../../../test/utils';
+import { mount, nextTick } from '../../../test/utils';
 import CurrentRefinements from '../CurrentRefinements.vue';
 import { __setState } from '../../mixins/widget';
 
@@ -177,11 +177,8 @@ it('calls the Panel mixin with `canRefine`', async () => {
 
   expect(mapStateToCanRefine()).toBe(true);
 
-  await wrapper.setData({
-    state: {
-      items: [],
-    },
-  });
+  wrapper.vm.state.items = [];
+  await nextTick();
 
   expect(mapStateToCanRefine()).toBe(false);
 
--- a/src/components/__tests__/DynamicWidgets.js
+++ b/src/components/__tests__/DynamicWidgets.js
@@ -50,27 +50,29 @@ it('renders all children without state', () => {
     },
   });
 
-  expect(wrapper.classes()).toContain('ais-DynamicWidgets');
-  expect(wrapper).toBeHidden();
-  expect(wrapper.element.innerHTML).toMatchInlineSnapshot(`
-<div class="ais-DynamicWidgets-widget">
-  <div widget-name="ais-refinement-list">
-    test1
+  expect(wrapper.htmlCompat()).toMatchInlineSnapshot(`
+<div class="ais-DynamicWidgets"
+     hidden="hidden"
+>
+  <div class="ais-DynamicWidgets-widget">
+    <div widget-name="ais-refinement-list">
+      test1
+    </div>
   </div>
-</div>
-<div class="ais-DynamicWidgets-widget">
-  <div widget-name="ais-menu">
-    test2
+  <div class="ais-DynamicWidgets-widget">
+    <div widget-name="ais-menu">
+      test2
+    </div>
   </div>
-</div>
-<div class="ais-DynamicWidgets-widget">
-  <div class="ais-Panel">
-    <div class="ais-Panel-body">
-      <div widget-name="ais-hierarchical-menu">
-        [
-  "test3",
-  "test4"
-]
+  <div class="ais-DynamicWidgets-widget">
+    <div class="ais-Panel">
+      <div class="ais-Panel-body">
+        <div widget-name="ais-hierarchical-menu">
+          [
+          "test3",
+          "test4"
+          ]
+        </div>
       </div>
     </div>
   </div>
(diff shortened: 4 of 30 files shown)


src/components/__tests__/CurrentRefinements.js
```diff
@@ -177,11 +177,8 @@ it('calls the Panel mixin with `canRefine`', async () => {
 
   expect(mapStateToCanRefine()).toBe(true);
 
-  await wrapper.setData({
-    state: {
-      items: [],
-    },
-  });
+  wrapper.vm.state.items = [];
```
eunjae-lee, line 180: https://github.com/vuejs/vue-test-utils-next/issues/766

this will be reverted once we get the new VTU release.

src/components/__tests__/DynamicWidgets.js
```diff
@@ -1,89 +1,76 @@
...
-  render(h) {
-    return h('div', {
-      attrs: {
-        'widget-name': 'ais-refinement-list',
-        attribute: this.attribute,
-      },
-    });
-  },
+  template: `
+    <div widget-name="ais-refinement-list">{{ attribute }}</div>
```
eunjae-lee, line 10: In vue 3, `attribute: this.attribute` was not rendered on `div` properly. So I've moved it as textContent.

src/components/__tests__/MenuSelect.js
```diff
@@ -298,7 +295,7 @@ describe('custom default render', () => {
           :value="item.value"
           :selected="item.isRefined"
         >
-          {{item.label}}
+          {{item.label.toUpperCase()}}
```
eunjae-lee, line 298: (As explained below) If value attribute and the textContent are the same, Vue 3 won't render the value attribute. It makes writing tests harder. To prevent it, I made this label uppercase. See the snapshot file to see its result.

src/components/__tests__/SearchBox.js
```diff
@@ -99,30 +93,22 @@ test('with stalled search hides the submit, reset and displays the loader', () =
...
-    'hidden'
-  );
+  expect(wrapper.find('.ais-SearchBox-submit')).toBeHidden();
 
-  expect(wrapper.find('.ais-SearchBox-reset').attributes().hidden).toBe(
-    'hidden'
-  );
+  expect(wrapper.find('.ais-SearchBox-reset')).toBeHidden();
 
-  expect(wrapper.contains('.ais-SearchBox-loadingIndicator')).toBe(true);
```
eunjae-lee, line 110: `contains` removed in VTU@2.

src/mixins/__tests__/widget.test.js
```diff
@@ -24,20 +24,22 @@ describe('on root index', () => {
...
+    });
 
     mount(Test, {
       provide: {
         $_ais_instantSearchInstance: instance,
       },
-      data: () => ({
-        widgetParams,
-      }),
     });
```
eunjae-lee, line 43: It seems correct to have the widget mixin and widgetParams at the same level. The previous test worked with vue 2 (& VTU@1), but it didn't with vue 3. This change ensures it works with both versions.

jest.setup.js
```diff
@@ -13,10 +13,30 @@ expect.addSnapshotSerializer(
...
+      message: () => `expected ${wrapper} to have \`${attribute}\` attribute`,
+    };
+  }
+};
+
 expect.extend({
-  toHaveEmptyHTML: received => {
-    const html = received.html();
-    if ((isVue2 && html === '') || (isVue3 && html === '<!---->')) {
+  toHaveEmptyHTML: wrapper => {
```
Haroenv, line 34: they should all be declared outside I think, for consistency

src/components/SearchBox.vue
```diff
@@ -21,15 +21,44 @@
...
+        >
+          <slot
+            name="loading-indicator"
+          />
+        </template>
         <slot
+          v-if="isVue2"
           name="loading-indicator"
           slot="loading-indicator"
         />
```
Haroenv, line 36: what is this for? I don't understand it. Could we support only the vue-3 style with vue 2.6 support?

src/components/__tests__/DynamicWidgets.js
Haroenv: good enough, but I wonder if something like `attribute-name` could work

src/components/__tests__/HierarchicalMenu.js
```diff
@@ -435,10 +435,14 @@ describe('default render', () => {
       propsData: defaultProps,
     });
 
+    // wrapper.at(index) is deprecated at VTU@2.
+    // VTU@1: wrapper.at(index)
+    // VTU@2: wrapper[index]
+    // For simplicity, I've used `nth-child` modifier.
```
Haroenv, line 441: comment isn't needed I think, maybe we can @deprecate or remove .at in our version of VTU1?

src/components/__tests__/HierarchicalMenu.js
```diff
@@ -478,11 +482,10 @@ it('calls the Panel mixin with `items.length`', async () => {
...
   expect(mapStateToCanRefine()).toBe(true);
 
-  await wrapper.setData({
-    state: {
-      items: [],
-    },
-  });
+  // should've used wrapper.setData({ state: { items: [] }})
+  // but https://github.com/vuejs/vue-test-utils-next/issues/766
+  wrapper.vm.state.items = [];
```
Haroenv, line 487: note to also revert this once https://github.com/vuejs/vue-test-utils-next/issues/766 is released

src/components/__tests__/InstantSearch-integration.js
```diff
@@ -167,9 +176,7 @@ describe('middlewares', () => {
       uiState: { indexName: { query: 'a' } },
     });
 
-    await wrapper.setData({
-      middlewares: [middleware1],
-    });
+    wrapper.vm.middlewares = [middleware1];
```
Haroenv, line 179: also should be reverted

src/components/__tests__/InstantSearchSsr.js
```diff
@@ -98,33 +98,39 @@ it('does not dispose if not yet started', async () => {
...
     components: {
-      AisSearchBox: SearchBox,
-    },
-    slots: {
-      default: SearchBox,
+      InstantSearchSsr,
+      SearchBox,
     },
+    template: `
+      <InstantSearchSsr>
```
Haroenv, line 130: since all of these tests have been updated, does __Template-test.js also need to be updated?

src/components/__tests__/Menu.js
```diff
@@ -170,7 +170,20 @@ describe('default render', () => {
     });
 
     expect(wrapper.findAll('.ais-Menu-item--selected')).toHaveLength(1);
-    expect(wrapper.find('.ais-Menu-item--selected').text()).toBe('Samsung 25');
+    // In vue 3, .text() returned 'Samsung25',
+    // whereas in vue 2, it wass 'Samsung 25'.
+    // Thus, just checking the snapshot ↓
```
Haroenv, line 175: comment not needed I think

src/components/__tests__/MenuSelect.js
```diff
@@ -341,19 +338,46 @@ describe('custom default render', () => {
...
+    const selected = wrapper.find('[data-test="Samsung"]');
+    const options = wrapper.findAll('option:not([data-test="Samsung"])');
 
     expect(selected.element.selected).toBe(true);
 
-    options.wrappers.forEach(option => {
-      expect(option.element.selected).toBe(false);
-    });
+    for (let i = 0; i < options.length; i++) {
+      // `[i]` is for VTU@2, and `.at(i)` is for VTU@1.
```
Haroenv, line 378: does the VTU2 version work for VTU1?

src/components/__tests__/RangeInput.js
```diff
@@ -390,9 +390,8 @@ describe('refinement', () => {
     expect(refine).toHaveBeenCalledWith(['10', '100']);
 
     // update the state
-    await wrapper.setData({
-      state: { start: [50, 200] }, // min: 10 -> 50, max: 100 -> 200
-    });
+    wrapper.vm.state.start = [50, 200]; // min: 10 -> 50, max: 100 -> 200
+    await nextTick();
```
Haroenv, line 394: marker for update

src/mixins/__tests__/panel.test.js
```diff
@@ -112,13 +112,19 @@ describe('createPanelConsumerMixin', () => {
...
-    await Vue.nextTick();
+    if (isVue3) {
+      await wrapper.setData({ state: { attributeName: true } });
+    } else {
+      // ↓ this should be replaceable with `wrapper.setData()` but it didn't
+      // trigger the watcher in `createPanelConsumerMixin`.
+      // It's probably a bug from vue-test-utils.
+      // https://github.com/vuejs/vue-test-utils/issues/1756
+      // https://github.com/vuejs/vue-test-utils/issues/149
+      wrapper.vm.$set(wrapper.vm, 'state', {
```
Haroenv, line 123: does this not work in vue 3? I'd rather have less conditional code in the tests

src/util/vue-compat/index-3.js
```diff
@@ -4,5 +4,5 @@ const isVue2 = false;
 const isVue3 = true;
 const Vue2 = undefined;
 
-export * from 'vue';
+export { createApp, createSSRApp, h, version, nextTick } from 'vue';
```
Haroenv, line 7: what's this for? did it accidentally bundle other parts of vue?

Haroenv (commented):
looks great, thanks @eunjae-lee !

src/components/__tests__/DynamicWidgets.js
eunjae-lee: You're right. I didn't try `:attribute="attribute"` after changing it to template form. Now that I tried it, it just works. Now the snapshots are the same as before. https://github.com/algolia/vue-instantsearch/pull/1021/commits/9676a4621cdce5315874cf80b0eb742149a05594

jest.setup.js
eunjae-lee: https://github.com/algolia/vue-instantsearch/pull/1021/commits/a43c889402532f9bfda816fed1ffc45a9225781d

src/components/SearchBox.vue
eunjae-lee: Forgot to mention. 

`<slot>` alone didn't enable passing the slot from the parent down to `<search-input>` in vue 3. And wrapping it with `template` didn't work for vue 2.6 either. So I had to keep the both versions.

src/components/__tests__/HierarchicalMenu.js
eunjae-lee: What do you mean by removing .at? You mean at `test/utils` level? I tried to patch the `mount` function for VTU@1 but it can be 

- wrapper.at(index)
- wrapper.find(something).at(index)
- wrapper.find(something).findAll(else).at(index)

Unless we modify its prototype, it will be difficult.

src/components/__tests__/InstantSearchSsr.js
eunjae-lee: Not sure what you have in mind, but there seems not much to do about the file.
https://github.com/algolia/vue-instantsearch/blob/a43c889402532f9bfda816fed1ffc45a9225781d/src/components/__tests__/__Template.js#L10:L10

src/components/__tests__/MenuSelect.js
eunjae-lee: It's exclusive.

`[i]` only for VTU@2 and `.at(i)` only for VTU@1.

src/mixins/__tests__/panel.test.js
eunjae-lee: It works with vue 3 + VTU 2 but it doesn't with vue 2 + VTU 1 (due to VTU 1's bug). The panel is watching `state` (https://github.com/algolia/vue-instantsearch/blob/a43c889402532f9bfda816fed1ffc45a9225781d/src/mixins/panel.js#L59:L60) and that wathcer is not triggered in VTU 1. So unfortunately we have the conditional codes in the tests for the panel. We could just go for `$set` for both vue versions if you want.

src/util/vue-compat/index-3.js
eunjae-lee: Not sure why it happened but eslint couldn't find those, so I had to explicitly export them.

src/components/SearchBox.vue
Haroenv: is there a vue issue or documentation for this? I'm surprised it wasn't needed for the other slots then?

src/components/__tests__/HierarchicalMenu.js
Haroenv: ok let's just rely on the test failing and remove the comment then

src/components/__tests__/InstantSearchSsr.js
Haroenv: unlike the other tests, the template still mounts the component directly, which means it would be different with slots, should it also use template?

src/mixins/__tests__/panel.test.js
Haroenv: I think $set for both is better, leaving the comment here noting that maybe in the future it can be setData, I don't think triggering change in a different way based on a conditional is so safe in tests

src/components/__tests__/DynamicWidgets.js
Haroenv: the test on vue 3 fails now though, is there some special case for props that don't get added to elements that don't "support" it? maybe data-attribute worst case? Also putting it all in text is fine I think (maybe json stringified)

eunjae-lee: Ah okay, I've got a better understanding.
https://codesandbox.io/s/busy-roentgen-3opm3?file=/src/components/HelloWorld.vue:13-172

`attributes` as an attribute is not rendered by Vue 3 at all.

src/components/SearchBox.vue
eunjae-lee: I couldn't find anything related. And in case of Pagination, it's different from this.

What's particular about this is, for example,

There is a slot named `loading-indicator` in SearchBox. If a user passes something to that slot, SearchBox relays it to SearchInput (which accepts a slot with the same name). If nothing's passed to `loading-indicator`, then SearchInput fills with its default template.

I guess there's something to do with passing slots between components. On contrary, there's nothing like that in case of Pagination.

src/components/__tests__/InstantSearchSsr.js
eunjae-lee: Oh I get what you mean. But there are still lots of tests mounting component directly when there's nothing to do with slots. Do you see it as inconsistency?

image

src/mixins/__tests__/panel.test.js
eunjae-lee: oh, I just tried it and in vue 3, we don't have `.$set` :(

src/components/__tests__/DynamicWidgets.js
Haroenv: Screenshot 2021-07-20 at 11 34 57

`attributes` specifically is special cased it seems?

src/components/SearchBox.vue
Haroenv: ok let's go with this then :)

src/components/__tests__/DynamicWidgets.js
eunjae-lee: https://user-images.githubusercontent.com/499898/126318350-c4ff30a8-803c-4f8a-909b-487f12c8d0fe.mp4

oh I didn't know about this ↑

eunjae-lee:
FYI, I added `About the scope of this PR` section to the summary of this PR.

src/components/__tests__/DynamicWidgets.js
Haroenv: let's use `data-attribute` then :)

eunjae-lee: yeah, updated it to use JSON.stringify() https://github.com/algolia/vue-instantsearch/commit/f73aefa1b1b761178936f7eda653022eae337818

Haroenv (approved):
let's go with this for now

PR #1022, opened 2021-07-20
refactor(vue 3): add utils for compatibility

eunjae-lee:
## Summary

This PR adds `renderCompat` and `getDefaultSlot` to support compatibility between the versions and to have less conditional codes.


Haroenv (approved):
looks good, and actually is less conditional, great!

PR #1023, opened 2021-07-20
chore(vue 3): extract commands as a file to run tests with vue 3

eunjae-lee:
## Summary

This PR extracts the commands run in CircleCI as a script file to run tests with vue 3.

### DIFF
--- a/scripts/test-vue3.sh
+++ b/scripts/test-vue3.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+set -e
+
+yarn remove vue vue-jest vue-template-compiler babel-preset-es2015
+yarn add vue@3.1.2 @vue/server-renderer@3.1.2 vue-jest@5.0.0-alpha.10 vue-loader@16.1.2 jest@26.6.3 babel-jest@25.2.6 @babel/preset-env@7.14.5 @babel/plugin-transform-runtime@7.14.5 -D -E
+echo "export * from './index-3';" > src/util/vue-compat/index.js
+rm .babelrc
+cat <<EOT > babel.config.js
+// eslint-disable-next-line import/no-commonjs
+module.exports = {
+  plugins: ['@babel/plugin-transform-runtime'],
+  presets: ['@babel/preset-env'],
+};
+EOT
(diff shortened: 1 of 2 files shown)


scripts/test-vue3.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+set -e
+
+yarn remove vue vue-jest vue-template-compiler babel-preset-es2015
+yarn add vue@3.1.2 @vue/server-renderer@3.1.2 vue-jest@5.0.0-alpha.10 vue-loader@16.1.2 jest@26.6.3 babel-jest@25.2.6 @babel/preset-env@7.14.5 @babel/plugin-transform-runtime@7.14.5 -D -E
+echo "export * from './index-3';" > src/util/vue-compat/index.js
+rm .babelrc
+cat <<EOT > babel.config.js
```
Haroenv, line 8: as a babel config is conditional, how come we have to write to it, and not use an env variable to either export this or the default?

scripts/test-vue3.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+set -e
+
+yarn remove vue vue-jest vue-template-compiler babel-preset-es2015
+yarn add vue@3.1.2 @vue/server-renderer@3.1.2 vue-jest@5.0.0-alpha.10 vue-loader@16.1.2 jest@26.6.3 babel-jest@25.2.6 @babel/preset-env@7.14.5 @babel/plugin-transform-runtime@7.14.5 -D -E
+echo "export * from './index-3';" > src/util/vue-compat/index.js
+rm .babelrc
```
Haroenv, line 7: ah it's because the default is babelrc, could we use .config.js everywhere?

eunjae-lee: probably? but I got hit by an issue when using babel.config.js with vue 2:

```
import createSerializer from 'jest-serializer-html/createSerializer';                                                                                          ^^^^^^
    SyntaxError: Cannot use import statement outside a module
```

It works with `.babelrc` but not with `.config.js`.

However, in vue 3, it works only with `.config.js`. haha...

Haroenv (approved):

PR #1024, opened 2021-07-21
fix(vue 3): fixes createServerRootMixin

eunjae-lee:
## Summary

This PR fixes `createServerRootMixin`:

1. In case of vue 3, it doesn't `require` `@vue/server-renderer`, because it's not a good practice to use `require` in ESM and it obviously makes an error with `vite`. Now it dynamically imports it.

2. Fixes a regression that was not covered by our tests. Added a new test.


src/util/createServerRootMixin.js
```diff
@@ -132,13 +132,16 @@ function augmentInstantSearch(
...
+    let clonedComponentInstance;
 
     return Promise.resolve()
       .then(() => {
         app = cloneComponent(componentInstance, {
           mixins: [
             {
               created() {
+                // eslint-disable-next-line consistent-this
+                clonedComponentInstance = this;
```
Haroenv, line 144: looks like you reverted this?

eunjae-lee: yeah it turned out there was a reason why I did that in the first place, which I forgot about.

```js
app = createApp({
  created() {
    console.log(this);
  },
  methods: {
    hello() { }
  }
})
```

In vue 3, that `app` is an app instance which includes methods regarding apps like `mount, use, provide, ...`.

And `this` inside `created()` is an actual component (root component in this case) instance that includes `data` and the methods like `hello`.

src/util/createServerRootMixin.js
```diff
@@ -15,26 +15,28 @@ function walkIndex(indexWidget, visit) {
...
+    let module;
+    try {
+      module = await import('@vue/server-renderer');
+    } catch (err) {
+      // error is handled by regular if, in case it's `undefined`
+    }
+    if (!module) {
       throw new Error('you need to install @vue/server-renderer');
-    } else {
-      throw new Error('you need to install vue-server-renderer');
```
Haroenv, line 31: why is the vue 2 code removed?

wasn't this in specific files at some point?

eunjae-lee: It's still below down there.

src/util/vue-compat/index-2.js
```diff
@@ -17,6 +17,24 @@ export function getDefaultSlot(component) {
   return component.$slots.default;
 }
 
+export function renderToString(app) {
+  let _renderToString;
+  try {
+    _renderToString = require('vue-server-renderer/basic');
```
Haroenv, line 23: why not via dynamic import, does it not work?

eunjae-lee: I haven't tried dynamic import with the vue 2 part. It should work. 
However, its ESM output will include dynamic import syntax and we don't know if our customers' setup supports it. So technically speaking, it could break. So I left the vue 2 part as-is.

Haroenv: we probably should release this in a major version, for these small changes though 🤔

Haroenv, line 23: should this also be eslint-disabled or is it always present?

eunjae-lee: For now, we already have it as a dev dep, eslint doesn't complain about it.

eunjae-lee: Hmmm do you see something other than this?

Haroenv: it will make it clearer from which version vue 3 is supported, and we don't need to be as careful to prevent breaks where we think they make sense

Haroenv (approved):
good like this, but I would go for dynamic import in both code paths as a breaking change

eunjae-lee:
> good like this, but I would go for dynamic import in both code paths as a breaking change

I haven't used dynamic import in vue. I just tried it and it seems to work with vue 2 and nuxt. So I guess it's fine to use it for vue 2 and even safe not to mark it as breaking change (but just mentioning in the changelog).

vue 2: https://codesandbox.io/s/priceless-microservice-l9xux?file=/src/components/HelloWorld.vue:2056-2223

nuxt: https://codesandbox.io/s/hardcore-napier-hogj3?file=/pages/index.vue:610-777

https://github.com/algolia/vue-instantsearch/pull/1024/commits/4ecb2597d2135cf87efa1bfc446df3890669f47c

Haroenv (approved):
approved

PR #1029, opened 2021-07-29
fix(vue 3): do not pass props for cloned component in vue 3

eunjae-lee:
## Summary

This PR removes the part where it was spreading the component to pass the propsData to the cloned component in Vue 3.

Theoretically we should pass the propsData from the original component to the newly cloned component, but I don't see the concrete real life example for that. Besides, the way we spread the component isn't ideal and we get warning messages from vue.

```
      [Vue warn]: Avoid app logic that relies on enumerating keys on a component instance. The keys will be empty in production mode to avoid performance overhead.
```

If it turns out later to be a valid point with an example, we can discuss and address it.

### DIFF
--- a/src/util/__tests__/createServerRootMixin.test.js
+++ b/src/util/__tests__/createServerRootMixin.test.js
@@ -396,57 +396,56 @@ Array [
       await renderToString(wrapper);
     });
 
-    it('forwards props', async () => {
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
safer for now!

PR #1033, opened 2021-08-06
Passing algoliaStateScript and preloadLinks from render

jenjen75:
Hello,

On SRR example with vite `render` function from `entry-server.js` return an array of 3 (html, algoliaStateScript and preloadLinks)
but on `server.js`, `render` is destructured with an array of 2 (html and preloadLinks)
It is better to pass an object to prevent argument shifting.

Now the example handle algoliaStateScript + preloadLinks with running for production !


Haroenv (approved):
good catch, thanks!

PR #1035, opened 2021-08-16
fix(vue 3): let users pass renderToString to createServerRootMixin

eunjae-lee:
## Summary

This PR removes the code that dynamically importing `vue-server-renderer/basic` or `@vue/server-renderer` and now lets users pass `renderToString` function to `createServerRootMixin`.

Vue CLI and vite failed to run when those dependencies were not installed even though that importing code is never executed. Using `require` is not a good idea (especially in the ESM output). So we let users pass the function, although it's a breaking change.

### DIFF
--- a/examples/nuxt/pages/search.vue
+++ b/examples/nuxt/pages/search.vue
@@ -64,6 +64,16 @@ import {
   createServerRootMixin,
 } from 'vue-instantsearch'; // eslint-disable-line import/no-unresolved
 import algoliasearch from 'algoliasearch/lite';
+import _renderToString from 'vue-server-renderer/basic';
+
+function renderToString(app) {
+  return new Promise((resolve, reject) => {
+    _renderToString(app, (err, res) => {
+      if (err) reject(err);
+      resolve(res);
+    });
+  });
+}
 
 const searchClient = algoliasearch(
   'latency',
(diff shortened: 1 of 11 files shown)


examples/nuxt/pages/search.vue
```diff
@@ -64,6 +64,16 @@ import {
   createServerRootMixin,
 } from 'vue-instantsearch'; // eslint-disable-line import/no-unresolved
 import algoliasearch from 'algoliasearch/lite';
+import _renderToString from 'vue-server-renderer/basic';
+
+function renderToString(app) {
+  return new Promise((resolve, reject) => {
```
Haroenv, line 70: I think we should accept both the callback form and the promise form (instanceof Promise to detect maybe) to avoid this glue code?

eunjae-lee: Do you mean that we should promisify it inside createServerRootMixin in case it's vue 2?

Haroenv: I think that's an option, but not sure. What do you think?

eunjae-lee:
```js
const result = renderToString(app);
if (result instanceof Promise) {
  ...
}     
```

but at that point, it's already invoked without a callback function. We need to know if the function is going to take a callback or just return a promise. So the only condition I can think of is to check vue version.

Handling callback vs promise based on vue version being used, it feels less adjacent (cannot think of the expression). what do you think?

Haroenv (approved):
this makes sense, as we decided to make it a breaking change. Don't forget to update all server side rendering docs and examples after this of course :)

PR #1036, opened 2021-08-17
fix(range-input): remove default min and max values from `ais-range-input`

eunjae-lee:
## Summary

This PR removes the default min and max values from `ais-range-input`.

BREAKING CHANGE, but it won't likely affect the behavior.

### DIFF
--- a/src/components/RangeInput.vue
+++ b/src/components/RangeInput.vue
@@ -80,14 +80,12 @@ export default {
     min: {
       type: Number,
       required: false,
-      // @major: remove this default
-      default: -Infinity,
+      default: undefined,
     },
     max: {
       type: Number,
       required: false,
-      // @major: remove this default
-      default: Infinity,
+      default: undefined,
     },
     precision: {
       type: Number,


eunjae-lee:
It has nothing to do with vue 3 compat, but we're not maintaining `next` branch at this moment and all the breaking changes are on `feat/vue3-compat`, so I'm basing this PR on it.

src/components/RangeInput.vue
```diff
@@ -80,14 +80,12 @@ export default {
     min: {
       type: Number,
       required: false,
-      // @major: remove this default
-      default: -Infinity,
```
shortcuts, line 84: Not related to the changes itself, but do you know why we had this in the first place?

eunjae-lee: Not sure, probably as a safety net? Anyway [in InstantSearch.js](https://github.com/algolia/instantsearch.js/blob/4bea07b99f492441eb94e483378e0778f90c5b43/src/connectors/range/connectRange.ts#L80:L89), we assume it's optional and take care of it.

francoischalifour (approved):

Haroenv (approved):

src/components/RangeInput.vue
Haroenv: Yes, that's why, there were a couple places where we repeat the default values as InstantSearch as props, but this one couldn't be removed without breaking change

shortcuts (approved):

src/components/RangeInput.vue
shortcuts: Ok thanks!

PR #1037, opened 2021-08-17
fix(state-results): clean up slot props

eunjae-lee:
## Summary

This PR cleans up the slot props to `ais-state-results` component.

```vue
<ais-state-results>
  <template v-slot="myProps">

  </template>
</ais-state-results>
```

Here, `myProps` used to be something like this:

```js
myProps = {
  ...results,
  results,
  state
}
```

It was an object with `results` and `state` keys and also had `results` object spread.

Now it is an object with only `results` and `state` keys.

### How does it change?

If there is the following code:

```vue
<ais-state-results>
  <template v-slot="{ state, results }">
    // do something with `state` or `results`
  </template>
</ais-state-results>
```

then it's fine. Nothing is broken.

However, 

```vue
<ais-state-results>
  <template v-slot="{ hits }">

  </template>
</ais-state-results>
```

If it was accessing something from results like that, it should be replaced with `v-slot="{ results: { hits } }"`

BREAKING CHANGE

### DIFF
--- a/src/components/StateResults.vue
+++ b/src/components/StateResults.vue
@@ -3,7 +3,10 @@
     :class="suit()"
     v-if="state && state.state && state.results"
   >
-    <slot v-bind="stateResults">
+    <slot
+      :state="state.state"
+      :results="state.results"
+    >
       <p>
         Use this component to have a different layout based on a certain state.
       </p>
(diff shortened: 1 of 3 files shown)


Haroenv:
This is good, but possibly should be paused, as I really like @francoischalifour' suggestion splitting state and results into separate getters, and we should probably do that in Vue InstantSearch too

eunjae-lee:
meaning `ais-state-results` should be broken into two components in the end, IIUC?

Haroenv:
I guess so, but it could still be interesting to do that in Vue InstantSearch v4

eunjae-lee:
Yes, for now, we could leave `ais-state-results` as-is. Later we could deprecate it and introduce new things.

PR #1040, opened 2021-08-19
chore(vue 3): shorten import path for vue3

eunjae-lee:
## Summary

This PR shortens the import path for vue3

from

```js
import InstantSearch from 'vue-instantsearch/dist/vue3/es';
```

to

```js
import InstantSearch from 'vue-instantsearch/vue3/es';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "version": "4.0.0-beta.1",
   "files": [
     "dist",
+    "vue3",
     "src"
   ],
   "main": "dist/vue2/cjs",
@@ -26,7 +27,7 @@
   "sideEffects": false,
   "repository": "https://github.com/algolia/vue-instantsearch",
   "scripts": {
-    "prebuild": "rm -rf dist",
+    "prebuild": "rm -rf dist && rm -rf vue3",
     "build": "yarn build:vue2 && yarn build:vue3",
     "build:vue2": "./scripts/build-vue2.sh",
     "build:vue3": "./scripts/build-vue3.sh",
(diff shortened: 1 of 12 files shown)


rollup.config.js
```diff
@@ -20,8 +20,7 @@ const processEnv = conf => ({
 });
 
 const vuePlugin = process.env.VUE_VERSION === 'vue3' ? vueV3 : vueV2;
-const outputDir =
-  process.env.VUE_VERSION === 'vue3' ? 'dist/vue3' : 'dist/vue2';
+const outputDir = process.env.VUE_VERSION === 'vue3' ? 'vue3' : 'dist/vue2';
```
Haroenv, line 23: should it not also be /vue2 then? instead of dist/vue2?

eunjae-lee: yeah let's go with that

eunjae-lee:
The Netlify build failure is expected. (will be fixed later)

package.json
```diff
@@ -18,15 +18,16 @@
...
   "files": [
-    "dist",
+    "vue2",
+    "vue3",
     "src"
   ],
-  "main": "dist/vue2/cjs",
-  "module": "dist/vue2/es",
+  "main": "vue2/cjs",
+  "module": "vue2/es",
```
Haroenv, line 26: was already the case, but why not vue2/cjs/index.js ?

package.json
```diff
@@ -18,15 +18,16 @@
...
   "sideEffects": false,
   "repository": "https://github.com/algolia/vue-instantsearch",
   "scripts": {
-    "prebuild": "rm -rf dist",
+    "prebuild": "rm -rf vue2 && rm -rf vue3",
```
Haroenv, line 30:
```suggestion
    "prebuild": "rm -rf vue2 vue3",
```

Haroenv (approved):

package.json
eunjae-lee: any reason or concern not having `/index.js` in the path?

Haroenv: I guess it's supported, but I assumed that module and main were files, not module identifiers. If it works, then there's probably no issue

PR #1041, opened 2021-08-19
chore(vue 3): add non-ssr examples

eunjae-lee:
## Summary

This PR adds two examples for non-ssr usage (Vue CLI + vite).


Haroenv:
should these maybe live in doc-code-samples so they don't slow down netlify as much (or ignored from the build)

eunjae-lee:
> should these maybe live in doc-code-samples so they don't slow down netlify as much (or ignored from the build)

Yeah I wasn't sure if this should go there or here. I'll move it over there.

eunjae-lee:
@Haroenv but what about ssr examples we already have in this repo? does it also make sense to move them over doc-code-samples?

Haroenv:
Yes, I think probably all examples should move to doc-code-samples, except e-commerce as it's used for e2e

PR #1042, opened 2021-08-19
chore: remove examples

eunjae-lee:
## Summary

This PR removes examples because they're moved to [doc-code-samples](https://github.com/algolia/doc-code-samples/pull/245).


Haroenv:
what about the other examples, we're keeping them because they're linked in the documentation and removing later?

eunjae-lee:
> what about the other examples, we're keeping them because they're linked in the documentation and removing later?

yes, I don't know how much/where they're spread at the moment, so we can keep them as-is and clean it up later (including build scripts around netlify, etc)

Haroenv (approved):

PR #1044, opened 2021-08-20
fix(vue 3): pass renderToString to findResultsState instead of createServerRootMixin

eunjae-lee:
## Summary

This PR fixes the regression from #1035.

## Details

Before #1035, we used to dynamically import `@vue/server-renderer` and called `renderToString()`, but it was only inside `search.findResultsState()`.

After #1035, importing `@vue/server-renderer` is on the user land, but it happens at `main.js` file where it exports `createApp` function.
And the error now happens because it is unnecessarily trying to import `@vue/server-renderer` on the browser and it's not meant to be (`main.js` is used in both client and server).

The solution here is to pass from entry-server.js `renderToString` to `createApp()` , because `serverPrefetch()` is called only on server, and entry-client.js not passing it won't make any issue.

This new API will help people (like me) not put server renderer in client bundle mistakenly, which is hard to figure out afterwards.

### DIFF
--- a/src/util/__tests__/createServerRootMixin.test.js
+++ b/src/util/__tests__/createServerRootMixin.test.js
@@ -182,9 +182,9 @@ describe('createServerRootMixin', () => {
         ),
         serverPrefetch() {
           expect(() =>
-            this.instantsearch.findResultsState(this)
+            this.instantsearch.findResultsState({ component: this })
           ).toThrowErrorMatchingInlineSnapshot(
-            `"findResultsState requires \`renderToString: (componentInstance) => Promise<string>\` in the second argument."`
+            `"findResultsState requires \`renderToString: (component) => Promise<string>\` in the first argument."`
           );
         },
       };
@@ -231,7 +231,10 @@ describe('createServerRootMixin', () => {
           ])
         ),
         serverPrefetch() {
-          return this.instantsearch.findResultsState(this, renderToString);
+          return this.instantsearch.findResultsState({
+            component: this,
+            renderToString,
+          });
         },
         created() {
           mainIndex = this.instantsearch.mainIndex;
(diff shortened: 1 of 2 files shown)


src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -52,7 +52,6 @@ describe('createServerRootMixin', () => {
         createSSRApp({
           mixins: [
             createServerRootMixin({
-              renderToString,
```
Haroenv, line 55: we could still make the signature to `importRenderToString: () => import('render-to-string')` that should be tree-shaken in mai

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -214,7 +231,7 @@ describe('createServerRootMixin', () => {
           ])
         ),
         serverPrefetch() {
-          return this.instantsearch.findResultsState(this);
+          return this.instantsearch.findResultsState(this, renderToString);
```
Haroenv, line 234: is this the best signature, should it maybe be `{ component: this, renderToString }`?

src/util/__tests__/createServerRootMixin.test.js
eunjae-lee: It's imported only on server and from `entry-server.js`. So I'm not sure how that'd help with tree-shaking. And even if we make the signature like that, it'd be better to pass it directly to `findResultsState`, because previously I had to pass `renderToString` all the way from createServerRootMixin to augmentInstantSearch → ...

src/util/__tests__/createServerRootMixin.test.js
eunjae-lee: yeah it looks better!

Haroenv (approved):
let's go with this, don't forget to update the docs :)

PR #1052, opened 2021-09-01
feat: add experimental composition APIs

eunjae-lee:
## Summary

This PR adds experimental composition APIs.

For now, useConfigure, useRefinementList, useSearchBox, and useConnector are implemented.

In case of `useRefinementList`, I referred to François' PoC with React InstantSearch.

### vue 3

```vue
<template>
  <div>
    <p>query: {{ searchBox.query }}</p>
  </div>
</template>

<script>
import { EXPERIMENTAL_useSearchBox } from 'vue-instantsearch/vue3/es';

export default {
  setup() {
    return { searchBox: EXPERIMENTAL_useSearchBox() };
  }
}
```

### vue 2

```sh
yarn add @vue/composition-api
# or
npm install @vue/composition-api
```

```js
// main.js
import Vue from 'vue';
import VueCompositionAPI from '@vue/composition-api';

Vue.use(VueCompositionAPI);
```

```vue
// Child.vue

<template>
  <div>
    <p>query: {{ searchBox.query }}</p>
  </div>
</template>

<script>
import { EXPERIMENTAL_useSearchBox } from 'vue-instantsearch/vue2/es/compositions';
// Import composition APIs from the path above.
// The rest still remain as `import xxx from 'vue-instantsearch'`.

export default {
  setup() {
    return { searchBox: EXPERIMENTAL_useSearchBox() };
  }
}
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -164,7 +164,6 @@ export default [
               format: 'es',
             },
           ],
-          // preserveModules: true,
           plugins: [applyVueCompatForVue2Compositions, ...plugins],
         },
       ]
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -8,7 +8,10 @@ import { terser } from 'rollup-plugin-terser';
 import replace from 'rollup-plugin-replace';
 import json from 'rollup-plugin-json';
 
-if (process.env.VUE_VERSION !== 'vue2' && process.env.VUE_VERSION !== 'vue3') {
+const isVue2 = process.env.VUE_VERSION === 'vue2';
+const isVue3 = process.env.VUE_VERSION === 'vue3';
+
+if (!isVue2 && !isVue3) {
   throw new Error(
     'The environment variable VUE_VERSION (`vue2` | `vue3`) is required.'
   );
@@ -19,10 +22,20 @@ const processEnv = conf => ({
   'process.env': `(${JSON.stringify(conf)})`,
 });
 
-const vuePlugin = process.env.VUE_VERSION === 'vue3' ? vueV3 : vueV2;
-const outputDir = process.env.VUE_VERSION === 'vue3' ? 'vue3' : 'vue2';
+const vuePlugin = isVue3 ? vueV3 : vueV2;
+const outputDir = isVue3 ? 'vue3' : 'vue2';
+
+const excludeCompositionsAPI = {
+  load(id){
+    if (id.endsWith("/src/compositions/index.js")) {
+      return ''
+    }
+    return null;
+  }
+}
 
 const plugins = [
+  isVue2 && excludeCompositionsAPI,
   vuePlugin({ compileTemplate: true, css: false }),
   commonjs(),
   json(),
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -7,6 +7,7 @@ import commonjs from 'rollup-plugin-commonjs';
 import { terser } from 'rollup-plugin-terser';
 import replace from 'rollup-plugin-replace';
 import json from 'rollup-plugin-json';
+import fs from 'fs';
 
 const isVue2 = process.env.VUE_VERSION === 'vue2';
 const isVue3 = process.env.VUE_VERSION === 'vue3';
@@ -29,6 +30,30 @@ const excludeCompositionsAPI = {
   load(id) {
     if (id.endsWith('/src/compositions/index.js')) {
       return '';
+    }
+    return null;
+  },
+};
+
+const applyVueCompat = {
+  load(id) {
+    if (id.endsWith('src/util/vue-compat/index.js')) {
+      if (isVue2) {
+        return fs.readFileSync('./src/util/vue-compat/index-2.js').toString();
+      } else if (isVue3) {
+        return fs.readFileSync('./src/util/vue-compat/index-3.js').toString();
+      }
+    }
+    return null;
+  },
+};
+
+const applyVueCompatForVue2Compositions = {
+  load(id) {
+    if (id.endsWith('src/util/vue-compat/index.js')) {
+      return fs
+        .readFileSync('./src/util/vue-compat/index-2-compositions.js')
+        .toString();
     }
     return null;
   },
@@ -67,7 +92,11 @@ export default [
         exports: 'named',
       },
     ],
-    plugins: [isVue2 && excludeCompositionsAPI, ...plugins].filter(Boolean),
+    plugins: [
+      isVue2 && excludeCompositionsAPI,
+      applyVueCompat,
+      ...plugins,
+    ].filter(Boolean),
   },
   {
     input: 'src/instantsearch.js',
@@ -80,7 +109,11 @@ export default [
       },
     ],
     preserveModules: true,
-    plugins: [isVue2 && excludeCompositionsAPI, ...plugins].filter(Boolean),
+    plugins: [
+      isVue2 && excludeCompositionsAPI,
+      applyVueCompat,
+      ...plugins,
+    ].filter(Boolean),
   },
   {
     input: 'src/instantsearch.umd.js',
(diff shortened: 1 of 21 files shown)


Haroenv:
I thought you could use the composition api in vue 2 with @vue/composition-api or something like that?

eunjae-lee:
> I thought you could use the composition api in vue 2 with @vue/composition-api or something like that?

You're right. I haven't figured out how to accept that optional dependency for vue 2 yet. So I'm just starting with Vue 3 for now.

rollup.config.js
```diff
@@ -19,8 +23,41 @@ const processEnv = conf => ({
...
 
-const vuePlugin = process.env.VUE_VERSION === 'vue3' ? vueV3 : vueV2;
-const outputDir = process.env.VUE_VERSION === 'vue3' ? 'vue3' : 'vue2';
+const vuePlugin = isVue3 ? vueV3 : vueV2;
+const outputDir = isVue3 ? 'vue3' : 'vue2';
+
+const excludeCompositionsAPI = {
+  load(id) {
+    if (id.endsWith('/src/compositions/index.js')) {
+      return '';
```
eunjae-lee, line 32: Import with `'/src/compositions/index.js'` gets replaced with an empty string. It's to exclude this line from Vue 2 bundle:
https://github.com/algolia/vue-instantsearch/blob/58618c5609c17e8feb3f842dfa2a30b2cb72721b/src/instantsearch.js#L7

rollup.config.js
```diff
@@ -19,8 +23,41 @@ const processEnv = conf => ({
...
+    }
+    return null;
+  },
+};
+
+const applyVueCompat = {
```
eunjae-lee, line 38: Instead of modifying vue-compat/index.js at build time, we can just replace it with index-2 or -3 file on rollup, which seems better.

rollup.config.js
```diff
@@ -19,8 +23,41 @@ const processEnv = conf => ({
...
+        return fs.readFileSync('./src/util/vue-compat/index-2.js').toString();
+      } else if (isVue3) {
+        return fs.readFileSync('./src/util/vue-compat/index-3.js').toString();
+      }
+    }
+    return null;
+  },
+};
+
+const applyVueCompatForVue2Compositions = {
```
eunjae-lee, line 51: When bundling `vue-instantsearch/compositions` for vue 2, we need a different file for `vue-compat` util because we need to bind `@vue/compositions` api.

src/compositions/__tests__/useConnector.js
```diff
@@ -0,0 +1,288 @@
+jest.unmock('instantsearch.js/es');
+import InstantSearch from '../../components/InstantSearch';
+import Index from '../../components/Index';
+import { createFakeClient } from '../../util/testutils/client';
+import { isVue3, isVue2, ref } from '../../util/vue-compat';
```
eunjae-lee, line 5: This `vue-compat` needs to be replaced with `vue-compat/index-2-compositions.js` in case of vue 2. I don't know how to mock it that way. I had some issues with jest.mock().

For now, we're testing only on vue 3.

Haroenv: could it be required inside the conditional, or a file for vue 3 and a file for vue 2?

src/compositions/useConnector.js
```diff
@@ -0,0 +1,48 @@
...
+  onMounted,
+  onUnmounted,
+} from '../util/vue-compat';
+
+export function useConnector(connector, widgetParams, initialState) {
+  const instantSearchInstance = inject('$_ais_instantSearchInstance');
+  const getParentIndex = inject('$_ais_getParentIndex', undefined);
+  const indexWidget = getParentIndex && getParentIndex();
+  const parent = indexWidget || instantSearchInstance;
+  const state = ref(initialState);
```
Haroenv, line 15: this is different to the widget mixin, I think it's not a bad idea per se, but not sure if we should introduce that parameter now or later

src/util/noop.js
```diff
@@ -0,0 +1 @@
+export const noop = () => {};
```
Haroenv, line 1: in the rest of the code base, noop is written inline, just adding this would be inconsistent probably

scripts/build-vue2.sh
```diff
@@ -1,15 +1,7 @@
 #!/bin/sh
 set -e
 
-originalContent=`cat src/util/vue-compat/index.js`
-
-# use v2 for src/util/vue-compat
-echo "export * from './index-2';" > src/util/vue-compat/index.js
```
Haroenv, line 7: what makes this not needed anymore?

Haroenv: ah I see it's in the rollup config, much better!

rollup.config.js
```diff
@@ -86,11 +131,42 @@ export default [
...
+          input: 'src/compositions-for-vue2.js',
+          external,
+          output: [
+            {
+              sourcemap: true,
+              file: `${outputDir}/es/compositions.js`,
+              format: 'es',
+            },
+          ],
+          // preserveModules: true,
```
Haroenv, line 167: should this comment be removed?

eunjae-lee: ah yes!

rollup.config.js
```diff
@@ -86,11 +131,42 @@ export default [
       },
     ],
     plugins: [
+      applyVueCompat,
```
Haroenv, line 134: looks like this plugin is always used, so it can be part of plugins?

src/compositions/useConnector.js
eunjae-lee: Agree. I'm not sure if this is the best. It'd be good if we somehow can get initial state from each connector, so that we could use it here. Otherwise, widget's init happens in the next tick, meaning `useXYZ()` will have to return null for the first time. And it seems troublesome for users to always check if it's null.

Probably we could do something like

```js
// connectXYZ.js

export const initialState = {

}

export default function connectXYZ() { ... }
```

That way we get to keep the current shape of connectors, and also get the initial state without having to duplicate code over vue and react instant search.

src/compositions/__tests__/useConnector.js
eunjae-lee: this `vue-compat` is also imported within the source `useConnector.js`

rollup.config.js
eunjae-lee: not always, read line 142~170

src/compositions/useConnector.js
Haroenv: I actually thought a little more about it, and I think it makes more sense to have the state as null initially (like the mixin) to make integration for each connector easier (we don't even have to ship those initially)

Baroshem:
Bumping this topic as it looks great!

I just released an Algolia module for Nuxt 3 -> https://algolia-nc.netlify.app/ 
And I was looking to include vue-instantsearch there as well. 

If you can finish this PR and make these components Vue 3 compatible, I can try to import them to the module as well so that the users can use your components in Nuxt 3 as well 😉

Haroenv:
@Baroshem, the components are already vue 3 compatible if you import from `vue-instantsearch/vue3`, but that's indeed not really properly documented yet

Baroshem:
@Haroenv I see. I will try then to use them in the module so that they can be imported into the Nuxt 3 right now. Thanks for claryfing!

Baroshem:
Hi @Haroenv 

I tried yesterday to import vue-instantsearch/vue3 but failed after many attempts to fix the errors. I had issues with 
* "type": "module" in package.json (it was not there so I had to add manually)
* directory imports
* external packages imports (instantsearch.js).

I have very little experience with the source code of vue-instantsearch so I am not sure if the changes I was doing are even correct.

Could you please try to install vue-instantseach/vue3 on fresh Nuxt 3 project and see how it works for you?

PR #1061, opened 2021-09-15
feat(dynamic): add fallback-component

Haroenv:
```vue
<ais-dynamic-widgets>
  <template v-slot:fallback-component="{ attribute }">
    <ais-panel>
      <template v-slot:header>{{ attribute }}</template>
      <template v-slot>
      </template>
    </ais-panel>
  </template>
</ais-dynamic-widgets>
```

still WIP as i didn't figure out how to make ssr work with it

FX-5

### DIFF
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -54,6 +54,10 @@ export default {
   render: renderCompat(function(h) {
     const components = new Map();
 
+    const fallbackComponent =
+      (isVue3 ? this.$slots : this.$scopedSlots)['fallback-component'] ||
+      (() => undefined);
+
     (getDefaultSlot(this) || []).forEach(vnode => {
       const attribute = getWidgetAttribute(vnode);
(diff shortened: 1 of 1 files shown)


src/components/DynamicWidgets.js
```diff
@@ -65,6 +69,7 @@ export default {
     });
 
     // by default, render everything, but hidden so that the routing doesn't disappear
+    // TODO: make sure fallback component gets rendered here too, but how do we know which attributes?
```
Haroenv, line 72: I tried to find a similar solution with rendering/searching an extra time like in React InstantSearch, but couldn't find it :(

francoischalifour: What are the fundamental differences between CSR and SSR in the Dynamic widget?

Haroenv:
I'll leave this feature out in Vue InstantSearch for now, you can still use the connector to create dynamic experiences with a fallback.

PR #1090, opened 2021-12-02
feat(ssr): prevent initial network request

Haroenv:
FX-623

relies on https://github.com/algolia/instantsearch.js/pull/4967

to do:
- [ ] is there other code that should be removed?
- [ ] does this work in all cases?


src/util/createServerRootMixin.js
```diff
@@ -130,37 +130,21 @@ function augmentInstantSearch(
       .then(() => renderToString(app))
       .then(() => searchOnlyWithDerivedHelpers(helper))
       .then(() => {
-        const results = {};
-        walkIndex(renderedComponent.instantsearch.mainIndex, widget => {
-          results[widget.getIndexId()] = widget.getResults();
+        resultsState = {};
+        walkIndex(instance.mainIndex, widget => {
+          const { _state, _rawResults } = widget.getResults();
```
Haroenv, line 135: same as the react implementation, but I wonder if this instead was:

```js
const state = widget.getWidgetSearchParameters(new SearchParameters(), { uiState: search._initialUiState[widget.getIndexId()] }
```

if we then still have the wrong parameters available `helper.state`, or if that's correct then?

src/util/createServerRootMixin.js
```diff
@@ -205,7 +187,10 @@ function augmentInstantSearch(
       results,
       scopedResults: parent.getScopedResults().map(result =>
         Object.assign(result, {
-          results: search.__initialSearchResults[result.indexId],
+          results: new SearchResults(
+            new SearchParameters(search._initialResults[result.indexId].state),
+            search._initialResults[result.indexId].results
+          ),
```
Haroenv, line 193: this is a little inefficient, maybe at this point they are already set on the parent?

package.json
```diff
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "algoliasearch-helper": "^3.6.2",
-    "instantsearch.js": "^4.32.0",
+    "instantsearch.js": "https://pkg.csb.dev/algolia/instantsearch.js/commit/3f954126/instantsearch.js",
```
Haroenv, line 52: obviously we need to wait until algolia/instantsearch.js#4967 is released

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -693,57 +691,20 @@ Array [
...
+        })
       );
 
-      expect(instantsearch.__initialSearchResults.hello).toEqual(
+      expect(instantsearch._initialResults.hello).toEqual(
         expect.objectContaining(serialized)
       );
     });
 
-    it('accepts non-stringified results', () => {
```
Haroenv, line 705: this is no longer relevant, hydrate only accepts "stringified" results

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -834,6 +793,7 @@ Array [
         created() {
           instantSearchInstance = this.instantsearch;
         },
+        render() {},
```
Haroenv, line 796: these tests were warning without render method, unrelated to the PR

src/util/createServerRootMixin.js
francoischalifour: Wdym?

src/util/createServerRootMixin.js
```diff
@@ -104,63 +94,44 @@ function augmentInstantSearch(
...
         });
       })
       .then(() => renderToString(app))
-      .then(() => searchOnlyWithDerivedHelpers(helper))
+      .then(() => searchOnlyWithDerivedHelpers(instance.mainHelper))
       .then(() => {
-        const results = {};
-        walkIndex(renderedComponent.instantsearch.mainIndex, widget => {
-          results[widget.getIndexId()] = widget.getResults();
+        resultsState = {};
```
francoischalifour, line 119: We call this `initialResults` in InstantSearch now.

src/util/createServerRootMixin.js
```diff
@@ -184,30 +155,23 @@ function augmentInstantSearch(
...
-    const results = search.__initialSearchResults[parent.getIndexId()];
-
     // this happens when a different InstantSearch gets rendered initially,
     // after the hydrate finished. There's thus no initial results available.
-    if (!results) {
+    if (parent.getResults() === null) {
       return;
     }
-
+    const results = parent.getResults();
```
francoischalifour, line 163: We can declare this variable above to also use it in the condition.

package.json
```diff
@@ -82,6 +81,7 @@
     "@wdio/spec-reporter": "^5.11.7",
     "@wdio/static-server-service": "^5.11.0",
     "algoliasearch": "4.0.1",
+    "algoliasearch-helper": "3.6.2",
```
francoischalifour, line 84: Why this change?

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -660,7 +659,7 @@ Array [
   });
 
   describe('hydrate', () => {
-    it('sets __initialSearchResults', () => {
+    it('sets _initialResults', () => {
```
francoischalifour, line 662: It would be better to test the consequence of setting `_initialResults` rather than setting it.

src/util/createServerRootMixin.js
```diff
@@ -104,63 +94,44 @@ function augmentInstantSearch(
...
 
-                this.instantsearch.mainIndex.init({
-                  instantSearchInstance: this.instantsearch,
-                  parent: null,
-                  uiState: this.instantsearch._initialUiState,
-                });
+                instance.start();
+                // although we use start for initializing the main index,
+                // we don't want to send search requests yet
+                instance.started = false;
```
francoischalifour, line 110: Do we have a test to make sure that the network request doesn't happen?

src/util/createServerRootMixin.js
Haroenv: instead of taking the search parameters from the results (as is done now, and in React InstantSearch too), we could retrieve the "local" index search parameters by using getWidgetSearchParameters of each index, although it likely still would not work completely out of the box, as by default it's recursive. If that could be avoided, there would be a way to get the correct search parameters that can be applied to helper.state without issue

src/util/createServerRootMixin.js
Haroenv: Without this one of the tests asserts that search has only been called once, but there's indeed no explicit test

src/util/__tests__/createServerRootMixin.test.js
Haroenv: The later tests (setting helper, and init index) already test that, right? Do you have an idea which test to write?

src/util/createServerRootMixin.js
Haroenv: 6e69caf

src/util/createServerRootMixin.js
Haroenv: 6e69caf

src/util/createServerRootMixin.js
Haroenv: 904bde4

src/util/createServerRootMixin.js
francoischalifour: I don't see how that's easier than using `getResults()`?

src/util/__tests__/createServerRootMixin.test.js
francoischalifour: I think that this test covers it: [`904bde4` (#1090)](https://github.com/algolia/vue-instantsearch/pull/1090/commits/904bde4464dd1c90c787fadf51de7937efc3f9e6#diff-7f49c18a452238e67a534ec6a7197429b8cdc0da21ed98ea38e1ec85ece88201R923)

We may not need this test altogether if the behavior is tested at a higher level elsewhere then.

francoischalifour (approved):

src/util/createServerRootMixin.js
Haroenv: I didn't mean easier, I meant more correct. I was trying to find a way to get only the "local index search parameters", but indeed what I was suggesting wasn't correct, so I guess it doesn't matter much

PR #1104, opened 2022-01-28
fix(ssr): extend component correctly if at root (vue2)

### LINKED ISSUE #1054: SSR is broken
**Bug 🐞**

The SSR example causes an error, which I'm also receiving in my own implementation.

```
TypeError: (intermediate value)(intermediate value)(intermediate value) is not a constructor
    at a (node_modules/vue-instantsearch/vue2/es/src/util/createServerRootMixin.js:1:363)
    at app.js:14981:1049
    at async Promise.all (index 0)
```

https://github.com/algolia/vue-instantsearch/tree/master/examples/ssr

The only things I changed, are:

```
   "dependencies": {
     "algoliasearch": "^4.0.1",
     "instantsearch.css": "^7.1.1",
-    "vue": "^2.6.7",
-    "vue-instantsearch": "file:../../",
+    "vue": "2.6.11",
+    "vue-instantsearch": "^4.0.0",
     "vue-router": "^3.0.2",
-    "vue-server-renderer": "^2.6.11"
+    "vue-server-renderer": "2.6.11"
   },
   "devDependencies": {
     "@akryum/vue-cli-plugin-ssr": "^0.5.0",
@@ -25,7 +25,7 @@
     "babel-eslint": "^10.0.1",
     "eslint": "^5.14.1",
     "eslint-plugin-vue": "^5.2.2",
-    "vue-template-compiler": "^2.6.7"
+    "vue-template-compiler": "2.6.11"
   },
```

Not receiving that error

No

Seems like the code which errors is trying to use the `this` passed in `findResultsState` as a constructor. I don't understand the context, so that's as far as I got :-)

```js
    serverPrefetch() {
      return this.instantsearch.findResultsState({
        component: this, // I think the value passed here is being used as a constructor, which it isn't
        renderToString
      });
    },
```

4.0.0

Haroenv:
fixes #1054

Essentially the problem is that $vnode is usually available, but not when the this is a root Vue instance. In that case we are in the "Vue.component" case, which before now always was wrong (it takes two arguments, not one)


dhayab (approved):
👍

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -719,6 +719,82 @@ Array [
     },
   },
 ]
+`);
+      });
+
+      it('works if component has no $vnode', async () => {
```
sarahdayan, line 725: It's currently hard to know why there's no `$vnode` in this use case. Can we either comment, or add an assertion that makes it obvious?

Haroenv:
```suggestion
      it('works if component is at root, and has no $vnode', async () => {
```

is this clearer @sarahdayan ?

sarahdayan:
```suggestion
      it('works when component is at root (and therefore has no $vnode)', async () => {
```

sarahdayan (approved):

PR #1105, opened 2022-01-31
fix(server): extend component correctly if at root

Haroenv:
Essentially the problem is that $vnode is usually available, but not when the this is a root Vue instance. In that case we are in the "Vue.component" case, which before now always was wrong (it takes two arguments, not one)

Same as #1104
see also #1054

### DIFF
--- a/src/util/__tests__/createServerRootMixin.test.js
+++ b/src/util/__tests__/createServerRootMixin.test.js
@@ -682,6 +682,78 @@ Array [
       expect(instantsearch.helper).toEqual(expect.any(AlgoliaSearchHelper));
       expect(instantsearch.mainHelper).toEqual(expect.any(AlgoliaSearchHelper));
     });
+
+    it('works if component has no $vnode', async () => {
+      const searchClient = createFakeClient();
+      let mainIndex;
+
+      const app = {
+        render: h =>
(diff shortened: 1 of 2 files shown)


src/util/createServerRootMixin.js
```diff
@@ -239,6 +242,8 @@ function augmentInstantSearch(
       parent: null,
       uiState: search._initialUiState,
     });
+
+    search.middleware.forEach(({ instance }) => instance.subscribe());
```
Haroenv, line 246:
```suggestion
```

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -682,6 +682,78 @@ Array [
       expect(instantsearch.helper).toEqual(expect.any(AlgoliaSearchHelper));
       expect(instantsearch.mainHelper).toEqual(expect.any(AlgoliaSearchHelper));
     });
+
+    it('works when component is at root (and therefore has no $vnode)vnode', async () => {
```
sarahdayan, line 686:
```suggestion
    it('works when component is at root (and therefore has no $vnode)', async () => {
```

sarahdayan (approved):

Issue #1066, opened 2021-09-30
Nuxt Instant Search broken when used with router and index mapping


podlebar:
**Bug 🐞**

### What is the current behavior?
When used with Nuxt combined with routing and singleIndexMapping the query parameters are removed after hitting refresh. 

when this is added on line 106 in the sandbox:
```
stateMapping: singleIndexMapping("instant_search"),
```
the route on click on a facet does not change at all.

### Make a sandbox with the current behavior
https://codesandbox.io/s/friendly-parm-hy2m2?file=/pages/search.vue

### What is the expected behavior?
1. Clock on Apple in the facets
2. Url changes to https://hy2m2.sse.codesandbox.io/search?refinement%5BrefinementList%5D%5Bbrand%5D%5B0%5D=Apple
3. hit the refresh button
4. the url stays same and the UiState is recreated

### Does this happen only in specific situations?
Happens all the time

### What is the version you are using?
algoliasarch 4.10.5
nuxt 2.15.8
vue-instantsearch 4.0.1
vue-server-renderer 2.6.11

EduardoMateos:
same here ;(

podlebar:
@EduardoMateos did you find a solution? it seems that nobody cares here :)

podlebar:
@EduardoMateos did you find any solution?

EduardoMateos:
Hello,

I fixed it temporarily, it is not the best solution but this works for me. I get the parameters from the url manually to initialize Algolia.

im use Nuxt 2.15.8,  IMPORTANT: my index name is "product"

Code:



```javascript
import {
  AisInstantSearchSsr,
  // AisRefinementList,
  AisInfiniteHits,
  AisHighlight,
  AisSearchBox,
  AisStats,
  AisPagination,
  AisSnippet,
  AisStateResults,
  AisPoweredBy,
  createServerRootMixin,
} from "vue-instantsearch";

import algoliasearch from "algoliasearch/lite";
import _renderToString from "vue-server-renderer/basic";
import { createInfiniteHitsSessionStorageCache } from "instantsearch.js/es/lib/infiniteHitsCache";

function renderToString(app) {
  return new Promise((resolve, reject) => {
    _renderToString(app, (err, res) => {
      if (err) reject(err);
      resolve(res);
    });
  });
}

const indexName = "product";

const searchClient = algoliasearch(
  "XXXXX",
  "XXXXX"
);

// read and write router state to support algolia query parameter
[...]

Issue #1062, opened 2021-09-20
During SSR the url doesn't get read


ArcoMul:
**Bug 🐞**

### What is the current behavior?

During SSR the page gets rendered with an empty state because the state doesn't get read from the url.

### Make a sandbox with the current behavior

See SSR example, neither `routing.router.read` or `routing.router.onUpdate` gets called.

### What is the expected behavior?

That vue-instantsearch tries to retrieve the current state from the URL, using the routing.router methods

### Does this happen only in specific situations?

No

### What is the proposed solution?

That either `routing.router.onUpdate` gets called so that the callback can be used to push the state into vue-instantsearch, or that `routing.router.read` gets called so that either the state is retrieved automatically.

### What is the version you are using?

3.3.0 and 4.0.1

ArcoMul:
Might be that I'm not aware of another way to push the state into vue-instantsearch, the example suggest though that `onUpdate` should be called during server-side rendering.

See also the post I created on Discourse: https://discourse.algolia.com/t/how-to-pass-current-url-to-vue-instantsearch-routing-during-ssr/13476

reinoldus:
I think it's not the url, my ssr is not working either anymore, this sandbox works:
https://codesandbox.io/s/serene-leftpad-jp7gb

this one doesn't:
https://codesandbox.io/s/crazy-sid-tkwu7?file=/package.json

Only change is that nuxt was upgrade to 2.15.8

not sure what deps it broke

reinoldus:
I reused my old yarn.lock and it works again, either some vue package or some instansearch.js package caused this issue. I leave it to the maintainers to figure out which one sh*t the bed

Haroenv:
Those two sandboxes look exactly the same to me, even the nuxt version resolves to 2.15.0 in both, the . Unfortunately I have no idea what's the difference that could be causing the url not to be read. I'm guessing maybe codesandbox doesn't respect the lock file somehow? can you reproduce this working/not working locally too?

reinoldus:
That is highly interesting, so what I did was "yarn upgrade" in the other sandbox, not sure why the yarn.lock didn't update, I ran yarn upgrade again and now the lock file has been updated.

And yes I had the exact same issue locally:
I did "yarn upgrade" locally and the "url not read" bug occurred.

[...]

PR #1106, opened 2022-01-31
feat(ssr): prevent initial network request

Haroenv:
Same as #1090, but for Vue InstantSearch v3

fixes algolia/vue-instantsearch#1066
fixes algolia/vue-instantsearch#1062
closes https://github.com/algolia/instantsearch.js/pull/5002

### DIFF
--- a/src/util/__tests__/createServerRootMixin.test.js
+++ b/src/util/__tests__/createServerRootMixin.test.js
@@ -10,11 +10,7 @@ import SearchBox from '../../components/SearchBox.vue';
 import { createWidgetMixin } from '../../mixins/widget';
 import { createFakeClient } from '../testutils/client';
 import { createSerializedState } from '../testutils/helper';
-import {
-  SearchResults,
-  SearchParameters,
-  AlgoliaSearchHelper,
-} from 'algoliasearch-helper';
+import { SearchParameters, SearchResults } from 'algoliasearch-helper';
 
 jest.unmock('instantsearch.js/es');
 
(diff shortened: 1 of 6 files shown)


src/util/createServerRootMixin.js
```diff
@@ -168,30 +142,26 @@ function augmentInstantSearch(
...
     widget.render({
       helper: localHelper,
       results,
-      scopedResults: parent.getScopedResults().map(result =>
-        Object.assign(result, {
-          results: search.__initialSearchResults[result.indexId],
-        })
-      ),
+      scopedResults: parent.getScopedResults(),
+      parent,
```
sarahdayan, line 167: Any reason why we used not to pass `parent`?

src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -519,63 +514,76 @@ Array [
...
+    },
+  },
+]
+`);
     });
+  });
 
-    it('accepts non-stringified results', () => {
+  describe('hydrate', () => {
+    it('sets _initialResults', () => {
```
sarahdayan, line 578: I see there was a conversation to discard this test [here](https://github.com/algolia/vue-instantsearch/pull/1090/files?show-viewed-files=true&file-filters%5B%5D=#r764181800), but it wasn't removed in the end. Any reason not to? Anything that it tests that isn't covered elsewhere?

sarahdayan (approved):

src/util/__tests__/createServerRootMixin.test.js
Haroenv: It's indeed testing some InstantSearch internals, but the main reason it stayed is that Vue InstantSearch is extending InstantSearch with a couple methods, which could in theory cause hydrate not to work correctly anymore

dhayab (approved):

PR #1110, opened 2022-02-04
ci(vue3): pin lower version of @babel/core

Haroenv:
As we manually install the versions, the latest version of transitive dependencies, which happens to be @babel/core@7.17.0, which has the bug https://github.com/babel/babel/issues/14233.

Really we should solve this problem completely different, maybe with a separate folder, a separate package.json/yarn.lock possibly, but I couldn't find a straightforward way to do that right now.

Test plan: ci for vue 3 is no longer failing


dhayab (approved):
The job passes successfully 👍.

Issue #1098, opened 2022-01-12
Components does not work in a fresh Nuxt 3 project


Baroshem:
**Bug 🐞**

### What is the current behavior?

![image](https://user-images.githubusercontent.com/37120330/149143427-c7835275-1cbc-4387-b38c-19708b08eca4.png)

### Make a sandbox with the current behavior

1. Create new nuxt 3 project.
2. Follow the tutorial here https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/vue/#with-nuxt (but use `vue-instantsearch/vue3/es instead`)
3. See the error

Template: https://codesandbox.io/s/github/algolia/create-instantsearch-app/tree/templates/vue-instantsearch

### What is the expected behavior?

It should render the components correctly

### Does this happen only in specific situations?

Always

### What is the proposed solution?

Review this hogan library.

### What is the version you are using?

"algoliasearch": "^4.11.0",
"vue-instantsearch": "^4.3.1"

> Always try the latest one before opening an issue.

Haroenv:
Thanks for leaving this comment, I haven't yet had the chance to try out nuxt 3 since it's out of private beta.

Haroenv:
The main issue seems to be that ESM is set up to be more strict in this version of vite, so we'll need to do the same changes as https://github.com/algolia/react-instantsearch/pull/3232 and https://github.com/algolia/react-instantsearch/pull/3257

If I did those changes manually I have a working project (no server side rendering yet): https://github.com/Haroenv/nuxt3-vue-instantsearch-demo

Baroshem:
Thanks @Haroenv for the update. Are there any plans to update vue instantsearch accordingly to your changes in the demo repo similar to react-instantsearch? Or should I use the package with patch for the nuxt module for algolia?

Haroenv:
We hadn't anticipated any changes needed, so this still needs to become prioritised etc. I'd expect a couple weeks before we can finish this out of the box, the patches would work for the time being, or a pull request is also welcome of course!

Baroshem:
I see. Should I create a PR with the changes from your repository? I am not sure if that won't break the current package for Vue 3 projects though

Baroshem:
[...]

PR #1111, opened 2022-02-09
fix(esm): mark esm exports as type: module

Haroenv:
fixes https://github.com/algolia/vue-instantsearch/issues/1098

- makes all imports from InstantSearch.js resolved
- refactor build script and tests
- make /cjs import from InstantSearch cjs

Notes for review: I refactored the build to not require to separate calls to `rollup`, this makes the diff a little harder to follow, as new plugins are also added. You can see the individual commits for each of the steps that happened


Haroenv:
That would be good, but seems like it breaks fsevents on macOS for me, so will do that in a separate PR @dhayab

PR #1112, opened 2022-02-10
fix(nuxt3): forward $nuxt on the instance by default

Haroenv:
This is similar to how we forward $router and $store, but in this case it needs to happen before the beforeCreate of nuxt happens, as it uses that property

### DIFF
--- a/src/util/createServerRootMixin.js
+++ b/src/util/createServerRootMixin.js
@@ -105,9 +105,11 @@ function augmentInstantSearch(instantSearchOptions, cloneComponent) {
           mixins: [
             {
               beforeCreate() {
-                // In case of Nuxt (3), we ensure the context is shared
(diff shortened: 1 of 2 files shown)


src/util/createServerRootMixin.js
```diff
@@ -104,6 +104,11 @@ function augmentInstantSearch(instantSearchOptions, cloneComponent) {
         app = cloneComponent(component, {
           mixins: [
             {
+              beforeCreate() {
+                // In case of Nuxt (3), we ensure the context is shared
+                // Note that this has no effect outside of Nuxt
+                this.$nuxt = component.$nuxt;
```
dhayab, line 110: Should we wrap this in a condition to check whether `component.$nuxt` actually exists?

Haroenv: yes, good point

PR #1115, opened 2022-02-14
fix(ssr): use dedicated instance per render

### LINKED ISSUE #1114: Nuxt dev server hangs when using `findResultsState` with children routes
**Bug 🐞**

I have an odd issue when using Nuxt, SSR, and vue-instantsearch. Following [this guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/vue/#with-nuxt), I've made [a stripped down demo repo](https://github.com/sheldonbaker/nuxt-instantsearch-hang) of the bug in question (I've confirmed that the issue is happening with a fresh `git pull`, `yarn` and `yarn dev` of that repo).

It simply seems that some combination of using `this.instantsearch.findResultsState` in `serverPrefetch` *and* having a `<router-view />` or `<NuxtChild />` in the same component causes the dev server to eat a bunch of memory and eventually crash. Sometimes an event emitter error is printed to the dev server console. Aside from that, not much to go on.

### LINKED ISSUE #5387: Nuxt dev server hangs when using `findResultsState` with children routes
**Bug 🐞**

I have an odd issue when using Nuxt, SSR, and vue-instantsearch. Following [this guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/vue/#with-nuxt), I've made [a stripped down demo repo](https://github.com/sheldonbaker/nuxt-instantsearch-hang) of the bug in question (I've confirmed that the issue is happening with a fresh `git pull`, `yarn` and `yarn dev` of that repo).

It simply seems that some combination of using `this.instantsearch.findResultsState` in `serverPrefetch` *and* having a `<router-view />` or `<NuxtChild />` in the same component causes the dev server to eat a bunch of memory and eventually crash. Sometimes an event emitter error is printed to the dev server console. Aside from that, not much to go on.

Haroenv:
fixes algolia/instantsearch#5387

We use the same InstantSearch instance for every call of `createServerRootMixin`, but that means we start multiple times, without destroying, causing memory leaks.

### DIFF
--- a/src/components/__tests__/InstantSearch.js
+++ b/src/components/__tests__/InstantSearch.js
@@ -274,7 +274,7 @@ it('does not warn when `routing` have either `router` or `stateMapping`', () => 
   expect(warn).toHaveBeenCalledTimes(0);
 });
 
-it('Does not allow a change in `routing`', () => {
+it('Does not allow a change in `routing`', async () => {
   const wrapper = mount(InstantSearch, {
     propsData: {
       searchClient: {},
@@ -282,9 +282,9 @@ it('Does not allow a change in `routing`', () => {
     },
   });
 
-  expect(
+  await expect(
     wrapper.setProps({
-      routing: false,
+      routing: { stateMapping: {} },
     })
   ).rejects.toMatchInlineSnapshot(`
 [Error: routing configuration can not be changed dynamically at this point.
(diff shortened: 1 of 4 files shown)


Haroenv:
I'll close this for now, as it causes some relevant issues in the tests. They might be fixable, but as the problem doesn't come up often, I'll leave it as-is for now, to avoid disturbing otherwise working implementations

PR #1117, opened 2022-02-28
fix(nuxt): only write $nuxt if it's writable

Haroenv:
In some nuxt 2 cases (https://github.com/algolia/vue-instantsearch/commit/acda29326475bd1b73f12b058a0c02df00b8b239#commitcomment-67430759) this.$nuxt is already defined by the framework, but not a setter, so we can't make it equal to the $nuxt of the component.

Luckily in Nuxt 2, setting $nuxt isn't crucial, and can thus be skipped.

### DIFF
--- a/src/util/createServerRootMixin.js
+++ b/src/util/createServerRootMixin.js
@@ -112,7 +112,7 @@ function augmentInstantSearch(instantSearchOptions, cloneComponent) {
 
                 const isWritable = descriptor
                   ? descriptor.writable || descriptor.set
-                  : true;
+                  : false;
 
                 if (component.$nuxt && isWritable) {
                   // In case of Nuxt (3), we ensure the context is shared between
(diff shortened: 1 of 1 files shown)


src/util/createServerRootMixin.js
```diff
@@ -105,7 +105,16 @@ function augmentInstantSearch(instantSearchOptions, cloneComponent) {
...
               beforeCreate() {
-                if (component.$nuxt) {
+                const descriptor = Object.getOwnPropertyDescriptor(
+                  this,
+                  '$nuxt'
+                );
+
+                const isWritable = descriptor
+                  ? descriptor.writable || descriptor.set
+                  : true;
```
Haroenv, line 115:
```suggestion
                  : false;
```

src/util/createServerRootMixin.js
```diff
@@ -105,7 +105,16 @@ function augmentInstantSearch(instantSearchOptions, cloneComponent) {
           mixins: [
             {
               beforeCreate() {
-                if (component.$nuxt) {
+                const descriptor = Object.getOwnPropertyDescriptor(
+                  this,
```
dhayab, line 109: Shouldn't this be `component` ?

```suggestion
                  component,
```

dhayab (approved):

PR #1133, opened 2022-07-18
fix(searchbox): prevent concurrent query updates from state while input is focused

### LINKED ISSUE #1123: Vue - Lag Typing (Meili Search)
https://user-images.githubusercontent.com/3077812/167888542-4d17f477-961d-4870-a37d-a11c6c4eae04.mov

When i try to type fast, it would automatically snapback and cancel my typing and replace the text with what the server respond with. I'm not sure where to look into as my configuration are almost identical to the demo located here
https://codesandbox.io/s/ms-vue-is-1d6bi?fontsize=14&hidenavigation=1&theme=dark&file=/src/App.vue

I don't have other JS files, so i don't think it's conflicting with others. I've tried disabling my alphine.js which i use for tailwind ui stack but still face the same issue.

sorry to post this here but the issue seems to be appearing depending on which vue-instantsearch version i set , at this moment certain version doesn't give issue on codesandbox but i can't seem to get it to work properly when i compile it on my laravel even with cache cleared and all. Any idea where i should look into to solve this issue ? 

app.js

```
window.Vue = require('vue').default;
Vue.component('example-component', require('./components/ExampleComponent.vue').default);
import Vue from "vue";
import InstantSearch from "vue-instantsearch";

Vue.use(InstantSearch);
Vue.config.productionTip = false;

const app = new Vue({
    el: '#app',
});
```

package.json

```
{
    "private": true,
    "scripts": {
        "dev": "npm run development",
        "development": "mix",
        "watch": "mix watch",
        "watch-poll": "mix watch -- --watch-options-poll=1000",
        "hot": "mix watch --hot",
        "prod": "npm run production",
        "production": "mix --production"
    },
    "devDependencies": {
        "@popperjs/core": "^2.10.2",
        "autoprefixer": "^10.2.5",
        "axios": "^0.21",
        "bootstrap": "^5.1.3",
        "laravel-mix": "^6.0.18",
        "lodash": "^4.17.19",
        "postcss": "^8.2.12",
        "postcss-purgecss-laravel": "^2.0.0",
        "resolve-url-loader": "^3.1.2",
        "sass": "^1.32.11",
        "sass-loader": "^11.0.1",
        "tailwindcss": "^2.1.2",
        "vue": "^2.6.12",
        "vue-loader": "^15.9.8",
        "vue-template-compiler": "^2.6.12"
    },
    "dependencies": {
        "@meilisearch/instant-meilisearch": "0.6.0",
        "@tailwindcss/aspect-ratio": "^0.2.0",
        "@tailwindcss/forms": "^0.3.2",
        "alpinejs": "^2.8.2",
        "vue-instantsearch": "4.0.0",
        "instantsearch.css": "7.4.2"
    }
}
```

You can test it here too 
https://codesandbox.io/s/ms-vue3-is-0293zk?file=/src/App.vue (vue 3)

dhayab:
JIRA: [FX-1552](https://algolia.atlassian.net/browse/FX-1552)

Fixes #1123 

This PR prevents state from overriding the query value in a search box if it still has focus. This can happen in a search-as-you-type environment when network speed is low and a user keeps on typing a query while requests are resolved.

### DIFF
--- a/src/components/SearchBox.vue
+++ b/src/components/SearchBox.vue
@@ -20,6 +20,7 @@
         :reset-title="resetTitle"
         :class-names="classNames"
         v-model="currentRefinement"
+        ref="searchInput"
       >
         <template
           v-slot:loading-indicator
@@ -148,6 +149,14 @@ export default {
           this.$emit('update:modelValue', this.model);
           this.state.refine(this.model);
         }
+
+        // we return the local value if the input is focused to avoid
+        // concurrent updates when typing
+        const { searchInput } = this.$refs;
+        if (searchInput && searchInput.isFocused()) {
+          return this.localValue;
+        }
+
         return this.model || this.state.query || '';
       },
       set(val) {
--- a/src/components/SearchInput.vue
+++ b/src/components/SearchInput.vue
@@ -170,6 +170,9 @@ export default {
     };
   },
   methods: {
+    isFocused() {
+      return document.activeElement === this.$refs.input;
+    },
     onFormSubmit() {
       const input = this.$refs.input;
       input.blur();
--- a/src/components/__tests__/SearchBox.js
+++ b/src/components/__tests__/SearchBox.js
@@ -147,6 +147,21 @@ test('refine on empty string on form reset', async () => {
   expect(state.refine).toHaveBeenCalledWith('');
 });
 
+test('keep local query when out of sync and input is focused', async () => {
+  const state = { ...defaultState, refine: jest.fn() };
+  __setState(state);
+
+  const wrapper = mount(SearchBox);
+  const input = wrapper.find('.ais-SearchBox-input');
+  await input.trigger('focus');
+  await input.setValue('hello');
+
+  await wrapper.setData({ state: { query: 'hel' } });
+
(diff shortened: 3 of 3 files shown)


src/components/SearchBox.vue
```diff
@@ -148,6 +149,14 @@ export default {
           this.$emit('update:modelValue', this.model);
           this.state.refine(this.model);
         }
+
+        // we return the local value if the input is focused to avoid
+        // concurrent updates when typing
+        const { searchInput } = this.$refs;
+        if (searchInput && searchInput.isFocused()) {
+          return this.localValue;
```
Haroenv, line 157: the localValue already should be equal to the model, no?

Haroenv (approved):
as far as I can tell the failure mode with slow network doesn't happen in this PR

src/components/SearchBox.vue
dhayab: From what I understand, only in controlled mode. `localValue` is directly tied to the setter of `currentRefinement`, so it made sense to return its value early in the getter when the input is focused.

PR #1136, opened 2022-07-25
fix(components): forward missing props in SearchBox and VoiceSearch

### LINKED ISSUE #1135: Voice search language option
**Feature ⚡️**

### What is your use case for such a feature?
To be able to set the language for voice search. There is no option on ais-voice-search to set the language as there would be when using the js voice search library.

### What is your proposal
Implement widget-param-language as an option on ais-voice-search
https://www.algolia.com/doc/api-reference/widgets/voice-search/js/?client=vue#widget-param-language

"vue-instantsearch": "^4.3.3",

Haroenv:
fixes #1135

I've wen through the list of all widgets, and only these on VoiceSearch and queryHook on SearchBox were missing.

I didn't add any tests, as none of the widgets test widgetParams (since it's not tested in any other widget, and likely would just be a copy paste of the code)

### DIFF
--- a/src/components/SearchBox.vue
+++ b/src/components/SearchBox.vue
@@ -120,6 +120,10 @@ export default {
       type: String,
       default: undefined,
     },
+    queryHook: {
+      type: Function,
+      default: undefined,
+    },
   },
   data() {
     return {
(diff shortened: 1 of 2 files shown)


dhayab (approved):
👏  great that you looked at other missing props!

PR #1141, opened 2022-08-30
feat(can-refine): provide `canRefine` for multiple widget slots

aymeric-giraudet:
Following [FX-740](https://algolia.atlassian.net/browse/FX-740), we want to provide `canRefine` to slots and use it for `createPanelConsumerMixin()`

### DIFF
--- a/src/components/HitsPerPage.vue
+++ b/src/components/HitsPerPage.vue
@@ -43,9 +43,7 @@ export default {
         $$widgetType: 'ais.hitsPerPage',
       }
     ),
-    createPanelConsumerMixin({
-      mapStateToCanRefine: state => Boolean(state.canRefine),
-    }),
+    createPanelConsumerMixin(),
   ],
   props: {
     items: {
--- a/src/components/HitsPerPage.vue
+++ b/src/components/HitsPerPage.vue
@@ -7,6 +7,7 @@
       :items="state.items"
       :refine="state.refine"
       :hasNoResults="state.hasNoResults"
+      :canRefine="state.canRefine"
     >
       <select
         :class="suit('select')"
@@ -43,7 +44,7 @@ export default {
       }
     ),
     createPanelConsumerMixin({
-      mapStateToCanRefine: state => state.hasNoResults === false,
+      mapStateToCanRefine: state => Boolean(state.canRefine),
     }),
   ],
   props: {
--- a/src/mixins/panel.js
+++ b/src/mixins/panel.js
@@ -39,7 +39,7 @@ export const createPanelProviderMixin = () => ({
   },
 });
 
-export const createPanelConsumerMixin = ({ mapStateToCanRefine }) => ({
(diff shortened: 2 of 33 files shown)


src/components/HitsPerPage.vue
```diff
@@ -43,7 +44,7 @@ export default {
       }
     ),
     createPanelConsumerMixin({
-      mapStateToCanRefine: state => state.hasNoResults === false,
+      mapStateToCanRefine: state => Boolean(state.canRefine),
```
dhayab, line 47: Isn't `state.canRefine` already a boolean?

aymeric-giraudet: Not always, it might be `undefined`, see here : https://github.com/algolia/vue-instantsearch/pull/876
Should I maybe mention it in a comment or would it be redundant ?

Haroenv, line 47: now that everything has canRefine, could we put the code inside the panel mixin itself?

dhayab: I see, thanks! I don't think there is a need to add a comment in that case.

aymeric-giraudet: Makes sense, thanks ! It's done, there's a lot more code to review now though

src/mixins/panel.js
```diff
@@ -39,7 +39,7 @@ export const createPanelProviderMixin = () => ({
   },
 });
 
-export const createPanelConsumerMixin = ({ mapStateToCanRefine }) => ({
+export const createPanelConsumerMixin = () => ({
```
Haroenv, line 42:
```suggestion
export const createPanelConsumerMixin = ({ mapStateToCanRefine = state => Boolean(state.canRefine) }) => ({
```

src/mixins/panel.js
```diff
@@ -64,8 +64,10 @@ export const createPanelConsumerMixin = ({ mapStateToCanRefine }) => ({
           return;
         }
 
-        const previousCanRefine = mapStateToCanRefine(previousState || {});
-        const nextCanRefine = mapStateToCanRefine(nextState);
+        const previousCanRefine = previousState
+          ? previousState.canRefine
+          : false;
+        const nextCanRefine = nextState.canRefine;
```
Haroenv, line 70:
```suggestion
        const previousCanRefine = mapStateToCanRefine(previousState || {});
        const nextCanRefine = mapStateToCanRefine(nextState);
```

src/mixins/panel.js
Haroenv: this way external usage doesn't experience a breaking change :)

Haroenv (commented):
👍

src/mixins/panel.js
aymeric-giraudet: Oh indeed, didn't see it was exported to users !

dhayab (approved):

Haroenv (approved):

PR #1142, opened 2022-09-01
feat(dynamicWidgets): add fallback slot to render attributes without matching component

aymeric-giraudet:
[FX-1683](https://algolia.atlassian.net/browse/FX-1683)

Due to library users need, this PR introduces a `fallback` slot to `ais-dynamic-widgets`, which works similarly as `fallbackComponent` for `react-instantsearch`.

### DIFF
--- a/src/components/DynamicWidgets.js
+++ b/src/components/DynamicWidgets.js
@@ -96,10 +96,18 @@ export default {
       );
     }
 
+    const fallbackSlot = isVue3
+      ? this.$slots.fallback
+      : this.$scopedSlots.fallback;
+
     return h(
       'div',
       { class: [this.suit()] },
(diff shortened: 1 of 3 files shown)


Haroenv:
Did you test this with server-side rendering? in React InstantSearch Hooks this required extra code to ensure the rendering happens twice so the required widgets can be rendered correctly. I'm not sure if this code exists in Vue InstantSearch too

aymeric-giraudet:
@Haroenv I guess I should close this PR.
I think it could be nice to add the sample I did to the DynamicWidgets doc vue flavor though ! What do you think ?

Haroenv (changes requested):
As we couldn't find a way to make this work in server side rendering (maybe after #1115), this shouldn't yet be merged

Issue #1145, opened 2022-09-12
Issue : “lost” characters when typing fast


ThomasKientz:
**Bug 🐞**

### What is the current behavior?

When typing fast, some characters are missing in the searchbox with a custom renderer 

```html
<ais-search-box >
  <template #default="{ currentRefinement, isSearchStalled, refine }">
    <input
      placeholder="search"
      type="search"
      :value="currentRefinement"
      @input="refine($event.currentTarget.value)"
    />
    <span :hidden="!isSearchStalled">Loading...</span>
    </template>
</ais-search-box>
``` 
Eg : Typing "house" fast. (reload sandbox when you try to clear cache between you tests).

I have link my production index in the repro. You can see the issue in production here (open sourced) : https://mdisearch.com/

https://user-images.githubusercontent.com/10157845/189675208-12299b32-64ba-42b2-af1b-ba0bebf69ddc.mov

### Sandbox

https://codesandbox.io/s/algolia-reproduction-issue-kt95nc

dhayab:
Hi, is there a reason you use a custom input element in `<ais-search-box>`?

The `currentRefinement` value exposed in the slot scope is set by the results coming from Algolia, which means that in suboptimal network settings or when typing too fast, it might override the most recent value with the one that was taken into account in the latest query.

In our default template we handle this by giving priority to the local input value as long as its focused, so if you change the template, you would need to re-implement this behavior.

ThomasKientz:
@dhayab Hi, reason is to customize in realword apps like https://github.com/ThomasKientz/mdisearch/blob/main/front/src/App.vue#L77

> In our default template we handle this by giving priority to the local input value as long as its focused, so if you change the template, you would need to re-implement this behavior.

Could you pinpoint me to a specific line please ? I will try to implement it as well.

Thanks !

ThomasKientz:
Got it : https://github.com/algolia/vue-instantsearch/blob/master/src/components/SearchBox.vue#L162

ThomasKientz:
@dhayab pushed for addressing this.

Haroenv:
I don't think this can fully be solved without a custom widget, a slot will always have too little information to make a truly correct search box. It could be interesting to change the documentation to migrate from slot to custom widget for a custom search box, but for now this issue should be enough reference for anyone else falling into this issue.

Thanks for reporting Thomas!

PR #1146, opened 2022-09-12
fix(SearchBox): currentRefinement slot prop is not in sync with custom input

ThomasKientz:
fix https://github.com/algolia/vue-instantsearch/issues/1145

### DIFF
--- a/src/components/SearchBox.vue
+++ b/src/components/SearchBox.vue
@@ -6,11 +6,12 @@
     <slot
       :current-refinement="currentRefinement"
       :is-search-stalled="state.isSearchStalled"
+      :set-is-focused="setIsFocused"
       :refine="state.refine"
     >
       <search-input
-        @focus="$emit('focus', $event)"
-        @blur="$emit('blur', $event)"
+        @focus="$emit('focus', $event); setIsFocused(true)"
+        @blur="$emit('blur', $event); setIsFocused(false)"
         @reset="$emit('reset')"
         :placeholder="placeholder"
         :autofocus="autofocus"
@@ -130,6 +131,7 @@ export default {
       localValue: '',
       isVue2,
       isVue3,
+      isFocused: false,
     };
   },
   computed: {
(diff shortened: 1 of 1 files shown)


src/components/SearchBox.vue
```diff
@@ -161,7 +161,12 @@ export default {
 
         // we return the local value if the input is focused to avoid
         // concurrent updates when typing
-        const { searchInput } = this.$refs;
+        const { searchInput } =
+          (this.$scopedSlots.default &&
+            this.$scopedSlots.default()[0] &&
+            this.$scopedSlots.default()[0].context.$refs) ||
```
Haroenv, line 167: this is essentially assuming there's a ref inside the scoped slot with that same name, but it probably would be better to pass `searchInputRef` as a data to the scoped slot, so that it can be bound as expected (bonus that it would also work in vue 3 without conditionals)

ThomasKientz: Pushed a new commit.

ThomasKientz:
@Haroenv What do you think ?

ThomasKientz:
@Haroenv It does work with second approach using `setIsFocus` slot prop.

```html
<ais-search-box>
  <template
    #default="{
      currentRefinement,
      isSearchStalled,
      refine,
      setIsFocus,
    }"
  >
    <SearchInput
      ref="searchInput"
      :value="currentRefinement"
      :loading="isSearchStalled"
      @focus="setIsFocus(true)"
      @blur="setIsFocus(false)"
      @input="refine($event)"
    />
  </template>
</ais-search-box>
```

I will have a look later on approach with react.

Haroenv:
@ThomasKientz, with this code the query actually never gets correctly updated: https://codesandbox.io/s/vue-instantsearch-e-commerce-forked-jpcjcj?file=/src/App.vue (as the local value gets used, but it never gets updated while it's focused)


https://user-images.githubusercontent.com/6270048/192267502-86491ba8-61a9-40bb-8757-976a0606eb2e.mov



While it's not ideal the default slot doesn't catch this race condition, I think it's better if you use `createWidgetMixin` for a custom input (I'll update the docs with that soon I think), as coordinating the focus state is too complex in Vue.

ThomasKientz:
My bad, my PR doesn't work at all... Will look into `createWidgetMixin`.

Haroenv:
No problem, I'll close the PR for now, thanks a lot and if you find another way to keep the inputs synced out of the box, that would be really cool (but unfortunately I've also looked for a while and can't find a sufficient solution)

PR #1151, opened 2022-10-03
feat(StateResults): give access to status and error

Haroenv:
FX-1770

Note that this changes the implementation from connectStateResults to a render listener, as `loading` is otherwise inconsistently visible. This means it will no longer be counted in metadata.

Updated the helper in devDep to use ^ to avoid a duplicate with the InstantSearch.js version

### DIFF
--- a/src/components/StateResults.vue
+++ b/src/components/StateResults.vue
@@ -12,34 +12,77 @@
       </p>
       <pre>results: {{ Object.keys(state.results) }}</pre>
       <pre>state: {{ Object.keys(state.state) }}</pre>
+      <pre>status: {{ state.status }}</pre>
+      <pre>error: {{ state.error }}</pre>
     </slot>
   </div>
 </template>
 
 <script>
+import { isVue3 } from '../util/vue-compat';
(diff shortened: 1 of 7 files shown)


src/components/StateResults.vue
```diff
@@ -12,34 +12,77 @@
...
   mixins: [
-    createWidgetMixin(
-      {
-        connector: connectStateResults,
-      },
-      {
-        $$widgetType: 'ais.stateResults',
-      }
-    ),
+    createWidgetMixin({ connector: true }),
```
dhayab, line 30: Maybe we could take the opportunity to add `$$widgetType` here?

dhayab (approved):
This works as expected and doesn't change the developer experience 👍 .

src/components/StateResults.vue
Haroenv: it doesn't do anything when there's no actual connector/widget, leaving it would be confusing IMO

dhayab: Oh I didn't know, thanks for clarifying!

aymeric-giraudet (approved):

PR #1158, opened 2022-10-21
feat(PoweredBy): update component logo

FabienMotte:
**Summary**

This PR updates the SVG of the `<PoweredBy>` widget component and test snapshot.


Haroenv (approved):
width is different, but this should be good

